PR #6, opened 2016-10-04
Add parentJobID feature to bedrock

iwiznia:
@cead22 please review.

This adds the parent job functionality.

Tested with the integration tests in https://github.com/Expensify/Web-Expensify/pull/15050

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -21,7 +21,7 @@ class BedrockPlugin_Jobs : public BedrockNode::Plugin {
     // Helper functions
     string _constructNextRunDATETIME(const string& lastScheduled, const string& lastRun, const string& repeat);
     bool _validateRepeat(const string& repeat) { return !_constructNextRunDATETIME("", "", repeat).empty(); }
-    int _getCountPendingChildJobs(SQLite& db, const int jobID);
+    int _getCountOfPendingChildJobs(SQLite& db, const int jobID);
 };
 
 // Register for auto-discovery at boot
@@ -259,7 +259,21 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
         // If no priority set, set it
         int priority = request.isSet("priority") ? request.calc("priority") : JOBS_DEFAULT_PRIORITY;
 
-        const string& safeParentJobID = request["parentJobID"].empty() ? SQ("NULL") : SQ(request["parentJobID"]);
+        string safeParentJobID = SQ("NULL");
+        if (!request["parentJobID"].empty()) {
+            safeParentJobID = SQ(request["parentJobID"]);
+            SQResult result;
+            if (!db.read("SELECT 1 "
+                         "FROM jobs "
+                         "WHERE jobID = " + safeParentJobID + ";",
+                         result)) {
+                throw "502 Select failed";
+            }
+
+            if (result.empty()) {
+                throw "403 parentJobID does not exist";
+            }
+        }
 
         // We'd initially intended for any value to be allowable here, but for performance reasons, we currently
         // will only allow specific values to try and keep queries fast. If you pass an invalid value, we'll throw
@@ -436,9 +450,9 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
         }
 
         // If we are finishing a job that has child jobs, set its state to paused.
-        const int childCount = _getCountPendingChildJobs(db, request.calc64("jobID"));
+        const int childCount = _getCountOfPendingChildJobs(db, request.calc64("jobID"));
         if (SIEquals(request.methodLine, "FinishJob") && childCount != 0) {
-            SINFO("Job has " + SToStr(childCount) + "child jobs, PAUSING");
+            SINFO("Job has " + SToStr(childCount) + " child jobs, PAUSING");
             if (!db.write("UPDATE jobs SET state=" + SQ("PAUSED") + " WHERE jobID=" + SQ(request.calc64("jobID")) + ";")) {
                 throw "502 Update failed";
             }
@@ -489,7 +503,7 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
             if (!db.write("DELETE FROM jobs WHERE jobID=" + SQ(request.calc64("jobID")) + ";")) {
                 throw "502 Delete failed";
             }
-            if (parentJobID > 0 && _getCountPendingChildJobs(db, parentJobID) == 0) {
+            if (parentJobID > 0 && _getCountOfPendingChildJobs(db, parentJobID) == 0) {
                 SINFO("Job has parentJobID: " + SToStr(parentJobID) + " and no other pending children, resuming parent job");
                 if (!db.write("UPDATE jobs SET state = 'QUEUED' where jobID=" + SQ(parentJobID) + ";")) {
                     throw "502 Update failed";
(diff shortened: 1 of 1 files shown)


iwiznia:
Off hold.

plugins/Jobs.cpp
```diff
@@ -40,16 +41,21 @@ void BedrockPlugin_Jobs::upgradeDatabase(BedrockNode* node, SQLite& db) {
                                    "repeat   TEXT NOT NULL, "
                                    "data     TEXT NOT NULL, "
                                    "priority INTEGER NOT NULL DEFAULT " +
-                                       SToStr(JOBS_DEFAULT_PRIORITY) + " ) ",
+                                       SToStr(JOBS_DEFAULT_PRIORITY) + ", "+
+                                   "parentJobID INTEGER )",
                            ignore));
 
+    // FIXME: Remove this after upgrade
+    SASSERT( db.write( "ALTER TABLE jobs ADD COLUMN parentJobID INTEGER;" ) );
```
iwiznia: I see this is how we added the priority column....

iwiznia:
Off hold

iwiznia:
Bump!

plugins/Jobs.cpp
```diff
@@ -638,3 +665,23 @@ string BedrockPlugin_Jobs::_constructNextRunDATETIME(const string& lastScheduled
...
     return "DATETIME( " + SComposeList(safeParts) + " )";
 }
+
+/**
+ * Returns the number of RUNNING or QUEUED jobs that have the passed jobID as their parent.
+ * @param db
+ * @param jobID
+ * @return
+ */
+int BedrockPlugin_Jobs::_getCountPendingChildJobs(SQLite& db, const int jobID)
```
cead22: `_getCountOfPendingChildJobs` > `_getCountPendingChildJobs`, nab

plugins/Jobs.cpp
```diff
@@ -638,3 +665,23 @@ string BedrockPlugin_Jobs::_constructNextRunDATETIME(const string& lastScheduled
     // Combine the parts together and return the full DATETIME statement
     return "DATETIME( " + SComposeList(safeParts) + " )";
 }
+
+/**
+ * Returns the number of RUNNING or QUEUED jobs that have the passed jobID as their parent.
+ * @param db
```
cead22: Why add these when not adding descriptions, convention? I find this useless, nab

cead22: Especially the `@return` one

plugins/Jobs.cpp
```diff
@@ -418,13 +427,25 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
...
+        const int parentJobID = SToInt(result[0][4]);
 
         // Make sure we're finishing a job that's actually running
         if (state != "RUNNING") {
             SWARN("Trying to finish job#" << request["jobID"] << ", but isn't RUNNING (" << state << ")");
             throw "405 Can only retry/finish RUNNING jobs";
         }
 
+        // If we are finishing a job that has child jobs, set its state to paused.
+        const int childCount = _getCountPendingChildJobs(db, request.calc64("jobID"));
```
cead22: You can save a query by moving this call into the `if(SIEquals(request.methodLine, "FinishJob"))`, nab  #prematureoptimization

plugins/Jobs.cpp
```diff
@@ -418,13 +427,25 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
...
+        if (SIEquals(request.methodLine, "FinishJob") && childCount != 0) {
+            SINFO("Job has " + SToStr(childCount) + "child jobs, PAUSING");
```
cead22: Need a space in front of the word `child`

cead22:
👍 from me pending thumbs up on database-upgrade-related changes

plugins/Jobs.cpp
```diff
@@ -252,6 +259,8 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
         // If no priority set, set it
         int priority = request.isSet("priority") ? request.calc("priority") : JOBS_DEFAULT_PRIORITY;
 
+        const string& safeParentJobID = request["parentJobID"].empty() ? SQ("NULL") : SQ(request["parentJobID"]);
```
iwiznia: Added.

plugins/Jobs.cpp
iwiznia: Yeah, was auto added by CLion... I don't see the usefulness, removing.

iwiznia:
Updated, retested.

plugins/Jobs.cpp
```diff
@@ -259,7 +259,21 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
...
+            SQResult result;
+            if (!db.read("SELECT 1 "
+                         "FROM jobs "
+                         "WHERE jobID = " + safeParentJobID + ";",
+                         result)) {
+                throw "502 Select failed";
+            }
+
+            if (result.empty()) {
+                throw "403 parentJobID does not exist";
```
flodnv: why 403 and not 404?

plugins/Jobs.cpp
```diff
@@ -418,13 +441,25 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
         const string& nextRun = result[0][1];
         const string& lastRun = result[0][2];
         string repeat = result[0][3];
+        const int parentJobID = SToInt(result[0][4]);
```
quinthar: Also, this should be an `int64_t`

plugins/Jobs.cpp
```diff
@@ -638,3 +679,20 @@ string BedrockPlugin_Jobs::_constructNextRunDATETIME(const string& lastScheduled
...
     return "DATETIME( " + SComposeList(safeParts) + " )";
 }
+
+/**
+ * Returns the number of RUNNING or QUEUED jobs that have the passed jobID as their parent.
+ */
+int BedrockPlugin_Jobs::_getCountOfPendingChildJobs(SQLite& db, const int jobID)
+{
+    SQResult result;
+    if (!db.read("SELECT count(1) "
```
quinthar: There's actually no reason to return a full count -- that is making it do extra work to produce an unnecessarily specific result.   How about instead just `SELECT 1 FROM ... WHERE ... LIMIT 1;` -- this way it'll return the instant it finds one result, rather than counting up all the results just so we can log an interesting number.

So, I suggest changing this to `_checkForPendingChildJobs()` and have it just return true/false.  This also simplifies the code that calls this function, as it eliminates the need to store this unnecessary result into an unnecessary variable.

Finally, don't use `const int jobID`.  The `const` is unnecessary, and the `int` should be an `int64`.

plugins/Jobs.cpp
```diff
@@ -418,13 +441,25 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
...
 
         // Make sure we're finishing a job that's actually running
         if (state != "RUNNING") {
             SWARN("Trying to finish job#" << request["jobID"] << ", but isn't RUNNING (" << state << ")");
             throw "405 Can only retry/finish RUNNING jobs";
         }
 
+        // If we are finishing a job that has child jobs, set its state to paused.
+        const int childCount = _getCountOfPendingChildJobs(db, request.calc64("jobID"));
+        if (SIEquals(request.methodLine, "FinishJob") && childCount != 0) {
```
iwiznia: Only for RetryJob commands, but yeah, we could save one query in that case.

plugins/Jobs.cpp
```diff
@@ -252,6 +259,22 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
...
 
+        string safeParentJobID = SQ("NULL");
+        if (!request["parentJobID"].empty()) {
+            safeParentJobID = SQ(request["parentJobID"]);
+            SQResult result;
+            if (!db.read("SELECT 1 "
+                         "FROM jobs "
+                         "WHERE jobID = " + safeParentJobID + ";",
+                         result)) {
+                throw "502 Select failed";
```
iwiznia: AFAIK this only returns false when the query syntax is wrong or the table or columns don't exist.

quinthar:
Oooooooohh..  Yep, sorry.  You're right.

PR #18, opened 2016-10-17
Add installation instructions to readmes

coleaeason:
@tylerkaraszewski Will you please review?

Adding instructions to install from the repo.

CC: @quinthar 

CC: @shawnborton I don't think the changes to the docs pages will break the styling, but I didn't know of a way to test either. 

Tests:
Ran commands in local dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,21 @@ Bedrock is a simple, modular, WAN-replicated data foundation for global-scale ap
 * **Bedrock is a data foundation.**  This means it is not just a simple database that responds to queries, but rather a platform on which data-processing applications (like databases, job queues, caches, etc) can be built.
 * **Bedrock is for global-scale applications.**  This means it is built to be deployed in a geo-redundant fashion spanning many datacenters around the world.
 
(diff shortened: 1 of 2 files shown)


quinthar:
Thanks!  I'll take care of the README and such; lmk when I can test this binary in the real world.

shawnborton:
To get the doc pages running locally you need to install Jekyll:

`gem install jekyll`

Now from the Bedrock `/docs/` folder, you would run:

`jekyll build` or better yet, `jekyll build --watch` as this will build the docs AND watch for new changes. 

You can access the files from the site folder it will generate (more on that below) or simply serve it up locally by doing:

`jekyll serve` - from here you should be able to fire up `http://localhost:4000/` and see the site.

Keep in mind though, that building locally is going to create the site's contents in `./_site` - *we do NOT want to commit this folder* as Github pages will handle the Jekyll builds automatically. So either add `./_site` to .gitignore (this would be the ideal thing to do) or don't commit that folder after you are done testing.

Your friendly design dad,
Shawn

PR #28, opened 2016-10-19
Add instructions for Xenial.

coleaeason:
@righdforsa Please review.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -29,8 +29,11 @@ Bedrock is not only production ready, but actively used by Expensify's many thou
 ## How to get it
 Bedrock can be compiled from source using the [Expensify/Bedrock](https://github.com/Expensify/Bedrock) public repo, or installed into your Ubuntu environment using the following commands:  (Go ahead and copy/paste this whole section and copy it into your Ubuntu terminal.)
 
-    # Add the Bedrock repo to apt sources:
+    # Add the Bedrock repo to apt sources for your distro:
+    # Trusty Ubuntu 14.04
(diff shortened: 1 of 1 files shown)


quinthar:
Ah, no, don't to this.  The point is to have a block of code you can just copy/paste safely.

coleaeason:
It's still copy and paste, just a different line based on distro. I'm totally open to reformat it however, but they'll have to download a different sources list based on which distribution they are running.

quinthar:
Hm, that sucks.  Is there really no way to make a single .list file that
works for both?  Or somehow make a block of code that can be copied/pasted
in its entirety safely?

coleaeason:
There's not a good option for it. If you created one .list file it would have to contain both list settings and the user would need to comment out the one they don't want. A copy pastable block of code would have to somehow detect what OS they were running and make a choice based off of that. This is the standard way of presenting source lists for multiple distros if you look at other projects. Unless @righdforsa knows of another way or something clever I don't.

quinthar:
I came up with a compromise solution that just asks them to rewrite the URL if they want a different version.  This way a 16.04 user can just copy/paste, and 14.04 can edit.

coleaeason:
Chris and I talked about this and we came up with a copy pastable one liner if you like this more:

`sudo wget -O /etc/apt/sources.list.d/bedrock.list https://apt.bedrockdb.com/ubuntu/dists/$(lsb_release -c | cut -f2)/bedrock.list`

tested and works on both 14.04 and 16.04

quinthar:
Oh, that's even better.  Go for it!

PR #85, opened 2017-01-12
Better c++

Vadius:
* Explicit `nullptr` for pointers instead of 0
* Init lists in constructors instead of assignments in the body
* Fixed memory leak (tls* hasn't been deleted on thread exit)

### DIFF
--- a/BedrockPlugin.cpp
+++ b/BedrockPlugin.cpp
@@ -2,7 +2,7 @@
 #include "BedrockPlugin.h"
 
 // Global static values
-list<BedrockPlugin*>* BedrockPlugin::g_registeredPluginList = 0;
+list<BedrockPlugin*>* BedrockPlugin::g_registeredPluginList = nullptr;
 
 BedrockPlugin::BedrockPlugin() : _enabled(false) {
     // Auto-register this instance into the global static list, initializing the list if that hasn't yet been done.
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -131,7 +131,7 @@ void BedrockServer_WorkerThread(void* _data) {
             node.openCommand(request, priority, false, creationTimestamp);
 
             // Now pull that same command off the internal queue and put it on the appropriate external (threaded) queue
-            BedrockNode::Command* command = 0;
+            BedrockNode::Command* command = nullptr;
             if ((command = node.getProcessedCommand())) {
                 // If it was fully processed in openCommand(), that means it was peeked successfully.
                 SINFO("Peek successful. Putting command '" << command->id << "' on processed list.");
@@ -231,7 +231,7 @@ void BedrockServer_WorkerThread(void* _data) {
                 ;
 
             // Put everything the replication node has finished on the threaded queue.
-            BedrockNode::Command* command = 0;
+            BedrockNode::Command* command = nullptr;
             while ((command = node.getProcessedCommand())) {
                 SAUTOPREFIX(command->request["requestID"]);
                 SINFO("Putting escalated command '" << command->id << "' on processed list.");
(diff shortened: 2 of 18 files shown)


BedrockPlugin.cpp
```diff
@@ -2,7 +2,7 @@
 #include "BedrockPlugin.h"
 
 // Global static values
-list<BedrockPlugin*>* BedrockPlugin::g_registeredPluginList = 0;
+list<BedrockPlugin*>* BedrockPlugin::g_registeredPluginList = nullptr;
```
quinthar: Personally, I don't really like this syntax.  Maybe I'm old school, but I don't find `nullptr` to be more readable or informative than `0`.  Thoughts @tylerkaraszewski?

BedrockServer.cpp
```diff
@@ -268,13 +268,9 @@ void BedrockServer_WorkerThread(void* _data) {
...
 // --------------------------------------------------------------------------
 BedrockServer::BedrockServer(const SData& args)
-    : STCPServer(""), _replicationState(SQLC_SEARCHING), _replicationCommitCount(0), _nodeGracefulShutdown(false),
-      _masterVersion("") {
-    // Initialize
-    _args = args;
-    _requestCount = 0;
-    _suppressCommandPort = false;
-    _suppressCommandPortManualOverride = false;
+    : STCPServer(""), _args(args), _requestCount(0), _writeThread(nullptr), _replicationState(SQLC_SEARCHING),
```
quinthar: @tylerkaraszewski What do you think of getting more rigorous about the use of constructor initializer lists?

quinthar:
Wow, thanks for this great submission!  I've added some comments.

BedrockPlugin.cpp
Vadius: While we're wainitg for Tyler:
How do your old school see to add `const` to definitely const methods like [this](https://github.com/Expensify/Bedrock/blob/master/libstuff/libstuff.h#L328)?
And make return values `size_t` for methods returning size values, like [this](https://github.com/Expensify/Bedrock/blob/master/libstuff/libstuff.h#L139)?

quinthar:
Well, I don't think that can be a const method, because you are actively
locking something.  I'm not sure why anybody would either lock a constant
value (no need to lock it), or make the lock itself constant (that would
imply it couldn't be locked).

As for size_t, that's a bigger debate.  I personally hate all these silly
little types and wish everybody just used `int` for about everything.  I
think if I were to write it all over again using new C++, I might just use
`auto` everywhere (though I wish they had named it something like `var` to
make it a bit less weird looking).  The upshot is I personally have an
uneasy truce with size_t -- I'm not eager to go through and remove it
everywhere, but nor am I eager to expand its use.

Vadius:
Ok, I see. Let the types and modifiers leave untouched =)
Asked to just don't do excessive work.

BedrockPlugin.cpp
tylerkaraszewski: I actually like `nullptr`. It's a new C++11 feature that actually is semantically significant, as opposed to the old `NULL` macro, that I agree didn't add much.

Here's a description of `nullptr`: http://www.devx.com/cplus/10MinuteSolution/35167/0/page/2

I do like making methods `const` when they can be, because if nothing else it lets you call them from other `const` methods, but agree with @quinthar that in this case it's not applicable because of the mutex that needs to be locked.

I don't have strong opinions on `size_t`.

BedrockServer.cpp
tylerkaraszewski: Sure, should be slightly faster. Mostly I don't have a particularly strong opinion, but no reason not to.

quinthar:
Ah, I didn't realize nullptr was strongly typed.  That's sorta interesting
then, though only sorta.  Anyway, I'll defer to Tyler here.

Tyler - Can you do a final review and merge if you're satisfied?

On Thu, Jan 12, 2017 at 7:48 PM, Tyler Karaszewski <notifications@github.com

PR #96, opened 2017-01-28
Add a new command to query for a job based on it's name

tgolen:
```
vagrant@vagrant-ubuntu-trusty-64:~$ nc localhost 8888
QueryJobByName
name: www-prod/UserActivityEvaluator?ownerAccountID=15

200 OK
commitCount: 361
nodeName: bedrock
processingTime: 1
totalTime: 1
waitTime: 0
Content-Length: 216

{"created":"2017-01-28 13:29:01","data":{"accountID":15},"jobID":11,"lastRun":"","name":"www-prod\/UserActivityEvaluator?ownerAccountID=15","nextRun":"2017-01-28 13:29:01","repeat":"STARTED, +1 DAY","state":"QUEUED"}
```

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -144,6 +144,57 @@ bool BedrockPlugin_Jobs::peekCommand(BedrockNode* node, SQLite& db, BedrockNode:
         content["repeat"] = result[0][6];
         content["data"] = result[0][7];
         return true; // Successfully processed
+    }
+
+    // ----------------------------------------------------------------------
+    else if (SIEquals(request.methodLine, "QueryJobByName")) {
+        // - QueryJobByName( name )
+        //
+        //     Returns all known information about a given job. If there are multiple jobs with this name, only the
+        //     first one is returned.
+        //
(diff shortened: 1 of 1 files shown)


tgolen:
This command is needed for the harpoon 2.0 project because I need to be able to tell if a user has a job existing for them, and I don't have a jobID to use `QueryJob()`

tgolen:
Closing this in favor of using the DB plugin and keeping this simple.

quinthar:
@mcnamamj @tylerkaraszewski - Curious for your thoughts on this.  I had a quick conversation with @tgolen and encouraged him to just use the `Bedrock::DB` interface and query the jobs table directly, basically with the goal of avoiding a bunch of specialized "query" functions and instead just maintaining a very simple internal datastructure (eg, the `jobs` table) that is publicly exposed.  This way you can do whatever queries you want (eg, to create alerts on how many jobs of a particular type, or maybe the oldest of a particular priority, look up the data of a given job, etc).  I know it's a little unorthodox, but I'm curious if you're ok with this approach.

PR #105, opened 2017-02-09
[HOLD] Remove the priority hacks from the GetJob query

cead22:
Original PR https://github.com/Expensify/Server-Expensify/pull/1589/files

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -250,14 +250,6 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
             }
         }
 
-        // We'd initially intended for any value to be allowable here, but for performance reasons, we currently
-        // will only allow specific values to try and keep queries fast. If you pass an invalid value, we'll throw
-        // here so that the caller can know that he did something wrong rather than having his job sit unprocessed
-        // in the queue forever. Hopefully we can remove this restriction in the future.
-        if (priority != 0 && priority != 500 && priority != 1000) {
-            throw "402 Invalid priority value";
-        }
-
         // Create this new job
         db.write("INSERT INTO jobs ( created, state, name, nextRun, repeat, data, priority, parentJobID ) "
                  "VALUES( " +
@@ -286,28 +278,13 @@ bool BedrockPlugin_Jobs::processCommand(BedrockNode* node, SQLite& db, BedrockNo
         SQResult result;
         const string& name = request["name"];
 
-        string query =
-            "SELECT jobID, name, data FROM ( "
-            "SELECT * FROM "
-            "(SELECT jobID, name, data, priority "
-            "FROM jobs "
-            "WHERE state='QUEUED' "
-            "  AND " +
-            SCURRENT_TIMESTAMP() + ">=nextRun " + "  AND name GLOB " + SQ(name) + " " + "  AND priority=" + SQ(1000) +
-            " " + "ORDER BY nextRun ASC " + "LIMIT 1) " + "UNION ALL " + "SELECT * FROM "
-                                                                         "(SELECT jobID, name, data, priority "
-                                                                         "FROM jobs "
-                                                                         "WHERE state='QUEUED' "
-                                                                         "  AND " +
-            SCURRENT_TIMESTAMP() + ">=nextRun " + "  AND name GLOB " + SQ(name) + " " + "  AND priority=" + SQ(500) +
-            " " + "ORDER BY nextRun ASC " + "LIMIT 1) " + "UNION ALL " + "SELECT * FROM "
-                                                                         "(SELECT jobID, name, data, priority "
-                                                                         "FROM jobs "
-                                                                         "WHERE state='QUEUED' "
-                                                                         "  AND " +
-            SCURRENT_TIMESTAMP() + ">=nextRun " + "  AND name GLOB " + SQ(name) + " " + "  AND priority=" + SQ(0) +
-            " " + "ORDER BY nextRun ASC " + "LIMIT 1) " + ") " + "ORDER BY priority DESC "
-                                                                 "LIMIT 1;";
+        string query = "SELECT jobID, name, data "
+					   "FROM jobs "
+					   "WHERE state='QUEUED' "
+					   "  AND " + SCURRENT_TIMESTAMP() + ">=nextRun " +
+					   "  AND name GLOB " + SQ(name) + " " +
+					   "ORDER BY priority DESC, nextRun ASC "
+					   "LIMIT 1;";
 
         if (!db.read(query, result)) {
             throw "502 Query failed";


quinthar:
Are we sure these aren't needed?  Why did we do them in the first place, and are we sure that was a mistake?  I'd suggest doing some performance testing on the queries before/after.

cead22:
That is the plan and that's why the PR is on HOLD with no requests for reviews yet. I'm planning on running the query at the times where we have a significant amount of jobs that are QUEUED
- 2AM UTC: lots of policy harvester jobs queued
- 5AM UTC: lots of user harvester jobs queued
- 7:30AM UTC: lots of report harvester jobs queued

cead22:
We did them in the first place because we thought that sqlite couldn't run the original query quickly enough but at the same time we added some indices.

The indices seem to have solved the problem without the need for the modified query. Ionatan ran the original query after we added the indices and realized that it ran quick and that perhaps we don't need to split the queries for the different priorities anymore.

cead22:
# 2017-02-09 10:30:00
```
Query:SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-09 10:30:00' >= nextRun  AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;

200 OK
commitCount: 257566517
nodeName: www1.la
processingTime: 19
totalTime: 27
waitTime: 7
Content-Length: 1


Query:SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-09 10:30:00' >= nextRun  AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;

200 OK
commitCount: 257566612
nodeName: www1.la
processingTime: 3
totalTime: 4
waitTime: 0
Content-Length: 1


Query:SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-09 10:30:00' >= nextRun  AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;

200 OK
commitCount: 257566698
nodeName: www1.la
processingTime: 2
totalTime: 4
waitTime: 1
Content-Length: 1
```

---

# 2017-02-08 10:30:00
```
Query:SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-08 10:30:00' >= nextRun  AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;

200 OK
commitCount: 257568750
nodeName: www1.la
processingTime: 3
totalTime: 3
waitTime: 0
Content-Length: 1


Query:SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-08 10:30:00' >= nextRun  AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;

200 OK
commitCount: 257568763
nodeName: www1.la
processingTime: 4
totalTime: 4
waitTime: 0
Content-Length: 1


Query:SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-08 10:30:00' >= nextRun  AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;

200 OK
commitCount: 257568772
nodeName: www1.la
processingTime: 6
totalTime: 6
waitTime: 0
Content-Length: 1
````

**Comment by quinthar** [comment]
What conclusions do you take away from this?

**Comment by cead22** [comment]
The queries all ran very quickly. I'm hoping they run very quickly when I run it at the other times specified above, in which case I'd move forward with this PR.

If you have any other tests in mind don't hesitate to share them and I'll be happy to run them

**Comment by quinthar** [comment]
Ah, cool.  Hm, well I'd first start by comparing the query plans to see how the queries differ.  For posterity, here's the EXPLAIN output for that new query:

```
dbarrett@www2:~$ sudo readdb.sh "explain query plan SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-08 10:30:00' >= nextRun  AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;"
----------  ----------  ----------  --------------------------------------------------------------------------
0           0           0           SEARCH TABLE jobs USING INDEX jobsStateNextRunName (state=? AND nextRun<?)
0           0           0           USE TEMP B-TREE FOR ORDER BY                                              
dbarrett@www2:~$
```

So this means it'll use the index for the `state=` and `nextrun<` clauses in the `WHERE`.  However, it'll then need to do a manual scan over the remainder for the `name glob 'www-prod/*'`, and then sort the results by priority in a temporary table in order to find the highest priority to return.

On the other hand, here's the plan for if we do an exact match on the priority:

```
dbarrett@www2:~$ sudo readdb.sh "explain query plan SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2017-02-08 10:30:00' >= nextRun  AND name glob 'www-prod/*' and priority=0 limit 1;"
----------  ----------  ----------  -------------------------------------------------------------------------------------------------
0           0           0           SEARCH TABLE jobs USING INDEX jobsStatePriorityNextRunName (state=? AND priority=? AND nextRun<?)
dbarrett@www2:~$ 
```

You can see that it's able to use a different index, meaning it matches both `state=` and `priority=` exactly.  This takes it directly to a group of rows in the index pre-sorted by `nextRun`, meaning it only needs to search down that list until it finds a row matching the `GLOB`.

The upshot is the original query should execute much, much faster than the new one when there are a large number of tasks that match the GLOB.  The original will only need to linearly search down the index until it finds one matching result.  The new query will need to find *every* matching result, put it into a temporary table, sort that table (which is O(log n)), and then take the top.

To simulate a large result set, I've updated the date to be way in the future, forcing it to match a huge number of rows.  Here's the time difference of the two queries in this scenario:

```
query: SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2018-01-01' >= nextRun AND name glob 'www-prod/*' and priority=500 limit 1;

200 OK
commitCount: 258015997
nodeName: www2.lv
processingTime: 4
totalTime: 5
waitTime: 1
Content-Length: 156

jobID | name | data
16362657 | www-prod/PolicyHarvester | {"policyID":"014980AEF3B30B97","harvestingGUID":"014980AEF3B30B97589d29c1924a0","isWaiting":true}

query: SELECT jobID, name, data FROM jobs WHERE state='QUEUED' AND '2018-01-01' >= nextRun AND name glob 'www-prod/*' ORDER BY priority DESC, nextRun ASC limit 1;

200 OK
commitCount: 258013988
latency: 
nodeName: www2.lv
processingTime: 16480
totalTime: 16481
waitTime: 0
Content-Length: 237

jobID | name | data
33395043 | www-prod/ChatbotIntercomRespond | {"chatID":116267,"adminID":316825,"body":"Just wanted to let you know that I've taken care of that Taxa 4x35 receipt you SmartScanned. Keep \u2018em coming!","close":true}
```

In conclusion, the old query runs in about 4ms, while the new query runs in about 16480 ms.  Given this, I'm still pretty skeptical about unwinding this change.  I think we should just add a fourth priority to Bedrock::Jobs and then use that.

iwiznia:
Wow, so I was totally wrong. 
I think that this would've worked if we still had the expression index, that we removed [here](https://github.com/Expensify/Server-Expensify/pull/1560).

cead22:
> I think we should just add a fourth priority to Bedrock::Jobs and then use that.

👍 Thanks for the thorough explanation

quinthar:
Do we really even need the fourth priority, or are the current three
adequate?  I mean, we can add a fourth, it's easy, I just wonder if that's
a real requirement.

quinthar:
There's something appealing about HIGH, MEDIUM, LOW.

cead22:
No.

Happy to continue this conversation on the design doc or the issue (since that's where this came from) but per my email when re-sending the doc, we can just merge the current two lowest priorities into one and have high, medium, low.

PR #116, opened 2017-02-27
Updating comments to clarify my understanding

quinthar:
Here are some comments I'd like to merge into https://github.com/Expensify/Bedrock/pull/112 to help document (and confirm) my understanding.

@tylerkaraszewski - Can you please review, confirm the comments are accurate, and merge?

### DIFF
--- a/BedrockNode.cpp
+++ b/BedrockNode.cpp
@@ -72,7 +72,10 @@ BedrockNode::~BedrockNode() {
 }
 
 bool BedrockNode::_passToExternalQueue(Command* command) {
+    // Check to see if we have a server
+    // **FIXME: Given that this->server is defined in the constructor, how can this be false?
     if (server) {
+        // No server, which probably means REASON
         server->enqueueCommand(command);
         return true;
     }
@@ -144,6 +147,8 @@ bool BedrockNode::_peekCommand(SQLite& db, Command* command) {
 }
 
 void BedrockNode::_setState(SQLCState state) {
+    // When we change states, our schema might change - note that we need to
+    // wait for any possible upgrade to complete again.
     _dbReady = false;
     SQLiteNode::_setState(state);
 }
@@ -158,14 +163,14 @@ bool BedrockNode::_processCommand(SQLite& db, Command* command) {
     SData& response = command->response;
     STable& content = command->jsonContent;
     SDEBUG("Received '" << request.methodLine << "'");
-
     bool needsCommit = false;
-
     try {
         if (SIEquals(request.methodLine, "UpgradeDatabase")) {
+            // Begin a non-concurrent transaction for database upgrading
             if (!db.beginTransaction()) {
                 throw "501 Failed to begin transaction";
             }
+
             // Loop across the plugins to give each an opportunity to upgrade the
             // database.  This command is triggered only on the MASTER, and only
             // upon it step up in the MASTERING state.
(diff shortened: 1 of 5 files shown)


quinthar:
Ok, I wanted to tackle this one separately as it's a sort of big design change.  Basically, the logic of the worker thread is very hard for me to follow.  There's a hack's nest of incestuous logic between `createCommand()`, `openCommand()`, `_openCommand()`, and `reopenCommand()` that I'm having a hard time teasing apart.  What would you think about refactoring out something like `SQLiteCore` that is just the command-processing part of `SQLiteNode`, such that the "worker" thread wouldn't need a full `SQLiteNode` object.  This way `SQLiteNode` can focus exclusively on the clustered state machine and replication logic, and wouldn't need to get so hacked up to handle concurrent commits.  I'm thinking this `SQLiteCore` object would be something like:

```
// Responsible for concurrent, multi-threaded command processing on a local
// database.  Is aware of plugins, clients, and remote servers (eg,
// HTTPSRequest), but is not at all aware of replication or peers.
class SQLiteCore {
    struct Command {
        // Everything currently in SQLiteNode::Command, though I'd move
        // "initiator" into some general purpose "data" SData that allows the
        // caller to attach arbitrary data to thi without violating the
        // encapsulation of this layer
    };

    // Allocates a new command
    static Command* openCommand(const SData& request);

    // Performs read-only actions for this command.  Returns true if the
    // command was fully handled no more processing.
    static bool peekCommand(SQLite& db, Command* command);

    // Potentially performs read/write actions for this command.  Returns true
    // if it made any changes that need to be committed to the database.
    static bool processCommand(SQLite& db, Command* command);

    // De-allocates a command
    static void closeCommand(Command* command);
};  
```

This would also allow us to replace our `SData` queues with global `SQLiteCore::Command*` queues.  Finally, with this, we could "unroll" a lot of the complex worker logic into something like this:

```
void WorkerThread() {
    // Open up the right database
    SQLite db(... figure out its threadID and such...);

    // Command loop
    while(... not shutting down ...) {
        // Wait for work
        ...

        // Work
        while (SQLiteCore::Command* command = data.escalatedCommands.pop() ?: data.queuedCommands.pop()) {
            // See if it only needs peeking
            if (SQLiteCore::peekCommand(db, command)) {
                // All done!  No more processing required.  Where did this come from?
                if (command->data.test("escalated")) {
                    // This was escalated to us, send to sync thread
                    data.syncCommands.push(command);
                } else {
                    // This came from a client, respond
                    data.processedCommands.push(command);
                }
            } else {
                // Needs more processing.  Do we want to let the workert thread do it?
                bool async = SIEquals(command->data["writeConsistency"], "ASYNC");
                if (_syncNode->dbReady() && async && !command->httpsRequest) {
                    // Try a few times to process concurrently
                    int tries = MAX_ASYNC_CONCURRENT_TRIES;
                    while (tries--) { 
                        // Begin a concurrent transaction and process 
                        db.beginConcurrent();
                        SASSERT(command.response.empty());
                        bool needToCommit = SQLiteCore::processCommand(db, command);
                        SASSERT(!command.response.empty());

                        // Do we need to commit anything?
                        if (!needToCommit) {
                            // Nothing to commit
                            break;
                        }

                        // This changed the database; commit
                        if (db.commit()) {
                            // Successful commit!
                            break;
                        }

                        // Commit failed, try again
                        command.response.clear();
                        db.rollback();
                    }
                }

                // Either we didn't try, tried and succeed, or tried and failed.
                // Regardless of which, send it to the sync thread to sort out.
                data.syncCommands.push(command);
            }
        }
    }
}
```

I imagine this is a pretty substantial refactoring on what's already a ton of code changes.  But if we're going to go through all this effort, let's make sure we are leaving the code *better* than we found it.

PR #169, opened 2017-05-04
Add `retryAfter` to Bedrock::Jobs::GetJob 

cead22:
@quinthar please review

- Fixes https://github.com/Expensify/Bedrock/issues/111
- I believe this captures everything in the issue except the part about `lastScheduled = nextRun-retryAfter;` per our offline agreement to punt implementing auto-retrying repeatable jobs when/if we need to

# Tests
## Error cases / safety checks
```
createjob
name:test
repeat:SCHEDULED, +1 HOUR
retryAfter:10

402 Recurring auto-retrying jobs are not supported
commitCount: 10745
Content-Length: 0

createjob
name:test
retryAfter:10

402 Malformed retryAfter
commitCount: 10745
Content-Length: 0

createjob
name:test
retryAfter:+10 HOUR
unique:true

405 Unique jobs can't be retried
commitCount: 10745
Content-Length: 0
```

## Lifecycle of a retryAble job
```
createjob
name:testRetryable
retryAfter:+20 SECONDS

200 OK
commitCount: 10745
Content-Length: 19

{"jobID":103374524}

query:select * from jobs where jobid=103374524

200 OK
commitCount: 10746
Content-Length: 192

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID
2017-05-13 01:09:11 | 103374524 | QUEUED | testRetryable | 2017-05-13 01:09:11 |  |  | {} | 500 | 0

getjob
name:testRetryable

200 OK
commitCount: 10746
Content-Length: 52

{"data":{},"jobID":103374524,"name":"testRetryable"}
query:select * from jobs where jobid=103374524

200 OK
commitCount: 10747
Content-Length: 241

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-05-13 01:09:11 | 103374524 | RUNQUEUED | testRetryable | 2017-05-13 01:09:53 | 2017-05-13 01:09:33 |  | {} | 500 | 0 | +20 SECONDS

getjob
name:testRetryable

404 No job found
commitCount: 10747
Content-Length: 0

getjob
name:testRetryable

200 OK
commitCount: 10747
Content-Length: 52

{"data":{},"jobID":103374524,"name":"testRetryable"}

getjob
name:testRetryable

404 No job found
commitCount: 10748
Content-Length: 0

updatejob
jobid:103374403

402 Missing data
commitCount: 10748
Content-Length: 0

updatejob
jobid:103374524
data:{}

[...]

PR #195, opened 2017-05-23
Add Bedrock::Jobs::CreateJobs

pecanoro:
@cead22 can you review this please?

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/49375

### Tests

```
createJobs  
jobs: [{"name":"testCreate", "data":{"blabla":"blabla"}, "parentJobID":"1000", "repeat": "SCHEDULED, +1 HOUR"}, {"testCreate2", "data":{"nope":"nope"}}]

401 Invalid JSON
commitCount: 16088
peekTime: 1200
totalTime: 1563
Content-Length: 0

createJobs
jobs: [{"name":"testCreate", "data":{"blabla":"blabla"}, "parentJobID":"3847", "repeat": "SCHEDULED, +1 HOUR"}, {"name":"testCreate2", "data":{"nope":"nope"}}]

405 Can only create child job when parent is RUNNING or PAUSED
commitCount: 15892
Content-Length: 0

createJobs
jobs: [{"name":"testCreate", "data":{"blabla":"blabla"}, "parentJobID":"3845", "repeat": "SCHEDULED, +1 HOUR"}, {"name":"testCreate2", "data":{"nope":"nope"}}]

200 OK
commitCount: 15894
Content-Length: 22

{"jobIDs":[3852,3853]}

query:select * from jobs where jobID IN (3852, 3853);

200 OK
commitCount: 15895
Content-Length: 355

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID 
2017-05-23 23:50:36 | 3852 | PAUSED | testCreate | 2017-05-23 23:50:36 |  | SCHEDULED, +1 HOUR | {"blabla":"blabla"} | 500 | 3845 
2017-05-23 23:50:36 | 3853 | QUEUED | testCreate2 | 2017-05-23 23:50:36 |  |  | {"nope":"nope"} | 500 | 0 

```

```
createJob
name:testCreate

200 OK
commitCount: 15886
Content-Length: 14

{"jobID":3844}

query:select * from jobs where jobID=3844; 

200 OK
commitCount: 15887
Content-Length: 200

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-05-23 22:28:49 | 3844 | QUEUED | testCreate | 2017-05-23 22:28:49 |  |  | {} | 500 | 0 | 

createJob
name:testCreate
priority: 1000

200 OK
commitCount: 15887
Content-Length: 14

{"jobID":3845}

query:select * from jobs where jobID=3845; 

200 OK
commitCount: 15888
Content-Length: 201

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-05-23 22:30:28 | 3845 | QUEUED | testCreate | 2017-05-23 22:30:28 |  |  | {} | 1000 | 0 | 

createJob
name:testCreate
repeat:blabla

402 Malformed repeat
commitCount: 1062
Content-Length: 0

createJob
name:testCreate
data:{"blabla":"blabla"}          

200 OK
commitCount: 15888
Content-Length: 14

{"jobID":3846}

query:select * from jobs where jobID=3846     
    
200 OK
commitCount: 15889
Content-Length: 217

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID
2017-05-23 22:36:56 | 3846 | QUEUED | testCreate | 2017-05-23 22:36:56 |  |  | {"blabla":"blabla"} | 500 | 0 

createJob
name:testCreate
repeat:SCHEDULED, +1 HOUR

200 OK
commitCount: 15889
Content-Length: 14

{"jobID":3847}

query:select * from jobs where jobID=3847;

200 OK
commitCount: 15890
Content-Length: 218

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID
2017-05-23 22:40:02 | 3847 | QUEUED | testCreate | 2017-05-23 22:40:02 |  | SCHEDULED, +1 HOUR | {} | 500 | 0

createJob                                                                       
name:testCreate

200 OK
commitCount: 15891
Content-Length: 14

{"jobID":3849}

query:select * from jobs where jobID=3849;

200 OK
commitCount: 15892
Content-Length: 208

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID
2017-05-23 22:42:33 | 3849 | QUEUED | testCreate | 2017-05-23 22:42:33 |  |  | {} | 500 | 0
```

```
createJob
name:blabla   
unique:true

200 OK
commitCount: 16038
peekTime: 3072
processTime: 1522
totalTime: 11765
Content-Length: 12

{"jobID":16}

query:select * from jobs where jobID IN (16);        

200 OK
commitCount: 16039
peekTime: 5246
totalTime: 5351
Content-Length: 194

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-06-12 19:11:28 | 16 | QUEUED | blabla | 2017-06-12 19:11:28 |  |  | {} | 500 | 0 | 

createJob
name:blabla
unique:true
data:blabla

402 Data is not a valid JSON Object
commitCount: 16040
peekTime: 158
totalTime: 263
Content-Length: 0

createJob
name:blabla
unique:true
data:{"blabla":"test"} 

200 OK
commitCount: 16040
peekTime: 3532
processTime: 1352
totalTime: 9811
Content-Length: 12

{"jobID":16}

query:select * from jobs where jobID IN (16);

200 OK
commitCount: 16041
peekTime: 5425
totalTime: 5527
Content-Length: 209

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-06-12 19:11:28 | 16 | QUEUED | blabla | 2017-06-12 19:11:28 |  |  | {"blabla":"test"} | 500 | 0 | 

```


pecanoro:
Oh shit, it has conflicts because of `retryAfter` being reverted

pecanoro:
Updated!! I removed `retryAfter` from the code but I left the other optional parameters even if master doesn't have them after reverting, not sure what to do otherwise.

plugins/Jobs.cpp
```diff
@@ -160,32 +160,50 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
     
     // ----------------------------------------------------------------------
-    else if (SIEquals(request.methodLine, "CreateJob")) {
-        // If unique flag was passed and the job exist in the DB, then we can
-        // finish the command without escalating to master.
-        if (!request.test("unique")) {
-            // Not unique; need to process
-            return false;
+    else if (SIEquals(request.methodLine, "CreateJob") || SIEquals(request.methodLine, "CreateJobs")) {
+        vector<STable> jsonJobs;
```
cead22: Why vector over list?

plugins/Jobs.cpp
```diff
@@ -160,32 +160,50 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
-            throw "502 Select failed";
-        }
-        // If there's no job, or the existing job doesn't match the data we've been passed, escalate to master.
-        if (result.empty() || result[0][1] != request["data"]) {
-            return false;
-        }
+        for (auto& job : jsonJobs) {
+            // If unique flag was passed and the job exist in the DB, then we can
+            // finish the command without escalating to master.
+            if (!SContains(job ,"unique") || job["unique"] == "false") {
```
cead22: Can't you use `if (job.test("unique") {`? 
I think this would be incorrect for a job that has the string `unique` in its name for instance

plugins/Jobs.cpp
```diff
@@ -160,32 +160,50 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
+
+            // Verify unique
+            SQResult result;
+            SINFO("Unique flag was passed, checking existing job with name " << job["name"]);
+            if (!db.read("SELECT jobID, data "
+                         "FROM jobs "
+                         "WHERE name=" + SQ(job["name"]) + ";",
+                         result)) {
+                throw "502 Select failed";
+            }
```
cead22: Add empty line before the comment

plugins/Jobs.cpp
```diff
@@ -160,32 +160,50 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
+            // If there's no job, escalate to master.
+            if (result.empty()){
+                return false;
```
cead22: This is escalating to master based on checks for a single job, but we could check that the data passed is correct for all the jobs before escalating to master to prevent escalating a command that we know will fail

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+        //
+        //     Parameters:
+        //     - jobs (json array):
+        //          - name  - An arbitrary string identifier (case insensitive)
+        //          - data  - A JSON object describing work to be done (optional)
+        //          - firstRun - A "YYYY-MM-DD HH:MM:SS" datetime of when
+        //                  this job should next execute (optional)
+        //          - repeat - A description of how to repeat (optional)
+        //          - priority - High priorities go first (optional, default 500)
+        //          - unique - if true, it will check that no other job with this name already exists, if it does it will
```
cead22: Put the docs for unique and firstrun in one line, or use the same formatting for both

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-            SQResult result;
-            SINFO("Unique flag was passed, checking existing job with name " << request["name"]);
-            if (!db.read("SELECT jobID, data "
-                         "FROM jobs "
-                         "WHERE name=" + SQ(request["name"]) + ";",
-                         result)) {
-                throw "502 Select failed";
+        vector<STable> jsonJobs;
+        if (SIEquals(request.methodLine, "CreateJob")) {
+            verifyAttributeSize(request, "name", 1, MAX_SIZE_SMALL);
```
cead22: Shouldn't this be in peek? and we should ensure we have job names for the CreateJobs command too

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-        }
+                // If we got a result, and it's data is the same as passed, we won't change anything.
+                if (!result.empty() && result[0][1] == job["data"]) {
+                    SINFO("Job already existed with matching data, and unique flag was passed, reusing existing job "
+                          << result[0][0]);
 
-        // If no "firstRun" was provided, use right now
-        const string& safeFirstRun = request["firstRun"].empty() ? SCURRENT_TIMESTAMP() : SQ(request["firstRun"]);
+                    // If we are calling GetJob, return early, there are no more jobs to create.
+                    if(SIEquals(request.methodLine, "CreateJob")){
```
cead22: Style `if (SIEquals(request.methodLine, "CreateJob")) {`

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-            }
-        }
+                // If we got a result, and it's data is the same as passed, we won't change anything.
+                if (!result.empty() && result[0][1] == job["data"]) {
+                    SINFO("Job already existed with matching data, and unique flag was passed, reusing existing job "
+                          << result[0][0]);
 
-        // If no "firstRun" was provided, use right now
-        const string& safeFirstRun = request["firstRun"].empty() ? SCURRENT_TIMESTAMP() : SQ(request["firstRun"]);
+                    // If we are calling GetJob, return early, there are no more jobs to create.
```
cead22: `// If we are calling CreateJob...` (not GetJob)

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+                    if(SIEquals(request.methodLine, "CreateJob")){
+                        content["jobID"] = result[0][0];
+                        return true;
+                    }
 
-        // If no data was provided, use an empty object
-        const string& safeData = request["data"].empty() ? SQ("{}") : SQ(request["data"]);
+                    // Append new jobID to list of created jobs.
+                    jobIDs.push_back(result[0][0]);
+                    continue;
```
cead22: Seems like with this continue we'll skip updating the updateJobID variable below and thus we won't update the job. Am I reading this correctly?

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
             }
-        }
 
+            // If no "firstRun" was provided, use right now
+            const string& safeFirstRun = !SContains(job, "firstRun") || job["firstRun"].empty() ? SCURRENT_TIMESTAMP() : SQ(job["firstRun"]);
 
-        // If no priority set, set it
-        int64_t priority = request.isSet("priority") ? request.calc("priority") : JOBS_DEFAULT_PRIORITY;
+            // If no data was provided, use an empty object
+            const string& safeData = !SContains(job, "data") || job["data"].empty() ? SQ("{}") : SQ(job["data"]);
```
cead22: Why did you add SContains to all the checks of this type? isn't `job["data"].empty()` enough here?

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
             if (parentJobID) {
-                auto parentState = db.read("SELECT state FROM jobs WHERE jobID=" + SQ(parentJobID) + ";");
-                if (SIEquals(parentState, "RUNNING")) {
-                    initialState = "PAUSED";
+                SQResult result;
+                if (!db.read("SELECT state FROM jobs WHERE jobID=" + SQ(parentJobID) + ";", result)) {
+                    throw "502 Select failed";
+                }
+                if (result.empty()) {
+                    throw "404 parentJobID does not exist";
```
cead22: Let's make sure to fail this case on peek to avoid creating some jobs and not others

plugins/Jobs.cpp
```diff
@@ -220,128 +238,192 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+                               "repeat   = " + SQ(SToUpper(job["repeat"])) + ", " +
+                               "data     = JSON_PATCH(data, " + safeData + "), " +
+                               "priority = " + SQ(priority) + " " +
+                             "WHERE jobID = " + SQ(updateJobID) + ";"))
+                {
+                    throw "502 update query failed";
+                }
+
+                // If we are calling GetJob, return early, there are no more jobs to create.
+                if(SIEquals(request.methodLine, "CreateJob")){
```
cead22: Style `if (SIEquals(request.methodLine, "CreateJob")) {`

plugins/Jobs.cpp
pecanoro: Because when I created it, I thought I was going to need to access the elements in the vector and I cannot do that with lists. Anyways, I checked and there shouldn't be any difference to choose one over the other one. Do you want me to change it?

plugins/Jobs.cpp
pecanoro: What? I am not following you. `SContains` can be used for maps

plugins/Jobs.cpp
pecanoro: You misread this haha. That line is only reached when the data is the same as the existing data for the unique job, so we don't need to update, we just return the jobID.

plugins/Jobs.cpp
pecanoro: When we use `CreateJob`, we are sure that all parameters are being sent by PHP, but I cannot assure the same is happening when we receive a JSON Array with multiples jobs, so if I try accessing an element that doesn't exist in the job map, I think it will be returning an `out_of_range ` exception.

plugins/Jobs.cpp
pecanoro: Ok

plugins/Jobs.cpp
cead22: Sorry, I deleted the comment by mistake trying to undo a new review and adding a single comment.
Re: changing vector to list, yes we use list everywhere

plugins/Jobs.cpp
cead22: You're right, should've read the docs we had an issue recently with SContains matching partial strings.
You should still use `job.test` (you can, right?) even if you keep the `SContains` condition

plugins/Jobs.cpp
cead22: My bad, I checked the diff like 4 times with and without `?w=1` so I decided to ask. Thanks for clarifying

plugins/Jobs.cpp
cead22: Oh wow this is a bug on GH actually, I deleted my last comment (again, after it was added as a new review) and it deleted the whole thread.

I didn't know about `SContains(map, string)` thanks for clarifying!

pecanoro:
Updated!

plugins/Jobs.cpp
pecanoro: Done!

pecanoro:
Updated!

pecanoro:
bump

plugins/Jobs.cpp
```diff
@@ -160,33 +160,72 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
+            // Validate that the parentJobID exists and is in the right state if one was passed.
+            int64_t parentJobID = SContains(job, "parentJobID") ? SToInt(job["parentJobID"]) : 0;
+            if (parentJobID) {
+                SQResult result;
+                if (!db.read("SELECT state FROM jobs WHERE jobID=" + SQ(parentJobID) + ";", result)) {
+                    throw "502 Select failed";
+                }
+                if (result.empty()) {
+                    throw "404 parentJobID does not exist";
+                } else if (!SIEquals(result[0][0], "RUNNING") && !SIEquals(result[0][0], "PAUSED")) {
```
cead22: I think we should make this an if instead of else/if for consistency with the code around it.
```
if (C) {
  throw;
}

if (C2) {
  throw;
}

if (C3) {
  throw;
}

// as opposed to
if (C) {
  throw;
}

if (C2) {
  throw;
} else if (C3) {
  throw;
}
```

plugins/Jobs.cpp
```diff
@@ -201,147 +240,208 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-            // Update the existing job.
-            if(!db.write("UPDATE JOBS SET "
-                           "repeat   = " + SQ(SToUpper(request["repeat"])) + ", " +
-                           "data     = JSON_PATCH(data, " + safeData + "), " +
-                           "priority = " + SQ(priority) + " " +
-                         "WHERE jobID = " + SQ(updateJobID) + ";"))
-            {
-                throw "502 update query failed";
+            // If no priority set, set it
+            int64_t priority = SContains(job, "priority") ? SToInt(job["priority"]) : JOBS_DEFAULT_PRIORITY;
```
cead22: Can you use `job.isSet("priority")` and `job.calc["priority"]` instead of ` SContains(job, "priority")` and  `SToInt(job["priority"])`?

plugins/Jobs.cpp
```diff
@@ -201,147 +240,208 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-        } else {
-            // Normal jobs start out in the QUEUED state, meaning they are ready to run immediately.
-            // Child jobs normally start out in the PAUSED state, and are switched to QUEUED when the parent
-            // finishes itself (and itself becomes PAUSED).  However, if the parent is already PAUSED when
-            // the child is created (indicating a child is creating a sibling) then the new child starts
-            // in the QUEUED state.
-            auto initialState = "QUEUED";
+
+            // Validate that the parentJobID exists and is in the right state if one was passed.
+            int64_t parentJobID = SContains(job, "parentJobID") ? SToInt(job["parentJobID"]) : 0;
```
cead22: Same note about isSet/calc as above

plugins/Jobs.cpp
```diff
@@ -201,147 +240,208 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-                auto parentState = db.read("SELECT state FROM jobs WHERE jobID=" + SQ(parentJobID) + ";");
-                if (SIEquals(parentState, "RUNNING")) {
-                    initialState = "PAUSED";
+                SQResult result;
+                if (!db.read("SELECT state FROM jobs WHERE jobID=" + SQ(parentJobID) + ";", result)) {
+                    throw "502 Select failed";
+                }
+                if (result.empty()) {
+                    throw "404 parentJobID does not exist";
+                } else if (!SIEquals(result[0][0], "RUNNING") && !SIEquals(result[0][0], "PAUSED")) {
```
cead22: Same not here as above about if vs else if

plugins/Jobs.cpp
cead22: Sorry I forgot to mention this before, but perhaps we should alert here instead of throwing since we have these checks in peekCommand and so we should never reach this point.

I added similar checks in the retryAfter [PR](https://github.com/Expensify/Bedrock/pull/169/files#r116340952) and opted to log an alert instead of throwing in processCommand.

What do you think?

pecanoro: Ok, I agree, but alert and continue? Is that safe enough? I mean, I know `we should never reach this point` but the mistakes happen.

pecanoro:
To your comments about why you got a different answer. My tests were performed before I had to remove `retryafter` from the code so both cases that you are mentioning used to return exceptions related to that parameters which is not here anymore.

plugins/Jobs.cpp
pecanoro: lol nope, I am not using `SData`anymore but `STable` instead

plugins/Jobs.cpp
pecanoro: Same as before

pecanoro:
Updated!

cead22:
Can you re-run the tests and update them in the description please?

plugins/Jobs.cpp
```diff
@@ -201,147 +241,209 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-                      << result[0][0]);
-                content["jobID"] = result[0][0];
-                return true;
-            }
+        list<string> jobIDs;
+        for (auto& job : jsonJobs) {
+            // If unique flag was passed and the job exist in the DB, then we can finish the command without escalating to
+            // master.
+            uint64_t updateJobID = 0;
+            if (SContains(job, "unique")) {
```
cead22: You should probably check that this value is true

pecanoro:
Updated!

plugins/Jobs.cpp
```diff
@@ -205,147 +250,209 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+                auto initialState = "QUEUED";
+                if (parentJobID) {
+                    auto parentState = db.read("SELECT state FROM jobs WHERE jobID=" + SQ(parentJobID) + ";");
+                    if (SIEquals(parentState, "RUNNING")) {
+                        initialState = "PAUSED";
+                    }
+                }
+                
+                // If no data was provided, use an empty object
+                const string& safeRetryAfter = job["retryAfter"].empty() ? "\"\"" : SQ(job["retryAfter"]);
```
cead22: I didn't catch this before but this is unused

pecanoro:
Updated!!

pecanoro:
Updated!!

cead22:
Everything looks good but I found a small bug, when you call CreateJob with unique:true and the data is the same as the data in the DB, the response is empty and doesn't contain the jobID, example:

```
query:select * from jobs;

200 OK
commitCount: 7848
peekTime: 3106
totalTime: 3166
Content-Length: 1


createjob
name:test
data:{"nope2":"nope1"}

200 OK
commitCount: 7848
peekTime: 3297
processTime: 1691
totalTime: 11430
Content-Length: 13

{"jobID":187}

createJobs
jobs: [{"name":"test", "data":{"nope2":"nope1"}, "unique":"true"}]

200 OK
commitCount: 7849
peekTime: 4512
processTime: 1253
totalTime: 6015
Content-Length: 0
```

pecanoro:
Updated!! I placed the returned jobID list inside the `for` loop by accident

plugins/Jobs.cpp
```diff
@@ -165,33 +165,89 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
+                }
+                if (!SIEquals(result[0][0], "RUNNING") && !SIEquals(result[0][0], "PAUSED")) {
+                    SWARN("Trying to create child job with parent jobID#" << parentJobID << ", but parent isn't RUNNING or PAUSED (" << result[0][0] << ")");
+                    throw "405 Can only create child job when parent is RUNNING or PAUSED";
+                }
+            }
+
+            // Verify unique
+            SQResult result;
+            SINFO("Unique flag was passed, checking existing job with name " << job["name"]);
```
fnwbr: @pecanoro @cead22 Since you removed https://github.com/Expensify/Bedrock/pull/195/files#diff-d116ebb4bb9e2a3fd799045c7b182707L171 we are always getting here, no?

plugins/Jobs.cpp
```diff
@@ -165,33 +165,89 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
+                         "FROM jobs "
+                         "WHERE name=" + SQ(job["name"]) + ";",
+                         result)) {
+                throw "502 Select failed";
+            }
+
+            // If there's no job or the existing job doesn't match the data we've been passed, escalate to master.
+            if (!result.empty() && result[0][1] == job["data"]){
+                // If we are calling CreateJob, return early, there are no more jobs to create.
+                if (SIEquals(request.methodLine, "CreateJob")) {
```
fnwbr: @pecanoro @cead22 Shouldn't we move this `if` statement a few lines up?
Like before the `// Verify unique` line, we don't need to verify that when creating multiple jobs, no?

plugins/Jobs.cpp
cead22: Looks like it, nice catch!

plugins/Jobs.cpp
cead22: Verify what?

fnwbr: What I'm asking is: Should we check for existing unique jobs in `peekCommand` when creating multiple jobs?

fnwbr: Or said in another way: I think we don't need to run the query in line 233 when using `CreateJobs` but only when using `CreateJob`

cead22: Oh I see. I guess not since there's no failure condition for it that would cause us to return early and not escalate

botify:

PR #223, opened 2017-06-16
Check the request's verb, not methodline

mea36:
Use the verb in the methodline to determine if the plugin supports this command

This is needed for https://github.com/Expensify/Expensify/issues/55620

### DIFF
--- a/plugins/Cache.cpp
+++ b/plugins/Cache.cpp
@@ -70,7 +70,7 @@ BedrockPlugin_Cache::~BedrockPlugin_Cache() {
 }
 
 // ==========================================================================
-void BedrockPlugin_Cache::initialize(const SData& args) {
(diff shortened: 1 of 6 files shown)


mea36:
Putting this on hold, I should update the WriteTest as well

mea36:
Updated and off hold

mea36:
Updated - I replaced usages of `request.methodLine` with `getVerb()` in the Jobs and Cache plugin
I didn't replace the one usage in the MySQL plugin, there's not documentation on how to use it

The output of my tests for the Jobs and Cache plugins can be found below
[curl.txt](https://github.com/Expensify/Bedrock/files/1099235/curl.txt)
[nc.txt](https://github.com/Expensify/Bedrock/files/1099234/nc.txt)

PR #241, opened 2017-07-10
Make SContains support finding char* in list<string>

mattabullock:
Currently `SContains` will only work with char* if you cast it to a string first, so this adds the shortcut.


flodnv:
Can you please add some tests for this function [here](https://github.com/Expensify/Bedrock/blob/b07d9da2d1a0ffb2a6eda74942554bf42c9972a9/test/tests/LibStuffTest.cpp#L26-L26) ? Just a few standard tests, a few lines

mattabullock:
Updated!

botify:

PR #242, opened 2017-07-11
Only check for uniqueness when creating a single job

fnwbr:
@pecanoro, will you please review this?

Coming from https://github.com/Expensify/Bedrock/pull/195/files#r126695588 and https://github.com/Expensify/Bedrock/pull/195/files#r126695946

### Tests

1. Make sure passing `unique: true` with `CreateJobs` still checks for an already existing job inside the peeking process:
```
CreateJob           
name: www-prod/IntercomCloseConversation?id=123
unique: true

200 OK
commitCount: 3467
peekTime: 7615
totalTime: 7760
Content-Length: 13

{"jobID":595}







CreateJob           
name: www-prod/IntercomCloseConversation?id=123
unique: true

200 OK
commitCount: 3467
peekTime: 2826
totalTime: 2903
Content-Length: 13

{"jobID":595}
```
```
Jul 14 19:45:57 vagrant-ubuntu-trusty-64 bedrock: xxxxx (Jobs.cpp:233) peekCommand [worker0] [info] {Jobs} Unique flag was passed, checking existing job with name www-prod/IntercomCloseConversation?id=123
Jul 14 19:45:57 vagrant-ubuntu-trusty-64 bedrock: xxxxx (libstuff.cpp:2150) SQuery [worker0] [dbug] SELECT jobID, data FROM jobs WHERE name='www-prod/IntercomCloseConversation?id=123';
Jul 14 19:45:57 vagrant-ubuntu-trusty-64 bedrock: xxxxx (Jobs.cpp:244) peekCommand [worker0] [info] {Jobs} Job already existed and unique flag was passed, reusing existing job 595
```

2. Make sure that `CreateJobs` still creates unique jobs just fine, but without checking for existing jobs (aka make sure the `checking existing job with name...` log line does not show up when creating multiple jobs using `CreateJobs`)

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -239,7 +239,7 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
                 }
 
                 // If there's no job or the existing job doesn't match the data we've been passed, escalate to master.
-                if (!result.empty() && result[0][1] == job["data"]) {
+                if (!result.empty() && result[0][1] == (job["data"].empty() ? SToStr("{}") : job["data"])) {
                     // Return early, no need to pass to master, there are no more jobs to create.
                     SINFO("Job already existed and unique flag was passed, reusing existing job " << result[0][0]);
                     content["jobID"] = result[0][0];
@@ -386,7 +386,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                 }
 
                 // If we got a result, and it's data is the same as passed, we won't change anything.
-                if (!result.empty() && result[0][1] == job["data"]) {
+                if (!result.empty() && result[0][1] == (job["data"].empty() ? SToStr("{}") : job["data"])) {
                     SINFO("Job already existed with matching data, and unique flag was passed, reusing existing job "
                           << result[0][0]);
 
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -227,20 +227,20 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
                 }
             }
 
-            // Verify unique
-            SQResult result;
-            SINFO("Unique flag was passed, checking existing job with name " << job["name"]);
-            if (!db.read("SELECT jobID, data "
-                         "FROM jobs "
-                         "WHERE name=" + SQ(job["name"]) + ";",
-                         result)) {
-                throw "502 Select failed";
-            }
+            // Verify unique, but only do so when creating a single job using CreateJob
+            if (SIEquals(request.methodLine, "CreateJob") && SContains(job, "unique") && job["unique"] == "true") {
+                SQResult result;
+                SINFO("Unique flag was passed, checking existing job with name " << job["name"]);
+                if (!db.read("SELECT jobID, data "
+                             "FROM jobs "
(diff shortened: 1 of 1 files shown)


pecanoro:
Wait, before I merge, did you test that this is working fine with `netcat`?

quinthar:
Sorry for being a bit late on this, but to clarify, is this trying to ignore `unique: true` to the `CreateJobs` command?   What would you think about returning an error instead so there is no confusion?

fnwbr:
@pecanoro Didn't test anything; just wanted to get your eyes on this first. Will add `netcat` tests to the PR description. Thanks for the reminder!

@quinthar If I didn't get this wrong this should just fix/disable the `unique`-ness check in `peekCommand` when using `CreateJobs` -

When creating a single job and using `unique: true` we _might_ want to return the job ID of the existing unique job and not escalate the whole command to master.

This does not affect allowing to create multiple jobs with `unique: true` - this still works and should work, right?

quinthar:
Ooh, got it.  Thanks, nm!

pecanoro:
@fnwbr Yes, creating multiple jobs will be still processed by the command, this should do an early return only for `createJob`

fnwbr:
Updated, also tested this.
@pecanoro Please check if this makes sense.

plugins/Jobs.cpp
```diff
@@ -239,7 +239,7 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
                 }
 
                 // If there's no job or the existing job doesn't match the data we've been passed, escalate to master.
-                if (!result.empty() && result[0][1] == job["data"]) {
+                if (!result.empty() && result[0][1] == (job["data"].empty() ? SToStr("{}") : job["data"])) {
```
pecanoro: I am curious, why are you using `SToStr`?

fnwbr: `"{}"` was a different type than `SString`, it was like `char` or whatever and therefore comparison didn't work
No CPP pro here, you tell me if there's a better way to do this

pecanoro: Interesting, I've done comparisons without `SToStr` in the past and it was working fine, so I am not sure why this doesn't work. I am not a C++ pro either, so ¯\_(ツ)_/¯ Btw, besides this one, we also have `to_string` which does the same thing.

I am really curious about the `SToStr ` so before merging I am going to ask Florent or someone who knows better.

mattabullock: ```plugins/Jobs.cpp:241:73: error: operands to ?: have different types ‘const char [3]’ and ‘std::map<std::basic_string<char>, SString, STableComp>::mapped_type {aka SString}’``` 

The actual compile error is a bit more telling. With ternary functions you can't have one side be a char* and the other side be a std::string (or SString or whatever). To fix that (without converting it) you can break it up into two `&&`s (which you might want to do anyway, not in love with a ternary inside an if).

pecanoro: Ok, I figured it out, I read the whole error in the logs haha It's because of the ternary operator, `"{}"` is `char*` while `job["data"]` is `SString`, it seems both have to be the same type. 

Anyways, should we move the ternary operator to a variable? It's a bit hard to read right now. Having a ternary condition in a comparator seems odd.

pecanoro: Disregard my last message, I sent it without seeing @mattabullock's comment

botify:

Issue #111, opened 2017-02-20
Add `retryAfter` to Bedrock::Jobs::GetJob


quinthar:
Problem:
====
Most jobs are automatically processed, such as via `BedrockWorkerManager`.  And automatically retrying a job that is automatically processed is generally a bad idea, as not knowing anything else, whatever failed the first time will likely fail again.  (And if it failed for a reason that will likely not happen again, then the worker should throw a RetryException... but that's a different story.)  

However, manually processed jobs (eg, by loading a webpage that dequeues a job for processing) often fail for reasons that are non-repeatable, such as the person who loaded the webpage to process that job just closing the browser and walking away.  In that scenario, we want to detect an "abandoned" manual job and automatically retry it.

(Note: in this scenario the website should also expire the job, such that if the web user *does* come back and try to finish it, the website should refuse it.  Otherwise we have two web users doing the same job, which ain't great.)

Solution:
====
There are likely several ways to do this, but here is my recommendation:

1. Add a `retryAfter` column to `jobs`, default 0.

2. Update `CreateJob` such that it records the new optional `retryAfter` parameter.

3. Update `GetJob` such that [the place that marks a job as `RUNNING`](https://github.com/Expensify/Bedrock/blob/master/plugins/Jobs.cpp#L363) to check if `retryAfter` is set.  If so:
    * set state=`RUNQUEUED`
        * We use a new state to signal that it is not just running, but queued to re-run in the future as well.
        * This requires updating everywhere that checks for `state="RUNNING"` to be `(state="RUNNING" || state="RUNQUEUED")`
    * set `nextRun=SCURRENT_TIMESTAMP()+retryAfter`, thereby ensuring it'll be re-run at the right time

4. Update `GetJob` (in both [`peekCommand`](https://github.com/Expensify/Bedrock/blob/master/plugins/Jobs.cpp#L84) and [`processCommand`](https://github.com/Expensify/Bedrock/blob/master/plugins/Jobs.cpp#L315)) to replace `state='QUEUED'` with `state IN ('QUEUED', 'RUNQUEUED')`

5. Update [the repeat logic in `FinishJob`](https://github.com/Expensify/Bedrock/blob/master/plugins/Jobs.cpp#L520) to set `lastScheduled = nextRun-retryAfter;`

[...]

PR #240, opened 2017-07-07
Prevent creating grandchildren from RUNNING jobs.

pecanoro:
@cead22 

I am leaving this on HOLD because it may create some conflicts with https://github.com/Expensify/Bedrock/pull/195

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/56724

### Tests

```
createJob
name:job

200 OK
commitCount: 16145
peekTime: 2063
processTime: 2782
totalTime: 15457
Content-Length: 12

{"jobID":98}

query:select * from jobs where jobID>97; 

200 OK
commitCount: 16146
peekTime: 7113
totalTime: 7459
Content-Length: 191

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-07-07 21:37:41 | 98 | QUEUED | job | 2017-07-07 21:37:41 |  |  | {} | 500 | 0 | 

getJob
name:job

200 OK
commitCount: 16146
peekTime: 4880
processTime: 2483
totalTime: 18713
Content-Length: 35

{"data":{},"jobID":98,"name":"job"}

query:select * from jobs where jobID>97; 

200 OK
commitCount: 16147
peekTime: 6326
totalTime: 6418
Content-Length: 211

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-07-07 21:37:41 | 98 | RUNNING | job | 2017-07-07 21:37:41 | 2017-07-07 21:37:58 |  | {} | 500 | 0 | 

createJob
name:job
parentJobID:98

200 OK
commitCount: 16147
peekTime: 3476
processTime: 3457
totalTime: 13410
Content-Length: 12

{"jobID":99}

query:select * from jobs where jobID>97;

200 OK
commitCount: 16148
peekTime: 7641
totalTime: 8382
Content-Length: 298

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-07-07 21:37:41 | 98 | RUNNING | job | 2017-07-07 21:37:41 | 2017-07-07 21:37:58 |  | {} | 500 | 0 | 
2017-07-07 21:38:24 | 99 | PAUSED | job | 2017-07-07 21:38:24 |  |  | {} | 500 | 98 | 

createJob
name:job
parentJobID:99

405 Cannot create grandchildren
commitCount: 16152
peekTime: 2981
processTime: 819
totalTime: 4670
Content-Length: 0
```

PR #191, opened 2017-05-22
Add Bedrock::Job::CancelJob

pecanoro:
@cead22 

I rewrote the command taking [Carlos's comment](https://github.com/Expensify/Bedrock/pull/191#issuecomment-304101067) into account.

## Tests

```
createJob
name:testCancel

200 OK
commitCount: 15970
Content-Length: 12

{"jobID":34}

query:select * from jobs where jobid=34

200 OK
commitCount: 15971
Content-Length: 198

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-06-06 23:32:16 | 34 | QUEUED | testCancel | 2017-06-06 23:32:16 |  |  | {} | 500 | 0 | 

createJob
name:testCancel
parentJobID:34

405 Can only create child job when parent is RUNNING or PAUSED
commitCount: 15971
Content-Length: 0

getJob
name:testCancel

200 OK
commitCount: 15971
Content-Length: 42

{"data":{},"jobID":34,"name":"testCancel"}

query:select * from jobs where jobid=34

200 OK
commitCount: 15972
Content-Length: 218

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-06-06 23:32:16 | 34 | RUNNING | testCancel | 2017-06-06 23:32:16 | 2017-06-06 23:34:14 |  | {} | 500 | 0 | 

createJob
name:testCancel
parentJobID:34

200 OK
commitCount: 15972
Content-Length: 12

{"jobID":35}

createJob
name:testCancel
parentJobID:34

200 OK
commitCount: 15973
Content-Length: 12

{"jobID":36}

createJob
name:testCancel
parentJobID:34

200 OK
commitCount: 15974
Content-Length: 12

{"jobID":37}

query:select * from jobs where jobid IN (35,36,37) 

200 OK
commitCount: 15975
Content-Length: 387

created | jobID | state | name | nextRun | lastRun | repeat | data | priority | parentJobID | retryAfter
2017-06-06 23:34:29 | 35 | PAUSED | testCancel | 2017-06-06 23:34:29 |  |  | {} | 500 | 34 | 
2017-06-06 23:34:32 | 36 | PAUSED | testCancel | 2017-06-06 23:34:32 |  |  | {} | 500 | 34 | 
2017-06-06 23:39:22 | 37 | PAUSED | testCancel | 2017-06-06 23:39:22 |  |  | {} | 500 | 34 | 

cancelJob
[...]

PR #243, opened 2017-07-11
Bring back `retryAfter` functionality

fnwbr:
This is in progress, @mea36 is going to complete the code including automated tests + the full description including test and QA steps

---

Closes https://github.com/Expensify/Bedrock/issues/111
Also related to https://github.com/Expensify/Expensify/issues/53817

Original PR was here: https://github.com/Expensify/Bedrock/pull/169
Which was reverted here: https://github.com/Expensify/Bedrock/pull/194

---
Update from @mea36 
I made the following changes to the original PR:

- I added tests for the retryAfter functionality (https://github.com/Expensify/Bedrock/pull/243/commits/0f549c344941fa20069413e6482b450f73493aa5)
- I updated `GetJob` in process to order by `nextRun` **and** `jobID` because two jobs that have the same `nextRun` time should return the job that was inserted first (fix is in 19983f1ffd2d8781536a18cce04cffafe12525df)
- I updated `GetJob` in peek to look for jobs in states `QUEUED` **and** `RUNQUEUED` (fix is in f3ea4f78fdc5b8e5314480284073e9c154c1d520)


mea36:
I'm going on vacation, so if someone wants to pick this up, feel free to do so.

Update:
I added all the tests from https://github.com/Expensify/Bedrock/pull/195 and https://github.com/Expensify/Bedrock/pull/169
I added these in a quick and dirty way, so I don't think the test names are the best and will change it later

Tests still need to be added from https://github.com/Expensify/Bedrock/pull/191

I had some difficulty after adding the tests for https://github.com/Expensify/Bedrock/pull/169
The test `retryWithChildren` works when run by itself but segfaults when run with all the other tests in the file.  I haven't figured out why

`retryJobComesFirst` fails.  This tests the actual lifecycle of retrying a job, therefore I think the code is buggy.
`retryJobComesFirst` creates 2 retryable jobs and one nonretryable job.  We use `GetJob` to get the first retryable job and then sleep for the retry after value.  After that `GetJob` should (if I understand the functionality correctly - I admittedly haven't looked at the code yet) return the first retryable job.  Instead it gets the nonretryable job.

To run just the CreateJob test execute
```
./test -only CreateJob
```

When `retryWithChildren` fails, it'll keep the process running. That looks like
```
$ ps aux | grep -i bedrock
root     17395  0.0  0.6 260864 13252 ?        Ssl  21:34   0:00 bedrock -fork -db /vagrant/data/auth/main.db -serverHost 0.0.0.0:4444 -nodeName auth-on-bedrock -nodeHost localhost:4445 -keyHost localhost:4001 -priority 200 -v -cache 10001 -debugKeys -plugins db,auth
```
If you don't kill that process then the next time you run the tests they'll fail

fnwbr:
Just to clarify: I'm not working on this; I opened this PR to kickstart this and give someone a base to get started off

By now there are conflicts already again; initially the only thing that would've been to do here was writing tests + testing

cead22:
@fnwbr @mea36 can you guys take ownership of this please since we need it for SmartScan Dos?

mea36:
I assumed this was on hold due to our feature freeze.  If that's the not the case then I can start working on it next week

cead22:
Awesome thank you, and to clarify, given that this is part of SmartScan Dos it should not be on hold.

mea36:
Note to self, I need to add the tests from https://github.com/Expensify/Bedrock/pull/240 as well

iwiznia:
This has no issue attached, thus can't add priority labels. This [should be treated as `Daily` now](https://expensify.slack.com/archives/C6V5B2KTR/p1504022505000063?thread_ts=1504017974.000384&cid=C6V5B2KTR).

fnwbr:
Wrong, the issue is linked above in the initial description. Can't add the `Daily` label though because it's in the Bedrock repo...

mea36:
All tests from #195 and #169 are added.
I didn't like the tests from https://github.com/Expensify/Bedrock/pull/191 so I didn't copy them.  Looking at the comments it looks like @pecanoro created `CancelJob` in one way and then updated the code but didn't update the tests she used to test it. @pecanoro is that correct?
For `CancelJob` (https://github.com/Expensify/Bedrock/pull/191) I instead created tests based on the code.

I also added a test for https://github.com/Expensify/Bedrock/pull/240

There were some bug fixes I had to make
- `CancelJob` was broken for when the jobID did not exist (fix is a83fad52b96e71810bfe1f1176d66a38266cbb39)
- I updated `GetJob` in process to order by `nextRun` **and** `jobID` because two jobs that have the same `nextRun` time should return the job that was inserted first (fix is in 19983f1ffd2d8781536a18cce04cffafe12525df)
- I updated `GetJob` in peek to look for jobs in states `QUEUED` **and** `RUNQUEUED` (fix is in f3ea4f78fdc5b8e5314480284073e9c154c1d520)

Note, all I did was add tests.  I didn't touch the functionality of `retryAfter` if it wasn't to fix a test.  I have some questions about the implementation and will comment on the appropriate lines to start a conversation

mea36:
@flodnv ready for review

mea36:
I'll resolve the conflicts tomorrow.  I'm also going to remove the tests that are unrelated to `retryAfter` so they can be reviewed independently of this functionality

mea36:
Holding on https://github.com/Expensify/Bedrock/pull/275

pecanoro:
> @pecanoro created CancelJob in one way and then updated the code but didn't update the tests she used to test it. @pecanoro is that correct?

Those test should be working, I kept updating them, @cead22 used them to test it before merging the code. What was the problem?

iwiznia:
SSDos first version is almost ready and we'll need this. @mea36 is it close to be done? Do you have an ETA?

mea36:
ETA - EOD today

mea36:
We had too many different versions of the test files to extracted the tests that @cead22 requested I make into https://github.com/Expensify/Bedrock/pull/275 (which was merged and this branch is up-to-date with master)

This is now ready to be reviewed.
- I added tests for the retryAfter functionality (https://github.com/Expensify/Bedrock/pull/243/commits/0f549c344941fa20069413e6482b450f73493aa5)
- I updated `GetJob` in process to order by `nextRun` **and** `jobID` because two jobs that have the same `nextRun` time should return the job that was inserted first (fix is in 19983f1ffd2d8781536a18cce04cffafe12525df)
- I updated `GetJob` in peek to look for jobs in states `QUEUED` **and** `RUNQUEUED` (fix is in f3ea4f78fdc5b8e5314480284073e9c154c1d520)

I'll update the description to reflect this comment

iwiznia:
@pecanoro, @tylerkaraszewski can you review the PR?

iwiznia:
Also, I'm not finding what made us revert this change last time... are we sure it will not happen again?

plugins/Jobs.cpp
```diff
@@ -647,9 +698,35 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-            throw "502 Update failed";
+
+        // Update jobs without retryAfter
+        if (!nonRetriableJobs.empty()) {
+            string jobIDs = SComposeList(nonRetriableJobs);
+            SINFO("Updating jobs without retryAfter " << jobIDs);
+            string updateQuery = "UPDATE jobs "
+                                 "SET state='RUNNING', "
+                                 "lastRun=" + SCURRENT_TIMESTAMP() + " "
+                                 "WHERE jobID IN (" + jobIDs + ");";
```
pecanoro: `SQList(nonRetriableJobs)`

plugins/Jobs.cpp
```diff
@@ -647,9 +698,35 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-        if (!db.write(updateQuery)) {
-            throw "502 Update failed";
+
+        // Update jobs without retryAfter
+        if (!nonRetriableJobs.empty()) {
+            string jobIDs = SComposeList(nonRetriableJobs);
+            SINFO("Updating jobs without retryAfter " << jobIDs);
+            string updateQuery = "UPDATE jobs "
+                                 "SET state='RUNNING', "
+                                 "lastRun=" + SCURRENT_TIMESTAMP() + " "
```
pecanoro: You are missing the indentation here, right?

plugins/Jobs.cpp
```diff
@@ -647,9 +698,35 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+        if (!retriableJobs.empty()) {
+            SINFO("Updating jobs with retryAfter");
+            for (auto job : retriableJobs) {
+                string jobID = job["jobID"];
+                string retryAfter = job["retryAfter"];
+                string updateQuery = "UPDATE jobs "
+                                     "SET state='RUNQUEUED', "
+                                     "  lastRun=" + SCURRENT_TIMESTAMP() + ", "
+                                     "  nextRun=DATETIME(" + SCURRENT_TIMESTAMP() + ", " + SQ(retryAfter) + ") "
+                                     "WHERE jobID = " + jobID + ";";
```
pecanoro: `SQ(jobID)`?

plugins/Jobs.cpp
```diff
@@ -647,9 +698,35 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+
+        // Update jobs with retryAfter
+        if (!retriableJobs.empty()) {
+            SINFO("Updating jobs with retryAfter");
+            for (auto job : retriableJobs) {
+                string jobID = job["jobID"];
+                string retryAfter = job["retryAfter"];
+                string updateQuery = "UPDATE jobs "
+                                     "SET state='RUNQUEUED', "
+                                     "  lastRun=" + SCURRENT_TIMESTAMP() + ", "
```
pecanoro: Those spaces `"  last` are weird, you should indent the line instead

plugins/Jobs.cpp
```diff
@@ -704,6 +781,11 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
 
         const string& nextRun = result[0][1];
         const string& lastRun = result[0][2];
+        const string& state = result[0][3];
+
+        if (state == "RUNQUEUED") {
+            throw "402 Auto-retrying jobs cannot be updated once running";
```
pecanoro: I am curious, why can't it be updated?

plugins/Jobs.cpp
```diff
@@ -1026,7 +1109,11 @@ string BedrockPlugin_Jobs::_constructNextRunDATETIME(const string& lastScheduled
     }
 
     // Validate the sqlite date modifiers
-    // See: https://www.sqlite.org/lang_datefunc.html
+    if (!_isValidSQLiteDateModifier(SComposeList(parts))){
+        SWARN("Syntax error, failed parsing repeat");
+        return "";
+    }
+
     for (const string& part : parts) {
```
pecanoro: If you created `_isValidSQLiteDateModifier`, this seems unneeded and doing the same thing.

plugins/Jobs.cpp
iwiznia: +1, we can update regular running jobs, why wouldn't we be able to do so for ones with retryAfter?

plugins/Jobs.cpp
```diff
@@ -406,6 +432,12 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                 if (!result.empty()) {
                     updateJobID = SToInt64(result[0][0]);
                 }
+
+                // Alert if job is unique and retryAfter is set
+                // This shouldn't happen since we validate this in peekCommand
+                if (SContains(job, "retryAfter") && job["retryAfter"] != "") {
```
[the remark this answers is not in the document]
mea36: This was here from the original code version of this code (remember that we cherry-picked the commits from the original attempt into this branch).

But I agree with you, it's not necessary.  @cead22 is there a specific reason you had this here?

plugins/Jobs.cpp
```diff
@@ -494,8 +532,11 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                     }
                 }
 
+                // If no data was provided, use an empty object
+                const string& safeRetryAfter = job["retryAfter"].empty() ? "\"\"" : SQ(job["retryAfter"]);
```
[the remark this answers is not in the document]
mea36: Actually, `job` is a `STable` which is a `map` which correctly handles a key that doesn't exist http://www.cplusplus.com/reference/map/map/operator[]/

I will change the `"\"\""` to `SQ("")`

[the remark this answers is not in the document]
pecanoro: > Actually, job is a STable which is a map which correctly handles a key that doesn't exist

Ok, perfect then, can you make sure though? So we don't get any weird exceptions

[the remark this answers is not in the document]
pecanoro: I mean, I am more concerned about calling `.empty()` on whatever it returns when the key doesn't exist, which I am not sure if it's a string

plugins/Jobs.cpp
```diff
@@ -601,13 +643,22 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+                nonRetriableJobs.push_back(result[c][0]);
+            } else {
+                STable job;
+                job["jobID"] = result[c][0];
+                job["retryAfter"] = result[c][4];
+                retriableJobs.push_back(job);
+            }
 
             // See if this job has any FINISHED/CANCELLED child jobs, indicating it is being resumed
             SQResult childJobs;
```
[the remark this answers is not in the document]
mea36: Good catch

plugins/Jobs.cpp
mea36: I have no idea.   @cead22 was this part of the original requirement?

mea36:
I made a few updates.  I think I should make some automated tests for `GetJob` to make sure we're not breaking any functionality. I'll get to that tomorrow (unless someone wants to do that for me :P )

test/tests/jobs/CreateJobTest.cpp
```diff
@@ -234,4 +240,167 @@ struct CreateJobTest : tpunit::TestFixture {
...
+        ASSERT_EQUAL(response["name"], jobName);
+
+        // Query the db and confirm that state, nextRun and lastRun are correct
+        SQResult jobData;
+        tester->readDB("SELECT state, nextRun, lastRun FROM jobs WHERE jobID = " + jobID + ";", jobData);
+
+        ASSERT_EQUAL(jobData[0][0], "RUNQUEUED");
+        string nextRun = jobData[0][1];
+        string lastRun = jobData[0][2];
+        struct tm tm1;
```
cead22: Can you please explain this in a comment, specifically this `ASSERT_EQUAL(difftime(nextRunTime, lastRunTime), 1);`?

test/tests/jobs/CreateJobTest.cpp
```diff
@@ -234,4 +240,167 @@ struct CreateJobTest : tpunit::TestFixture {
...
+        struct tm tm2;
+        strptime(nextRun.c_str(), "%Y-%m-%d %H:%M:%S", &tm1);
+        time_t nextRunTime = mktime(&tm1);
+        strptime(lastRun.c_str(), "%Y-%m-%d %H:%M:%S", &tm2);
+        time_t lastRunTime = mktime(&tm2);
+        ASSERT_EQUAL(difftime(nextRunTime, lastRunTime), 1);
+
+        // Get the job, confirm error
+        tester->executeWaitVerifyContent(command, "404 No job found");
+        // Wait 5 seconds, get the job, confirm no error
```
cead22: - Blank line before comment
- Comments says 5 seconds but sleep is called with 1

test/tests/jobs/FinishJobTest.cpp
```diff
@@ -353,4 +354,26 @@ struct FinishJobTest : tpunit::TestFixture {
...
+        command["name"] = "job";
+        command["retryAfter"] = "+1 SECOND";
+        STable response = tester->executeWaitVerifyContentTable(command);
+        string jobID = response["jobID"];
+
+        // Get the job
+        command.clear();
+        command.methodLine = "GetJob";
+        command["name"] = "job";
+        tester->executeWaitVerifyContent(command);
```
cead22: In previous tests you're asserting that you're getting the right info in the response, why not here?

test/tests/jobs/FinishJobTest.cpp
```diff
@@ -353,4 +354,26 @@ struct FinishJobTest : tpunit::TestFixture {
...
+
+        // Finish it
+        command.clear();
+        command.methodLine = "FinishJob";
+        command["jobID"] = jobID;
+        tester->executeWaitVerifyContent(command);
```
cead22: No assertions?

test/tests/jobs/RetryJobTest.cpp
```diff
@@ -278,4 +279,26 @@ struct RetryJobTest : tpunit::TestFixture {
...
+        command.clear();
+        command.methodLine = "GetJob";
+        command["name"] = "job";
+        tester->executeWaitVerifyContent(command);
+
+        // Finish it
+        command.clear();
+        command.methodLine = "RetryJob";
+        command["jobID"] = jobID;
+        tester->executeWaitVerifyContent(command);
```
cead22: No assertions?

plugins/Jobs.cpp
pecanoro: Yeah, but why is that? Why is it not safe for a RUNQUEUED job but it is for a RUNNING one? Maybe @quinthar knows the answer for this better

test/tests/jobs/FinishJobTest.cpp
mea36: This is a test for `FinishJob` functionality.  A test for `GetJob` should verify the return values of `GetJob`

plugins/Jobs.cpp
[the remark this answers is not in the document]
mea36: I'm going to take this out.  If this is happening then it's a symptom of a much larger problem (ie `peek` and `process` aren't being run sequentially on master)

test/tests/jobs/FinishJobTest.cpp
mea36: Added some

plugins/Jobs.cpp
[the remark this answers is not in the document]
mea36: good catch

plugins/Jobs.cpp
```diff
@@ -550,35 +592,35 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-                    "SELECT jobID, name, data, priority, parentJobID "
+                    "SELECT jobID, name, data, priority, parentJobID, retryAfter "
                     "FROM jobs "
-                    "WHERE state='QUEUED' "
+                    "WHERE state IN ('QUEUED', 'RUNQUEUED') "
                     "  AND priority=1000"
                     "  AND " + SCURRENT_TIMESTAMP() + ">=nextRun "
                     "  AND name GLOB " + SQ(name) + " "
-                    "ORDER BY nextRun ASC LIMIT " + safeNumResults +
+                    "ORDER BY nextRun, jobID ASC LIMIT " + safeNumResults +
```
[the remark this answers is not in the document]
mea36: 1.  I think what you're asking here is if `ORDER BY priority DESC` does a stable sort.  Is that correct?  If so, I created tests for `GetJob` that test the return order of jobs with the same `priority` and `nextRun` but different  `state` values.  Therefore I'm confident that `order by nextRun, jobID` isn't affected by `ORDER BY priority DESC`.  Check out the `testPrioritiesWithRunQueued` test
1.  I haven't benchmarked this but I did check that the query plan is the same.  If you want, I can benchmark it tomorrow
Query plan without `jobID ASC`
```
~$ sudo readdb.sh "explain query plan SELECT jobID, name, data, priority, parentJobID, retryAfter
> FROM jobs
> WHERE state IN ('QUEUED', 'RUNQUEUED')
>   AND priority=0
>   AND current_timestamp >=nextRun
>   AND name GLOB 'www-prod/*'
> ORDER BY nextRun ASC LIMIT 10;"
----------  ----------  ----------  -------------------------------------------------------------------------------------------------
0           0           0           SEARCH TABLE jobs USING INDEX jobsStatePriorityNextRunName (state=? AND priority=? AND nextRun<?)
0           0           0           EXECUTE LIST SUBQUERY 1
0           0           0           USE TEMP B-TREE FOR ORDER BY
```
Query plan with `jobID ASC`
```
~$ sudo readdb.sh "explain query plan SELECT jobID, name, data, priority, parentJobID, retryAfter
> FROM jobs
> WHERE state IN ('QUEUED', 'RUNQUEUED')
>   AND priority=0
>   AND current_timestamp >=nextRun
>   AND name GLOB 'www-prod/*'
> ORDER BY nextRun ASC, jobID ASC LIMIT 10;"
----------  ----------  ----------  -------------------------------------------------------------------------------------------------
0           0           0           SEARCH TABLE jobs USING INDEX jobsStatePriorityNextRunName (state=? AND priority=? AND nextRun<?)
0           0           0           EXECUTE LIST SUBQUERY 1
0           0           0           USE TEMP B-TREE FOR ORDER BY
```

plugins/Jobs.cpp
mea36: Per https://github.com/Expensify/Bedrock/issues/111#issuecomment-327648440, I'll remove this and create a test for it tomorrow

mea36: Actually, with this removed there is no change to `UpdateJob` so I won't create a test

plugins/Jobs.cpp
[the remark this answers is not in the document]
cead22: > created tests for GetJob that test the return order of jobs with the same priority and nextRun but different state values. Therefore I'm confident that order by nextRun, jobID isn't affected by ORDER BY priority DESC

The test should be with the same priority and nextRun but different priorities not different states, right?

What I'm thinking with this is that we _may_ be slowing down one of the main queries to solve what seems to me like a non-isssue which is the fact that Job B which was created after Job A may be processed before Job A if they both have the same priority and the very same nextRun. The reason I call it a non-issue is because I understand the case but it definitely seems like an edge case and I haven't seen that this is causing a problem, am I missing something? Would you agree that this change might not be worth implementing now?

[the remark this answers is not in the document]
mea36: If we generally consider this a non-issue, then I can revert this change.  The reason I implemented it was because this test failed
https://github.com/Expensify/Bedrock/pull/243/files#diff-c8ebb7cec8b361355acfa8a63051dacbR364

Now the following jobs also tests will also fail 
https://github.com/Expensify/Bedrock/pull/243/files#diff-c8ebb7cec8b361355acfa8a63051dacbR268
https://github.com/Expensify/Bedrock/pull/243/files#diff-cf845d40b06e01293a549aa13ac565adR326

So if this is a non-issue, I'll update the tests accordingly

iwiznia:
@mea36 tests are failing... 
Also, @tylerkaraszewski you are listed as reviewer, so can you review this or remove yourself as a reviewer?

pecanoro:
@mea36 forgot to remove one of the tests after the removing https://github.com/Expensify/Bedrock/pull/243/commits/cafa9fcd6cdd78847eccde605846bd0f491c1243

plugins/Jobs.cpp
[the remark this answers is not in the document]
mea36: Actually, @quinthar do you have an opinion here? 
I agree with @cead22 that this is an edge case but it's also the proper way to achieve the behavior we want.
If we move forward with this change, I should devise a plan to test this query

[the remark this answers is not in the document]
quinthar: I'm not sure I follow.  Are you asking if it's worth processing same-priority jobs in order, or if it's ok if same-priority jobs execute out of order?  I'd say it would be great to process them in order if possible, and it sounds like @mea36's test confirms that it is not just possible, but has no effect on the query plan (and thus should have the same performance).  Am I misunderstanding something?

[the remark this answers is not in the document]
cead22: No that's exactly right and for added context:
- I agree that Mel's implementation is correct
- I don't think that 2 equal query plans translates into two equal query execution times. My worry here specifically is that the temp btree used to sort could be bigger since it's now using 2 values and thus take longer to create and traverse (which we do 3 times)
- In light of that and our recent query perf degradation problems, I'd love for us to bechmark the new query vs the old one before releasing to be on the safer side
- And because of that I proposed that as an alternative we don't solve this edge case now given that this doesn't solve any problems from the user's perspective, and live with the fact that 2 jobs with the exact same nextRun and the exact same priority maybe be dispatched our of order.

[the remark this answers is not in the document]
mea36: Since the hard work was already done to do some performance testing (thanks @cead22 ), I followed the steps in https://github.com/Expensify/Expensify/issues/49378#issuecomment-298134063 using a jobs table from @cead22 that has 347K jobs;

I created a new database and created a table using the schema I have on dev (minus the `retryAfter` column
```sql
sqlite> .schema jobs
CREATE TABLE jobs ( created TIMESTAMP NOT NULL, jobID INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, state TEXT NOT NULL, name TEXT NOT NULL, nextRun TIMESTAMP NOT NULL, lastRun TIMESTAMP, repeat TEXT NOT NULL, data TEXT NOT NULL, priority INTEGER NOT NULL DEFAULT 500, parentJobID INTEGER NOT NULL DEFAULT 0);
CREATE INDEX jobsState    ON jobs ( state    );
CREATE INDEX jobsName     ON jobs ( name     );
CREATE INDEX jobsNextRun  ON jobs ( nextRun  );
CREATE INDEX jobsLastRun  ON jobs ( lastRun  );
CREATE INDEX jobsPriority ON jobs ( priority );
CREATE INDEX jobsParentJobIDState ON jobs ( parentJobID, state );
CREATE INDEX jobsStatePriorityNextRunName ON jobs ( state, priority, nextRun, name );
```

Then copied the subquery to a script that ran the query 1000 times

I ran this twice for each version of the query.
`ORDER BY nextRun ASC` had an average runtime of `0.003772` and `0.003973`
`ORDER BY nextRun ASC, jobID ASC` had an average runtime of `0.004018` and `0.003936`

I think the difference is inconsequential and I think we should move forward with these changes

plugins/Jobs.cpp
```diff
@@ -442,7 +465,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
             int64_t parentJobID = SContains(job, "parentJobID") ? SToInt(job["parentJobID"]) : 0;
             if (parentJobID) {
                 SQResult result;
-                if (!db.read("SELECT state, parentJobID FROM jobs WHERE jobID=" + SQ(parentJobID) + ";", result)) {
+                if (!db.read("SELECT state, parentJobID, retryAfter FROM jobs WHERE jobID=" + SQ(parentJobID) + ";", result)) {
```
mea36: I'm not sure why this is here, removing it.

plugins/Jobs.cpp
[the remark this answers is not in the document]
cead22: 👍

plugins/Jobs.cpp
mea36: This is here because auto-retrying parents shouldn't have children - https://github.com/Expensify/Bedrock/pull/169/files#diff-d116ebb4bb9e2a3fd799045c7b182707R372

Adding this back.  I think it disappeared in a bad merge with master when we added `parentJobID` to this query

mea36: Never mind, I'm confusing myself.  We check this in `peek` and the code I linked to is alerting, not throwing.  Removing this

mea36:
Woot! Tests are passing and I think I addressed everyone's comments

test/tests/jobs/GetJobTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+        command["query"] = "DELETE FROM jobs WHERE jobID > 0;";
+        tester->executeWaitVerifyContent(command);
+    }
+
+    void tearDownClass() { delete tester; }
+
+    // Simple GetJob
+    void getJob() {
+        // Create the job
+        SData command("CreateJob");
```
cead22: You could DRY this file with a func to create a job that takes name, priority, nextRun, etc. Nab

test/tests/jobs/GetJobTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+        // Medium
+        command["name"] = "medium_4";
+        command["priority"] = "500";
+        response = tester->executeWaitVerifyContentTable(command);
+        jobList.push_back(response["jobID"]);
+
+        // Confirm these jobs all have the same nextRun time
+        SQResult result;
+        tester->readDB("SELECT nextRun FROM jobs WHERE jobID IN (" + SComposeList(jobList) + ") GROUP BY nextRun;", result);
```
cead22: SELECT DISTINTC is clearer imo, nab

test/tests/jobs/GetJobTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+        command["priority"] = "0";
+        command["firstRun"] = SComposeTime("%Y-%m-%d %H:%M:%S", STimeNow());
+        STable response = tester->executeWaitVerifyContentTable(command);
+        list<string> jobList;
+        jobList.push_back(response["jobID"]);
+
+        // Medium
+        command["name"] = "medium";
+        command["priority"] = "500";
+        command["firstRun"] = SComposeTime("%Y-%m-%d %H:%M:%S", STimeNow()+1000000);
```
cead22: You should use spaces around `+`

test/tests/jobs/GetJobTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+        ASSERT_EQUAL(response["name"], "low");
+        sleep(1);
+        response = tester->executeWaitVerifyContentTable(command);
+        ASSERT_EQUAL(response["name"], "medium");
+        sleep(1);
+        response = tester->executeWaitVerifyContentTable(command);
+        ASSERT_EQUAL(response["name"], "high");
+
+        // Sleep 3 seconds so that the nextRun time for the last 'high' job is now
+        sleep(3);
```
cead22: Either remove this and change the assertions below or change the comment in the description ;)
`// Expect the jobs to be returned in order of low, medium, high, low, medium, high`

test/tests/jobs/GetJobTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+        // Checking the cancelled job
+        list<string> cancelledChildJobs = SParseJSONArray(response["cancelledChildJobs"]);
+        ASSERT_EQUAL(cancelledChildJobs.size(), 1);
+        childJob = SParseJSONObject(cancelledChildJobs.front());
+        ASSERT_EQUAL(childJob["jobID"], cancelledChildID);
+        ASSERT_EQUAL(childJob["data"], cancelledChildData);
+    }
+
+    // This is the same as testPriorities but some of the states are set to RUNQUEUED
+    // So we set the the firstRun for all the jobs that don't have a retryAfter to 1 second in the future
```
cead22: `the the`

test/tests/jobs/GetJobTest.cpp
```diff
@@ -0,0 +1,408 @@
...
+        // GetJob and confirm that the last 3 jobs are returned in priority order since now is past nextRun for all of them
+        response = tester->executeWaitVerifyContentTable(command);
+        ASSERT_EQUAL(response["name"], "high");
+        response = tester->executeWaitVerifyContentTable(command);
+        ASSERT_EQUAL(response["name"], "medium");
+        response = tester->executeWaitVerifyContentTable(command);
+        ASSERT_EQUAL(response["name"], "low");
+    }
+
+    // get a parent job that has finished and cancelled jobs
```
cead22: `Get`

mea36:
Updated

iwiznia:
This seems to be ready and all comments from reviews where addressed. Asking @pecanoro to merge so I can finish SSDos. If there's anything missing, we can address it on a different PR

fnwbr:
Apart from the included tests there are no further QA steps necessary?

mea36:
I think the tests should cover it

plugins/Jobs.cpp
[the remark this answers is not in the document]
quinthar: Looks like our performance testing wasn't adequate here -- when harvesting started, 1.5M jobs became runnable, and performance slowed to a near-halt.  What's interesting is we had more or less this exact fear, and actually thought we tested to confirm it, but still missed it.  What can we do differently next time to catch this?  Only do performance testing on prod, and only right after harvesting starts?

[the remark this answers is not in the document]
quinthar: Cc: @coleaeason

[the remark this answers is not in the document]
coleaeason: Well one way to catch this next time would be to make sure we time queries while harvesting is running and while it's not running. 

Another way to catch this would be if we could build a test that mimics harvesting, i.e. it queues 1.5mn run ready jobs that have similar parameters to harvesting jobs (retry times, data, etc.). It would be nice if we had a way to fail a test if a query or operation takes more time than X.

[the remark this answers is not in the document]
iwiznia: > Well one way to catch this next time would be to make sure we time queries while harvesting is running and while it's not running.

I agree, running queries in the jobs table when the queue is almost empty is almost as running them on dev. 
Not sure how big is the jobs table, but maybe we can grab a snapshot of it when it has data and save it to use it in any performance test we need.

PR #249, opened 2017-07-25
Update SQLite to support indexes using DATETIME and BEGIN CONCURRENT

pecanoro:
@cead22 @tylerkaraszewski 

New SQLite from 2017-07-20 supporting `DATETIME` on indexes

cc/ @flodnv

### DIFF
--- a/libstuff/sqlite3.h
+++ b/libstuff/sqlite3.h
@@ -1,5 +1,5 @@
 /*
-** 2001 September 15
+** 2001-09-15
 **
 ** The author disclaims copyright to this source code.  In place of
 ** a legal notice, here is a blessing:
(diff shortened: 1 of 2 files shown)


pecanoro:
@tylerkaraszewski Travis is failing and I don't know why 😞

pecanoro:
@quinthar I have downloaded the SQLite that Richard linked on the email and contains the `DATETIME` fix and the `BEGIN CONCURRENT` enhancement. We are not sure if Travis is failing due to the `BEGIN CONCURRENT` enhancement, but per your email reply, I assumed that we are already using it.

quinthar:
Can you restate your question?  I'm not sure what you are asking me.

pecanoro:
@quinthar Sure. The diff for this version of SQLite seems quite big and even though I thought it's because of the `BEGIN CONCURRENT` enhancement, after reading your reply on the "Use of DATETIME on indexes" email, I assume that we are already using `BEGIN CONCURRENT` and Travis failing is due to something else and not this enhancement. So I just wanted to confirm if `BEGIN CONCURRENT` code for SQLite is already being used or it's something I am introducing with this PR.

quinthar:
Yes, begin concurrent is being used already

pecanoro:
Wow finally, tests passed!! @flodnv thanks a lot for pointing out the [docs](https://github.com/Expensify/Web-Expensify/blob/master/_devportal/docs/sqlite.md), the problem was indeed not using the flag `OPT_FEATURE_FLAGS = -DSQLITE_ENABLE_UPDATE_DELETE_LIMIT=1`!!

I also tested it on my dev and the index was created successfully.

iwiznia:
This should not be on hold anymore, right @pecanoro?

flodnv:
@iwiznia we are holding on getting a stable Expensify Bedrock release out there before merging this in.

iwiznia:
Makes total sense. I hope next time we deploy we don't have to revert... 😁

flodnv:
Let's get this out now 😺

PR #270, opened 2017-08-25
Add functionality to RetryJob to change the job name and set a nextRun

mea36:
Fixes https://github.com/Expensify/Expensify/issues/59735

This makes the following changes to `RetryJob`
- Adds the functionality to change the name of the job (and thus change the `queue` the job is in
- Adds the functionality to explicitly set the `nextRun` time (currently you can only set the job to try again after a delay of some amount of seconds in the future)
- For backwards compatibility, it doesn't remove  `delay`, but uses `nextRun` if both `delay` and `nextRun` are given


plugins/Jobs.cpp
```diff
@@ -721,19 +721,22 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
     }
 
     // ----------------------------------------------------------------------
-    else if (SIEquals(requestVerb, "RetryJob") || SIEquals(requestVerb, "FinishJob")) {
-        // - RetryJob( jobID, delay, [data] )
+    else if (SIEquals(requestVerb, "RequeueJob") || SIEquals(requestVerb, "RetryJob") || SIEquals(requestVerb, "FinishJob")) {
+        // - RetryJob is deprecated, use RequeueJob
```
cead22: Beyond semantics, is there a benefit to renaming this command? This will eventually turn into a breaking change and not that we have millions of people using bedrock but something to keep in mind now that this is open source and given that we don't have versioning. This is not a blocker

mea36: There is no reason beyond semantics

mea36:
On hold for https://github.com/Expensify/Bedrock/pull/275/files

plugins/Jobs.cpp
mea36: There's a lot less overhead to stick with `RetryJob`, I'm going to go that route

mea36:
Off hold

mea36:
Off hold

iwiznia:
Not sure if I should merge or we are waiting for @cead22's review.

mea36:
Let's wait for @cead22 to review

plugins/Jobs.md
```diff
@@ -24,7 +24,14 @@ Bedrock::Jobs is a plugin to the [Bedrock data foundation](../README.md) that ma
    * *data* - (optional) New data object to associate with the job (especially useful if repeating, to pass state to the next worker).
 
  * **DeleteJob( jobID )** - Removes all trace of a job.
-   * *jobID* - Identifier of the job to delete 
+   * *jobID* - Identifier of the job to delete
+
+ * **Retry( jobID )** - Removes all trace of a job.
```
cead22: 🎯

plugins/Jobs.md
```diff
@@ -24,7 +24,14 @@ Bedrock::Jobs is a plugin to the [Bedrock data foundation](../README.md) that ma
...
+   * *jobID* - Identifier of the job to retry
+   * *delay* - (optional) Number of seconds to wait before retrying
+   * *nextRun* - (optional) The time/date on which the job should be set to run again, in "YYYY-MM-DD [HH:MM:SS]" format
```
cead22: I'd add "if the job has a `repeat` value set, then this param is ignored.

plugins/Jobs.md
```diff
@@ -24,7 +24,14 @@ Bedrock::Jobs is a plugin to the [Bedrock data foundation](../README.md) that ma
    * *data* - (optional) New data object to associate with the job (especially useful if repeating, to pass state to the next worker).
 
  * **DeleteJob( jobID )** - Removes all trace of a job.
-   * *jobID* - Identifier of the job to delete 
+   * *jobID* - Identifier of the job to delete
+
+ * **Retry( jobID )** - Removes all trace of a job.
+   * *jobID* - Identifier of the job to retry
+   * *delay* - (optional) Number of seconds to wait before retrying
```
cead22: I'd add "if the job has a `repeat` value set or `nextRun` is sent, then this param is ignored.

mea36:
A bunch of these tests were added in https://github.com/Expensify/Bedrock/pull/275
> Test FinishJob

See https://github.com/Expensify/Bedrock/blob/48e6c3cea1baecc9e1525e1af8f8e314e6ab9762/test/tests/jobs/FinishJobTest.cpp

> Test RetryJob on a job with repeat, sending nextRun and delay and making sure that the new nextRun is calculated using the lastRun + delay
Test RetryJob on a job without repeat, sending nextRun and delay and making sure that the new nextRun is calculated using the passed nextRun

I thought I created tests for this, apparently I didn't or it got lost in a merge
Nonetheless, I disagree.  I think if `nextRun` is set then we should always use that value. Why should it matter if the job is set to repeat or not?

> Test RetryJob on a job without repeat, sending only delay and making sure that the new nextRun is calculated using the lastRun + delay

https://github.com/Expensify/Bedrock/blob/993103d8eae73e82e39020ca44353bf0844f2b8c/test/tests/jobs/RetryJobTest.cpp#L214

> RetryJob with delay < 0

https://github.com/Expensify/Bedrock/blob/993103d8eae73e82e39020ca44353bf0844f2b8c/test/tests/jobs/RetryJobTest.cpp#L192

I'll create a for 
> RetryJob with malformed nextRun

And update the PR description

cead22:
> Nonetheless, I disagree. I think if nextRun is set then we should always use that value. Why should it matter if the job is set to repeat or not?

Oh I was going off of the code, [this line](https://github.com/Expensify/Bedrock/pull/270/files#diff-d116ebb4bb9e2a3fd799045c7b182707R831) specifically, maybe I'm misinterpreting it but it seems like delay takes precedence over nextRun

plugins/Jobs.cpp
```diff
@@ -815,30 +818,45 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+        // If this is set to repeat, get the nextRun value
         if (!repeat.empty()) {
+            safeNewNextRun = _constructNextRunDATETIME(nextRun, lastRun, repeat);
+        } else if (SIEquals(requestVerb, "RetryJob")) {
+            const string& newNextRun = request["nextRun"];
+
+            if (newNextRun.empty()) {
+                SINFO("nextRun isn't set, using delay");
+                int64_t delay = request.calc64("delay");
+                if (delay < 0) {
```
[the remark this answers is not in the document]
mea36: A check for `RetryJob` doesn't exist in `peek`
Same for `FinishJob` for that matter.

So I agree with you that this check should be in `peek`, but do want to do that refactor now?  I'm cool either way

mea36:
Good catch, I'll update the logic so `nextRun` has precedence over `delay`

mea36:
Sorry, I was confused. The original code gives precedence to `repeat` and I think we should continue doing that.  I'm not going to update that logic, but I will update the comment and create tests for this

mea36:
Updated.
I added tests for 
- RetryJob with repeat, sending nextRun and delay and making sure that the new nextRun is calculated using the **repeat**
- RetryJob without repeat, sending nextRun and delay and making sure that the new nextRun is the `nextRun` given
- FinishJob with nextRun ignores nextRun
- FinishJob with delay ignores delay

Outstanding questions:
- I didn't create a test for a malformed `nextRun` because we don't test for that in `CreateJob` with `firstRun`.  Though I think we should, if you agree, I'll implement that
- Neither`RetryJob` nor `FinishJob` make use of `peekCommand` when validating the job is eligible for the respective command.  Should I refactor them to utilize `peekCommand` or keep things the way they are?

quinthar:
> NeitherRetryJob nor FinishJob make use of peekCommand when validating the job is eligible for the respective command. Should I refactor them to utilize peekCommand or keep things the way they are?

I think we should only add a `peekCommand` if we feel it has a reasonable chance of handling a significant volume of commands on the slave.

cead22:
> I didn't create a test for a malformed nextRun because we don't test for that in CreateJob with firstRun. Though I think we should, if you agree, I'll implement that

+1 to implementing them but let's not block on it. Please create an issue, preferably a first pick



> I think we should only add a peekCommand if we feel it has a reasonable chance of handling a significant volume of commands on the slave.

Agreed. Given the amount of throws in ::process for retryjob/finishjob it seems to me that we should implement ::peek for these as well

quinthar:
>> I think we should only add a peekCommand if we feel it has a reasonable chance of handling a significant volume of commands on the slave.
>
> Agreed. Given the amount of throws in ::process for retryjob/finishjob it seems to me that we should implement ::peek for these as well

Oh, hah, I drew the opposite conclusion.  It feels like 99.99% of the time, if you call `RetryJob` or `FinishJob`, it will in fact be escalated to the master and processed there.   Accordingly, I don't think in practice a `peekCommand` makes sense here because in practice it will nearly never do anything useful.

cead22:
I was erring on the safe side but I'll defer to you guys.

mea36:
I'm going to leave it in the `RetryJob` and `FinishJob` in `process`.  
I made an issue to create a function to validate `nextRun` - https://github.com/Expensify/Expensify/issues/61325

mea36:
Updated and merged with master

test/tests/jobs/FinishJobTest.cpp
```diff
@@ -14,12 +15,17 @@ struct FinishJobTest : tpunit::TestFixture {
                               TEST(FinishJobTest::deleteFinishedJobWithNoChildren),
                               TEST(FinishJobTest::hasRepeat),
                               TEST(FinishJobTest::inRunqueuedState),
-                              AFTER(FinishJobTest::tearDown),
+                              TEST(FinishJobTest::hasRepeatWithDelay),
+                              TEST(FinishJobTest::hasDelay),
+                              TEST(FinishJobTest::hasRepeatWithNextRun),
+                              */
+                              TEST(FinishJobTest::hasNextRun),
+                              //AFTER(FinishJobTest::tearDown),
```
cead22: Why is this all commented out?

test/tests/jobs/FinishJobTest.cpp
```diff
@@ -377,4 +383,127 @@ struct FinishJobTest : tpunit::TestFixture {
...
+        // Finish it
+        command.clear();
+        command.methodLine = "FinishJob";
+        command["jobID"] = jobID;
+        command["nextRun"] = "2017-09-07 23:11:11";
+        tester->executeWaitVerifyContent(command);
+
+        // Confirm nextRun is in 1 hour, not in the given nextRun time
+        SQResult result;
+        tester->readDB("SELECT created, nextRun FROM jobs WHERE jobID = " + jobID + ";", result);
```
cead22: I just noticed this and I realize it doesn't matter for the test, but I think these queries should be `SELECT lastRun, nextRun` instead of `SELECT created, nextRun`, right?

test/tests/jobs/RetryJobTest.cpp
```diff
@@ -302,4 +307,167 @@ struct RetryJobTest : tpunit::TestFixture {
         tester->readDB("SELECT state FROM jobs WHERE jobID = " + jobID + ";",  result);
         ASSERT_EQUAL(result[0][0], "QUEUED");
     }
+
+    // Confirm nextrun is updated appropriately
```
cead22: nextRun

test/tests/jobs/RetryJobTest.cpp
```diff
@@ -302,4 +307,167 @@ struct RetryJobTest : tpunit::TestFixture {
...
+        command.clear();
+        command.methodLine = "GetJob";
+        command["name"] = "job";
+        tester->executeWaitVerifyContent(command);
+
+        // Retry it
+        command.clear();
+        command.methodLine = "RetryJob";
+        command["jobID"] = jobID;
+        command["delay"] = "";
```
cead22: I think you meant `"5";` here

test/tests/jobs/FinishJobTest.cpp
mea36: I did when testing and forgot to revert it :woman_facepalming:

plugins/Jobs.cpp
```diff
@@ -629,6 +629,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         SASSERT(!SIEquals(requestVerb, "GetJob") || result.size()<=1);
 
         // Prepare to update the rows, while also creating all the child objects
+        string updateQuery = "UPDATE jobs SET state='RUNNING', lastRun=" + SCURRENT_TIMESTAMP() + " WHERE jobID IN (";
```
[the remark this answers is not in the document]
mea36: Good catch.  The merge with master had a bunch of conflicts, I must of missed this

plugins/Jobs.cpp
```diff
@@ -879,30 +889,46 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
-                throw "402 Malformed repeat";
+                SINFO("nextRun isn't set, using delay");
+                int64_t delay = request.calc64("delay");
+                if (delay < 0) {
+                    throw "402 Must specify a non-negative delay when retrying";
+                }
+                repeat = "FINISHED, +" + SToStr(delay) + " SECONDS";
+                safeNewNextRun = _constructNextRunDATETIME(nextRun, lastRun, repeat);
+                if (safeNewNextRun.empty()) {
+                    throw "402 Malformed repeat";
```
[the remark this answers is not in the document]
mea36: Updating this to be `402 Malformed delay` and added a test for this

test/tests/jobs/FinishJobTest.cpp
mea36: you're right.  I'm going to change it

mea36:
Updated, addressed the comments, and the conflicts

fnwbr:
We're not going to actually rename the command name, right?
Can you update the title of the PR?

fnwbr:
Conflicts

mea36:
Updated with master and fixed conflicts

PR #281, opened 2017-09-12
Add new restrictions to CancelJob

pecanoro:
@mea36 

Like we discussed in the original issue, I added two new restrictions:

- CancelJob can only be called on a job with a parent
- CancelJob can only be called on child job that has at least one sibling in the RUNNING state

### Issues

$ https://github.com/Expensify/Expensify/issues/60822

### Tests

Automated


plugins/Jobs.cpp
```diff
@@ -319,6 +324,21 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
             return true; // Done
         }
 
+        // We can't control where CancelJob is called from, but we can verify that there's at least one RUNNING sibling around
+        if (!db.read("SELECT j.jobID "
```
mea36: We know the parentJobID from the query above.
We can simplify this to 
```sql
SELECT jobID
FROM jobs
WHERE parentID = result[0][2] and state = 'RUNNING'
LIMIT 1;
```

test/tests/jobs/CancelJobTest.cpp
```diff
@@ -84,18 +85,38 @@ struct CancelJobTest : tpunit::TestFixture {
         tester->executeWaitVerifyContent(command, "404 Invalid jobID - Cannot cancel a job with children");
     }
 
-    // Ignore canceljob for RUNNING jobs
+    // Ignore canceljob of RUNNING jobs
```
mea36: Can you change this back to `for`?

test/tests/jobs/CancelJobTest.cpp
```diff
@@ -206,28 +227,86 @@ struct CancelJobTest : tpunit::TestFixture {
...
-        // Cancel it
+        // Create a sibling
+        command.clear();
+        command.methodLine = "CreateJob";
+        command["name"] = "sibling";
+        command["parentJobID"] = parentID;
+        response = tester->executeWaitVerifyContentTable(command);
+        string siblingID = response["jobID"];
+
+        // Finish the parent to put the child in the QUEUED state
```
mea36: `to put the children in `

test/tests/jobs/CancelJobTest.cpp
```diff
@@ -206,28 +227,86 @@ struct CancelJobTest : tpunit::TestFixture {
...
+        command.methodLine = "FinishJob";
+        command["jobID"] = parentID;
+        tester->executeWaitVerifyContent(command);
+
+        // Try cancelling one of the children, but none of them are running so it should fail.
         command.clear();
         command.methodLine = "CancelJob";
-        command["jobID"] = jobID;
+        command["jobID"] = childID;
+        tester->executeWaitVerifyContent(command, "404 Invalid jobID - Cannot cancel a job that has no RUNNING siblings");
```
mea36: You don't need to test this here, you test this in `cancelJobWithoutSiblings`

test/tests/jobs/CancelJobTest.cpp
pecanoro: No idea, probably I wanted to write something else and wrote that when I undid it

test/tests/jobs/CancelJobTest.cpp
pecanoro: Copy/paste lol

pecanoro:
Updated!

pecanoro:
Updated!

PR #285, opened 2017-09-15
Improve performance for GetJob query

mea36:
cc @cead22 

Removing `jobID ASC` from the `ORDER BY` clause increases performance
by almost 500%

I tested on a table with ~1700000 rows.
The average performance of the query with `jobID ASC` and `LIMIT 1` in the inner query `ORDER BY` clauses was `2.33696`
Without `jobID ASC` the average performance was `0.00497`

Fixes
$ https://github.com/Expensify/Expensify/issues/61957

### DIFF
--- a/test/tests/jobs/CreateJobTest.cpp
+++ b/test/tests/jobs/CreateJobTest.cpp
@@ -19,7 +19,6 @@ struct CreateJobTest : tpunit::TestFixture {
                               TEST(CreateJobTest::retryUnique),
                               TEST(CreateJobTest::retryLifecycle),
                               TEST(CreateJobTest::retryWithChildren),
-                              TEST(CreateJobTest::retryJobComesFirst),
                               AFTER(CreateJobTest::tearDown),
                               AFTER_CLASS(CreateJobTest::tearDownClass)) { }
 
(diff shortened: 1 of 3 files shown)


test/tests/jobs/CreateJobTest.cpp
```diff
@@ -349,42 +348,4 @@ struct CreateJobTest : tpunit::TestFixture {
         command["parentJobID"] = jobID;
         tester->executeWaitVerifyContent(command, "402 Auto-retrying parents cannot have children");
     }
-
-    // Retryable job (after retry period has passed) comes before QUEUED job
-    void retryJobComesFirst() {
```
mea36: I removed this because this is basically testing `GetJob` and I think `testPrioritiesWithRunQueued` adequately covers this test case

mea36:
[This comment](https://github.com/Expensify/Expensify/issues/61957#issuecomment-329922471) states that we decided not to use the [query you linked to](https://github.com/Expensify/Expensify/issues/61957#issuecomment-329887815) because we want to minimize the things that we're changing.  While the second query performs better, it's only slightly better, not an order of magnitude better.
Therefore, we'll stick with the same query we had before https://github.com/Expensify/Expensify/issues/61957#issuecomment-329875574

iwiznia:
Oh, alright, sorry about that.

PR #289, opened 2017-09-22
Add PRAGMA synchronous flag

bondydaa:
@coleaeason will you review this, por favor?

This is to add a `-synchronous` flag that allows us to set `PRAGMA synchronous` to `true`, the default is set to `false` since that is currently what this setting is.

fixes: https://github.com/Expensify/Expensify/issues/62424

## Tests
1. Edited `/etc/init/bedrock.conf` and added `-synchronous EXTRA` at the end, stopped bedrock, started bedrock, got the following output in the logs: screen shot 2017-09-27 at 2 37 51 pm
1. Edited `/etc/init/bedrock.conf` and had no `-synchronous` flag at the end, stopped bedrock, started bedrock, got the following output in the logs: screen shot 2017-09-27 at 2 36 15 pm
1. Edited `/etc/init/bedrock.conf` and added `-synchronous XXXXXX` at the end, stopped bedrock, started bedrock, got the following output in the logs: screen shot 2017-09-27 at 2 38 50 pm

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -70,11 +70,8 @@ void BedrockServer::sync(SData& args,
     // If still no value, use the number of cores on the machine, if available.
     workerThreads = workerThreads ? workerThreads : max(1u, thread::hardware_concurrency());
 
-    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
-    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
-
     // Initialize the DB.
-    SQLite db(args["-db"], args.calc("-cacheSize"), 1024, args.calc("-maxJournalSize"), -1, workerThreads - 1, synchronous);
+    SQLite db(args["-db"], args.calc("-cacheSize"), 1024, args.calc("-maxJournalSize"), -1, workerThreads - 1, args["-synchronous"]);
 
     // And the command processor.
     BedrockCore core(db, server);
@@ -494,12 +491,9 @@ void BedrockServer::worker(SData& args,
 {
     SInitialize("worker" + to_string(threadId));
 
-    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
-    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
-
     // We pass `0` as the checkpoint size to disable checkpointing from workers. This can be a slow operation, and we
     // don't want workers to be able to block the sync thread while it happens.
-    SQLite db(args["-db"], args.calc("-cacheSize"), 0, args.calc("-maxJournalSize"), threadId, threadCount - 1, synchronous);
+    SQLite db(args["-db"], args.calc("-cacheSize"), 0, args.calc("-maxJournalSize"), threadId, threadCount - 1, args["-synchronous"]);
     BedrockCore core(db, server);
 
     // Command to work on. This default command is replaced when we find work to do.
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -70,8 +70,11 @@ void BedrockServer::sync(SData& args,
     // If still no value, use the number of cores on the machine, if available.
     workerThreads = workerThreads ? workerThreads : max(1u, thread::hardware_concurrency());
 
+    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
+    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
+
     // Initialize the DB.
-    SQLite db(args["-db"], args.calc("-cacheSize"), 1024, args.calc("-maxJournalSize"), -1, workerThreads - 1);
+    SQLite db(args["-db"], args.calc("-cacheSize"), 1024, args.calc("-maxJournalSize"), -1, workerThreads - 1, synchronous);
 
     // And the command processor.
     BedrockCore core(db, server);
@@ -491,9 +494,12 @@ void BedrockServer::worker(SData& args,
 {
     SInitialize("worker" + to_string(threadId));
 
+    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
+    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
+
     // We pass `0` as the checkpoint size to disable checkpointing from workers. This can be a slow operation, and we
     // don't want workers to be able to block the sync thread while it happens.
-    SQLite db(args["-db"], args.calc("-cacheSize"), 0, args.calc("-maxJournalSize"), threadId, threadCount - 1);
+    SQLite db(args["-db"], args.calc("-cacheSize"), 0, args.calc("-maxJournalSize"), threadId, threadCount - 1, synchronous);
     BedrockCore core(db, server);
 
     // Command to work on. This default command is replaced when we find work to do.
--- a/main.cpp
+++ b/main.cpp
@@ -274,6 +274,7 @@ int main(int argc, char* argv[]) {
     SETDEFAULT("-maxJournalSize", "1000000");
     SETDEFAULT("-queryLog", "queryLog.csv");
     SETDEFAULT("-enableMultiWrite", "true");
+    SETDEFAULT("-synchronous", "false");
 
     args["-plugins"] = SComposeList(loadPlugins(args));
 
--- a/main.cpp
+++ b/main.cpp
@@ -274,7 +274,6 @@ int main(int argc, char* argv[]) {
     SETDEFAULT("-maxJournalSize", "1000000");
     SETDEFAULT("-queryLog", "queryLog.csv");
     SETDEFAULT("-enableMultiWrite", "true");
-    SETDEFAULT("-synchronous", "false");
 
     args["-plugins"] = SComposeList(loadPlugins(args));
 
--- a/main.cpp
+++ b/main.cpp
@@ -215,6 +215,9 @@ int main(int argc, char* argv[]) {
              << endl;
         cout << "-maxJournalSize <#commits>  Number of commits to retain in the historical journal (default 1000000)"
              << endl;
+        cout << "-synchronous    <value>     Set the PRAGMA schema.synchronous "
+                "(defaults see https://sqlite.org/pragma.html#pragma_synchronous)"
+             << endl;
         cout << endl;
         cout << "Quick Start Tips:" << endl;
         cout << "-----------------" << endl;
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -74,8 +74,10 @@ SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int ma
     SASSERT(!SQuery(_db, "new file format for DESC indexes", "PRAGMA legacy_file_format = OFF"));
 
     // Compare to see if "false" wasn't passed meaning the -synchronous flag was set and run the query
-    if (!SIEquals(synchronous, "false")) {
-         SASSERT(!SQuery(_db, "setting synchronous commits", "PRAGMA synchronous = " + synchronous  + ";"));
+    if (!synchronous.empty()) {
+        SASSERT(!SQuery(_db, "setting custom synchronous commits", "PRAGMA synchronous = " + synchronous  + ";"));
+    } else {
+        DBINFO("Using SQLite default PRAGMA synchronous");
     }
 
     // These other pragmas only relate to read/write databases.
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -73,9 +73,9 @@ SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int ma
     // any tables to be effective.
     SASSERT(!SQuery(_db, "new file format for DESC indexes", "PRAGMA legacy_file_format = OFF"));
 
-    // Compare to see if "false" wasn't passed meaning the -synchronous flag was set and run the query
+    // Check if synchronous has been set and run query to use a custom synchronous setting
     if (!synchronous.empty()) {
-        SASSERT(!SQuery(_db, "setting custom synchronous commits", "PRAGMA synchronous = " + synchronous  + ";"));
+        SASSERT(!SQuery(_db, "setting custom synchronous commits", "PRAGMA synchronous = " + SQ(synchronous)  + ";"));
     } else {
         DBINFO("Using SQLite default PRAGMA synchronous");
     }
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -15,7 +15,7 @@ atomic_flag                         SQLite::_sqliteInitialized = ATOMIC_FLAG_INI
 SLockTimer<recursive_mutex> SQLite::g_commitLock("Commit Lock", SQLite::_commitLock);
 
 SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int maxJournalSize, int journalTable,
-               int maxRequiredJournalTableID, string synchronous) :
+               int maxRequiredJournalTableID, const string& synchronous) :
     whitelist(nullptr)
 {
     // Initialize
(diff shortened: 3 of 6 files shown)


bondydaa:
I'm getting an error, hence the `WIP` tag on this right now.

I believe it doesn't like me trying to concat this string here: https://github.com/Expensify/Bedrock/pull/289/files#diff-e0fd399d80d036c72780c20678a97b54R77

Here are the `make` logs:
```
vagrant@vagrant-ubuntu-trusty-64:/vagrant/Bedrock$ make
g++-6 -g -DSVERSION="\"a5101c338e606f9786db7272097a83ea5499b1a7\"" -Wall -O0 -std=gnu++14 -I/vagrant/Bedrock -I/vagrant/Bedrock/mbedtls/include -Werror -Wno-unused-result -MMD -MF .build/sqlitecluster/SQLite.d -include libstuff/libstuff.h -o .build/sqlitecluster/SQLite.o -c sqlitecluster/SQLite.cpp
sqlitecluster/SQLite.cpp: In constructor ‘SQLite::SQLite(const string&, int, int, int, int, int, std::string)’:
sqlitecluster/SQLite.cpp:77:81: error: invalid operands of types ‘const char [9]’ and ‘const char [21]’ to binary ‘operator+’
     string synchronousLog = (synchronous == "false") ? "disabling" : "enabling" + " synchronous commits";
                                                                      ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
In file included from <command-line>:0:0:
sqlitecluster/SQLite.cpp:79:58: error: no matching function for call to ‘SQuery(sqlite3*&, std::string&, std::string&)’
     SASSERT(!SQuery(_db, synchronousLog, synchronousQuery));
                                                          ^
./libstuff/libstuff.h:53:15: note: in definition of macro ‘SASSERT’
         if (!(_LHS_)) {                                                                                                \
               ^~~~~
In file included from <command-line>:0:0:
./libstuff/libstuff.h:757:5: note: candidate: int SQuery(sqlite3*, const char*, const string&, SQResult&, int64_t)
 int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result,
     ^~~~~~
./libstuff/libstuff.h:757:5: note:   candidate expects 5 arguments, 3 provided
./libstuff/libstuff.h:759:12: note: candidate: int SQuery(sqlite3*, const char*, const string&, int64_t)
 inline int SQuery(sqlite3* db, const char* e, const string& sql, int64_t warnThreshold = 1000 * STIME_US_PER_MS) {
            ^~~~~~
./libstuff/libstuff.h:759:12: note:   no known conversion for argument 2 from ‘std::string {aka std::basic_string<char>}’ to ‘const char*’
make: *** No rule to make target `.build/sqlitecluster/SQLite.d', needed by `.build/sqlitecluster/SQLite.o'.  Stop.
```

I think it has something to do with this but :shrug: https://stackoverflow.com/questions/15319859/how-to-concatenate-two-strings-in-c

Any help would be much appreciated.

sqlitecluster/SQLite.cpp
```diff
@@ -15,7 +15,7 @@ atomic_flag                         SQLite::_sqliteInitialized = ATOMIC_FLAG_INI
 SLockTimer<recursive_mutex> SQLite::g_commitLock("Commit Lock", SQLite::_commitLock);
 
 SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int maxJournalSize, int journalTable,
-               int maxRequiredJournalTableID) :
+               int maxRequiredJournalTableID, string synchronous) :
```
quinthar: please use `const string&` as it used by `filename`; this avoids an extraneous copy.  (Candidly, not important here, but best to establish the habit for consistency with the rest of the code.)

sqlitecluster/SQLite.cpp
```diff
@@ -74,7 +74,9 @@ SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int ma
     SASSERT(!SQuery(_db, "new file format for DESC indexes", "PRAGMA legacy_file_format = OFF"));
 
     // These other pragmas only relate to read/write databases.
-    SASSERT(!SQuery(_db, "disabling synchronous commits", "PRAGMA synchronous = OFF;"));
+    string synchronousLog = (synchronous == "false" ? "disabling" : "enabling") + " synchronous commits";
```
quinthar: This is trying to just add two `const char*` pointers, so probably won't work.  I'd suggest casting the result of the first parenthesis to `(string)` to convert the pointer to an STL string.  Then the expression is adding a `string` to a `const char*` which should work as you like it.

HOWEVER, I think a confusion here is in thinking `synchronous` is true/false.  Here are the actual string options: https://sqlite.org/pragma.html#pragma_synchronous  So rather than assuming it's enable/disable, just log what the new setting is without worrying about interpreting the result into an "enable/disable" message.

(Furthermore, for future reference -- when doing string comparisons, please use `SIEquals()` for a case-insensitive compare.  However, I still think you shouldn't compare it to anything and should instead just log the value requested.)

sqlitecluster/SQLite.h
```diff
@@ -23,7 +23,7 @@ class SQLite {
     // maxRequiredJournalTableID: This is the maximum journal table ID that we'll verify. If it's -1, we'll only verify
     //                            'journal' and no numbered tables.
     SQLite(const string& filename, int cacheSize, int autoCheckpoint, int maxJournalSize, int journalTable,
-           int maxRequiredJournalTableID);
+           int maxRequiredJournalTableID, string synchronous = "false");
```
quinthar: Rather than having a default one set here, I'd suggest just allowing this to remain unset -- and we only call the `PRAGMA` statement if it's not set.  So, we just use SQLite's default if unspecified on the command line, and only override that default if specified.

sqlitecluster/SQLite.cpp
bondydaa: 👍 that is similar to what Cole suggested 1:1. 

> except then you'll have issues down stream.
> `SQuery` takes `db, char, string`
> i would skip on making the log line say what we're setting it to
> because we're going to log that when we run the SQuery
> so you can just leave the `synchronousLog` as `setting setting database synchronicity` or something

Thanks for the tips though!

bondydaa:
Okay I pushed up the some changes that hopefully get me a little bit closer. I think one test is still failing and I'll copy over the docs for the new flag into `docs/` once what I have is approved.

Please let me know what else I'm missing or have messed up 😄

bondydaa:
Huzzahs, just pushed up some other changes and now Travis is passing!

I'm questioning whether I should have passed `false` within the test files since I believe in most cases we want to test a different type of `synchronous` PRAGMA so just lemme know if those should be different.

quinthar:
"false" isn't a valid option. Can you take a look at the sqlite docs?  I'd suggest just not calling the pragma if the setting isn't present on the command line.

bondydaa:
So I've set the `-synchronous` flag to `false` as the default for the command line this way the query to set PRAGMA synchronous is not called and the default PRAGMA is used (unless I've misunderstood the docs). 

I check if `!SIEquals(synchronous, "false")` and only call the query if it's something other than `false`.

```
// Compare to see if "false" wasn't passed meaning the -synchronous flag was set and run the query
if (!SIEquals(synchronous, "false")) {
    SASSERT(!SQuery(_db, "setting synchronous commits", "PRAGMA synchronous = " + synchronous  + ";"));
}
```

My understanding from the docs is that if no PRAMGA synchronous is set then it will be `synchronous=FULL` unless `SQLITE_EXTRA_DURABLE` option is set which changes it to `synchronous=EXTRA`

> The default setting is usually synchronous=FULL. The SQLITE_EXTRA_DURABLE compile-time option changes the default to synchronous=EXTRA.

Or are the docs saying that you still need to manually call this query and set it to `FULL` or `EXTRA` depending on that option?

sqlitecluster/SQLite.h
```diff
@@ -23,7 +23,7 @@ class SQLite {
     // maxRequiredJournalTableID: This is the maximum journal table ID that we'll verify. If it's -1, we'll only verify
     //                            'journal' and no numbered tables.
     SQLite(const string& filename, int cacheSize, int autoCheckpoint, int maxJournalSize, int journalTable,
-           int maxRequiredJournalTableID);
+           int maxRequiredJournalTableID, const string& synchronous);
```
tylerkaraszewski: Give this a default value of `""`, then you don't need to explicitly pass it everywhere (like in the tests).

sqlitecluster/SQLite.cpp
```diff
@@ -73,8 +73,12 @@ SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int ma
     // any tables to be effective.
     SASSERT(!SQuery(_db, "new file format for DESC indexes", "PRAGMA legacy_file_format = OFF"));
 
+    // Compare to see if "false" wasn't passed meaning the -synchronous flag was set and run the query
+    if (!SIEquals(synchronous, "false")) {
```
tylerkaraszewski: Just do `if (!synchronous.empty())` and if the caller's specified a string, use that, otherwise, do nothing.

main.cpp
```diff
@@ -271,6 +274,7 @@ int main(int argc, char* argv[]) {
     SETDEFAULT("-maxJournalSize", "1000000");
     SETDEFAULT("-queryLog", "queryLog.csv");
     SETDEFAULT("-enableMultiWrite", "true");
+    SETDEFAULT("-synchronous", "false");
```
tylerkaraszewski: We don't need this, let it be blank.

BedrockServer.cpp
```diff
@@ -491,9 +494,12 @@ void BedrockServer::worker(SData& args,
 {
     SInitialize("worker" + to_string(threadId));
 
+    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
+    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
+
     // We pass `0` as the checkpoint size to disable checkpointing from workers. This can be a slow operation, and we
     // don't want workers to be able to block the sync thread while it happens.
-    SQLite db(args["-db"], args.calc("-cacheSize"), 0, args.calc("-maxJournalSize"), threadId, threadCount - 1);
+    SQLite db(args["-db"], args.calc("-cacheSize"), 0, args.calc("-maxJournalSize"), threadId, threadCount - 1, synchronous);
```
tylerkaraszewski: Just pass `args["-synchronous"]` and let it be blank if not set (same as above).

BedrockServer.cpp
```diff
@@ -491,9 +494,12 @@ void BedrockServer::worker(SData& args,
 {
     SInitialize("worker" + to_string(threadId));
 
+    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
+    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
```
tylerkaraszewski: We don't need this either.

BedrockServer.cpp
```diff
@@ -70,8 +70,11 @@ void BedrockServer::sync(SData& args,
     // If still no value, use the number of cores on the machine, if available.
     workerThreads = workerThreads ? workerThreads : max(1u, thread::hardware_concurrency());
 
+    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
+    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
+
     // Initialize the DB.
-    SQLite db(args["-db"], args.calc("-cacheSize"), 1024, args.calc("-maxJournalSize"), -1, workerThreads - 1);
+    SQLite db(args["-db"], args.calc("-cacheSize"), 1024, args.calc("-maxJournalSize"), -1, workerThreads - 1, synchronous);
```
tylerkaraszewski: Just pass `args["-synchronous"]` and let it be blank if not set.

BedrockServer.cpp
```diff
@@ -70,8 +70,11 @@ void BedrockServer::sync(SData& args,
     // If still no value, use the number of cores on the machine, if available.
     workerThreads = workerThreads ? workerThreads : max(1u, thread::hardware_concurrency());
 
+    // Check if the synchronous flag was passed, if not set to false and let the SQLite default get used
+    string synchronous = args.isSet("-synchronous") ? args["-synchronous"] : "false";
```
tylerkaraszewski: We don't need this.

test/lib/BedrockTester.cpp
```diff
@@ -319,7 +319,7 @@ vector<SData> BedrockTester::executeWaitMultipleData(vector<SData> requests, int
 SQLite& BedrockTester::getSQLiteDB()
 {
     if (!_db) {
-        _db = new SQLite(_dbName, 1000000, 0, 3000000, -1, 0);
+        _db = new SQLite(_dbName, 1000000, 0, 3000000, -1, 0, "false");
```
tylerkaraszewski: You don't need these either, when the default is `""`.

test/tests/SQLiteNodeTest.cpp
```diff
@@ -31,7 +31,7 @@ struct SQLiteNodeTest : tpunit::TestFixture {
     void testFindSyncPeer() {
 
         // This exposes just enough to test the peer selection logic.
-        SQLite db(":memory:", 1000000, 100, 5000, -1, -1);
+        SQLite db(":memory:", 1000000, 100, 5000, -1, -1, "false");
```
tylerkaraszewski: You don't need these either, when the default is `""` (same as above).

sqlitecluster/SQLite.cpp
```diff
@@ -73,8 +73,12 @@ SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int ma
...
+         SASSERT(!SQuery(_db, "setting synchronous commits", "PRAGMA synchronous = " + synchronous  + ";"));
```
tylerkaraszewski: I feel like we might want to log that we've done this.

bondydaa:
Alrighty just pushed up fixes from Tyler.

coleaeason:
@bondydaa any tests?

sqlitecluster/SQLite.cpp
```diff
@@ -73,8 +73,14 @@ SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int ma
     // any tables to be effective.
     SASSERT(!SQuery(_db, "new file format for DESC indexes", "PRAGMA legacy_file_format = OFF"));
 
+    // Compare to see if "false" wasn't passed meaning the -synchronous flag was set and run the query
```
tylerkaraszewski: Comment seems wrong.

sqlitecluster/SQLite.cpp
```diff
@@ -73,8 +73,14 @@ SQLite::SQLite(const string& filename, int cacheSize, int autoCheckpoint, int ma
...
+    if (!synchronous.empty()) {
+        SASSERT(!SQuery(_db, "setting custom synchronous commits", "PRAGMA synchronous = " + synchronous  + ";"));
```
tylerkaraszewski: Should probably wrap `synchronous` in `SQ()`. Also, this doesn't log anything when it works.

sqlitecluster/SQLite.h
```diff
@@ -23,7 +23,7 @@ class SQLite {
     // maxRequiredJournalTableID: This is the maximum journal table ID that we'll verify. If it's -1, we'll only verify
     //                            'journal' and no numbered tables.
     SQLite(const string& filename, int cacheSize, int autoCheckpoint, int maxJournalSize, int journalTable,
-           int maxRequiredJournalTableID);
+           int maxRequiredJournalTableID, const string& synchronous="");
```
tylerkaraszewski: Standard style is spaces either side of `=`.

bondydaa:
Alrighty pushed again and addressed Tyler's comments, chatted with Cole and figured out how I could test this and added tests to my comment.

coleaeason:
I'm kind of surprised sqlite doesn't puke when you give it a bad value. Does it use DEFAULT?

bondydaa:
Looks like like if you give it a bad value it sets it to `1`/`NORMAL` which seems weird since that isn't a default.

image

bondydaa:
bumping, looks like tyler and cole are happy. what say ye @quinthar ?

bondydaa:
Doh just realized I had left the `WIP` flag in the title, just removed that.

PR #293, opened 2017-09-28
Allows to cancel jobs without running siblings

iwiznia:
This allows to cancel child jobs regardless of the state of its siblings and re-queues the parent when it's the last child queued.

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -275,7 +275,7 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
     else if (SIEquals(request.methodLine, "CancelJob")) {
         // - CancelJob(jobID)
         //
-        //     Cancel a QUEUED, RUNQUEUED, FAILED job from a sibling.
+        //     Cancel a QUEUED, RUNQUEUED, FAILED child job.
         //
         //     Parameters:
         //     - jobID  - ID of the job to cancel
@@ -322,21 +322,6 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
         if (result[0][0] == "PAUSED") {
             SALERT("Trying to cancel a job " << request["jobID"] << " that is PAUSED");
             return true; // Done
-        }
-
-        // We can't control where CancelJob is called from, but we can verify that there's at least one RUNNING sibling around
-        if (!db.read("SELECT jobID "
-                     "FROM jobs "
-                     "WHERE parentJobID = " + SQ(result[0][2]) + " "
-                        "AND jobID != " + SQ(jobID) + " "
-                        "AND state IN ('RUNNING', 'RUNQUEUED') "
-                     "LIMIT 1;",
-                    result)) {
-            STHROW("502 Select failed");
-        }
-
-        if (result.empty()) {
-            STHROW("404 Invalid jobID - Cannot cancel a job that has no RUNNING/RUNQUEUED siblings");
         }
 
         return false; // Need to process command
@@ -988,7 +973,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
     else if (SIEquals(request.methodLine, "CancelJob")) {
         // - CancelJob (jobID)
         //
-        //     Cancel a QUEUED, RUNQUEUED, FAILED job from a sibling.
+        //     Cancel a QUEUED, RUNQUEUED, FAILED child job.
         //
         //     Parameters:
         //     - jobID  - ID of the job to cancel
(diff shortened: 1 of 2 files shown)


iwiznia:
Damn, broke another test, looking...

iwiznia:
Damn, not that sure about the change after seeing the fails, want to confirm first before modifying all tests...
So, this is the scenario:
- One parent.
- 2 children.
- Parent runs and finishes, becoming PAUSED.
- We get the child one job.
- We cancel another child job. Now, the parent is QUEUED again.
- We try to finish the job that we got 2 steps above, but we can't, because the parent is QUEUED, not PAUSED.

I think this was what @pecanoro was saying, my expectation was to handle that in php (ie: when trying to finish a child job, we need to make sure to allow for it to fail, because another child or even itself could've been canceled).
This is from my investigation of `GetJobTest::testWithFinishedAndCancelledChildren`. `FinishJobTest::removeFinishedAndCancelledChildren` and `RetryJobTest::removeFinishedAndCancelledChildren` also fail and I assume they are the same kind of failure (given their names...). 
So the question is, should I change the behavior of these tests? If I do, will something already relying on it will break?

pecanoro:
You encountered the issue you mention because you forgot to check if there's any child running. If there's any, you shouldn't resume the parent. The child will do that

plugins/Jobs.cpp
```diff
@@ -1000,6 +985,29 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+                     "FROM jobs "
+                     "WHERE jobID=" + SQ(jobID) + ";",
+                     result)) {
+            STHROW("502 Select failed");
+        }
+        const string& safeParentJobID = SQ(result[0][0]);
+        if (!db.read("SELECT count(1) "
+                     "FROM jobs "
+                     "WHERE parentJobID=" + safeParentJobID + " AND "
+                       "state IN ('QUEUED', 'RUNQUEUED');",
```
pecanoro: You basically need to add RUNNING here too, otherwise, like you mention, you will resume the parent with RUNNING children and you don't want that, just let the RUNNING child resume the parent.

iwiznia: 🤦‍♂️

plugins/Jobs.cpp
```diff
@@ -1000,6 +985,29 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
 
+        // If this was the last queued child, resume the parent
+        SQResult result;
+        if (!db.read("SELECT parentJobID "
+                     "FROM jobs "
+                     "WHERE jobID=" + SQ(jobID) + ";",
+                     result)) {
+            STHROW("502 Select failed");
+        }
+        const string& safeParentJobID = SQ(result[0][0]);
```
cead22: Shouldn't you check that safeParentJobID isn't 0?

plugins/Jobs.cpp
```diff
@@ -1000,6 +985,29 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
+        if (!db.read("SELECT count(1) "
+                     "FROM jobs "
+                     "WHERE parentJobID=" + safeParentJobID + " AND "
+                       "state IN ('QUEUED', 'RUNQUEUED', 'RUNNING');",
+                     result)) {
+            STHROW("502 Select failed");
+        }
+        if (SToInt(result[0][0]) == 0) {
+            SINFO("Cancelled last QUEUED child, resuming the parent");
```
cead22: Maybe log jobID?

plugins/Jobs.cpp
pecanoro: If the parent is 0, we will never reach this point, we cannot cancel jobs without a parent, in the peek part of the command

plugins/Jobs.cpp
iwiznia: Added

test/tests/jobs/CancelJobTest.cpp
```diff
@@ -313,24 +313,47 @@ struct CancelJobTest : tpunit::TestFixture {
...
         command["jobID"] = parentID;
         tester->executeWaitVerifyContent(command);
 
-        // Try cancelling one of the children, but none of them are running so it should fail.
+        // Cancel one child
         command.clear();
         command.methodLine = "CancelJob";
         command["jobID"] = childID;
-        tester->executeWaitVerifyContent(command, "404 Invalid jobID - Cannot cancel a job that has no RUNNING/RUNQUEUED siblings");
+        tester->executeWaitVerifyContent(command);
```
mea36: Can you check the state of the child jobs and confirm that they are `CANCELLED`?

mea36:
@cead22 good point

PR #309, opened 2017-10-16
Add secure map, libstuff additions for new plugin

coleaeason:
@tylerkaraszewski please review. A handful of changes here. 

1. Plugins can now send control commands to BedrockServer and have it skip the command queue and get processed. Otherwise it would sit in the queue until it could be processed, which is never in the case of sending a detach then attach.

2. Plugins can now load a secure data map via file or blocking socket. 

3. A few new libstuff additions to generate random strings, new encryption/decryption versions.

4. Fixed an issue in main.cpp where STDOUT had been rerouted to devnull then we tried to cout some errors. Our logging functions are available at this point in main so I just used those.

5. Allows server to start in "bootstrap" mode, this mode starts the BedrockServer node detached so you can modify the db before creating the BedrockServer object.


BedrockServer.cpp
```diff
@@ -40,9 +40,20 @@ void BedrockServer::acceptCommand(SQLiteCommand&& command, bool isNew) {
             command.writeConsistency = SQLiteNode::QUORUM;
             SINFO("Forcing QUORUM consistency for command " << command.request.methodLine);
         }
-        SINFO("Queued new '" << command.request.methodLine << "' command from bedrock node, with " << _commandQueue.size()
-              << " commands already queued.");
-        _commandQueue.push(BedrockCommand(move(command)));
+
+        // If this is a status or a contorl command we need to handle it here, or else
```
tylerkaraszewski: contorl

BedrockServer.cpp
```diff
@@ -40,9 +40,20 @@ void BedrockServer::acceptCommand(SQLiteCommand&& command, bool isNew) {
...
+        // it will get stuck in the _commandQueue if we are in `Detached` mode and the
+        // command comes from a bedrock node.
+        SData request;
+        request.deserialize(command.request.content);
+        request.methodLine = command.request.methodLine;
+        BedrockCommand newCommand(request);
+        if (!_handleIfStatusOrControlCommand(newCommand)) {
+            SINFO("Queued new '" << command.request.methodLine << "' command from bedrock node, with " << _commandQueue.size()
+                    << " commands already queued.");
+            _commandQueue.push(BedrockCommand(move(command)));
```
tylerkaraszewski: We've already constructed a new command. How about:
```
_commandQueue.push(move(newCommand));
```

BedrockServer.cpp
```diff
@@ -1004,21 +1015,56 @@ BedrockServer::BedrockServer(const SData& args)
...
-        plugins.push_back(plugin);
-
-        // If the plugin has version info, add it to the list.
-        auto info = plugin->getInfo();
-        auto iterator = info.find("version");
-        if (iterator != info.end()) {
-            versions.push_back(plugin->getName() + "_" + iterator->second);
+
+        // Does this plugin need secured data? If the plugin already has all of
+        // it's secure data, this should return 0. Otherwise, it should return
```
tylerkaraszewski: comment out-of-date

BedrockServer.cpp
```diff
@@ -1004,21 +1015,56 @@ BedrockServer::BedrockServer(const SData& args)
...
+        // the number of entries this plugin requires. A plugin will already have
+        // it's secure data if the BedrockServer gets destroyed and re-created
+        // but the binary is not terminated. This happens when it gets interrupted to
+        // handle a signal (like a SIGHUP that triggers a DB backup). If they need
+        // secure data, we'll skip initializing them here then instead initialize
+        // them after we've loaded the data.
+        list<string> needsSecureData = plugin->needsSecureData();
+        if (needsSecureData.size()) {
+            pluginSecureDataMap.insert(make_pair(plugin, needsSecureData));
```
tylerkaraszewski: Why not `pluginSecureDataMap[plugin] = needsSecureData;`?

BedrockServer.cpp
```diff
@@ -1004,21 +1015,56 @@ BedrockServer::BedrockServer(const SData& args)
...
+            // If the plugin has version info, add it to the list.
+            auto info = plugin->getInfo();
+            auto iterator = info.find("version");
+            if (iterator != info.end()) {
+                versions.push_back(plugin->getName() + "_" + iterator->second);
+            }
+        }
+    }
+
+    // If any plguins needs secure data, load that now, then initialize.
```
tylerkaraszewski: plguins

BedrockServer.cpp
```diff
@@ -1004,21 +1015,56 @@ BedrockServer::BedrockServer(const SData& args)
...
+    if (!pluginSecureDataMap.empty()) {
+        _loadSecureData(pluginSecureDataMap, args);
+
+        for (auto plugin : pluginSecureDataMap) {
```
tylerkaraszewski: Why not just `auto plugin : plugins` and do them all here? Then we can remove the repeated code in the `else` block above? (the `if` three lines up would only wrap the `_loadSecureData` call)

BedrockServer.cpp
```diff
@@ -1054,6 +1100,12 @@ BedrockServer::BedrockServer(const SData& args)
     SINFO("Opening control port on '" << _args["-controlPort"] << "'");
     _controlPort = openPort(_args["-controlPort"]);
 
+    // If we're bootstraping this node we need to go into detached mode here.
+    // The syncWrapper will handle this for us.
+    if (_detach) {
+        SWARN("Bootstrap flag detected, starting sync node in detach mode.");
```
tylerkaraszewski: SINFO

BedrockServer.cpp
```diff
@@ -1862,3 +1914,87 @@ void BedrockServer::_acceptSockets() {
     }
 }
 
+void BedrockServer::_loadSecureData(const map<BedrockPlugin*, list<string>>& pluginSecureDataMap, const SData& args) {
+    // Do we need to load secure data? Skip loading if we don't. In this case, plugins
+    // should manage loading debug secure data inside of the plugin.
+    if (!args.isSet("-loadSecureData")) {
```
tylerkaraszewski: Why even have this flag? Plugins can not ask for this data when `-debug` is set, right?

BedrockServer.cpp
```diff
@@ -1862,3 +1914,87 @@ void BedrockServer::_acceptSockets() {
...
+                    plugin.first->loadSecureData(secureData);
+
+                    // We need one less key for this plugin now.
+                    fclose(file);
+                } else {
+                    SWARN("No secure data file " << fileName << " found for plugin " << pluginName << ".");
+                }
+            } else {
+                // Open a port to listen for secure data request
+                int p = S_socket(keyHost, true, true, true);  // tcp, port, blocking
```
tylerkaraszewski: What interface does this listen on?

BedrockServer.cpp
```diff
@@ -1862,3 +1914,87 @@ void BedrockServer::_acceptSockets() {
...
+                    FILE* file = fopen(fileName.c_str(), "r");
+                    if (!file) {
+                        SWARN("Couldn't open file secure data file " << fileName << " for reading. Error: " << errno << ", " << strerror(errno) << ".");
+                        continue;
+                    }
+                    SINFO("Successfully opened secure data file " << fileName << " for reading.");
+
+                    // Read the whole file into a buffer, these should be small text files,
+                    // buf if the file is too large we could die here with bad_alloc.
+                    size_t fileSize = SFileSize(fileName);
```
tylerkaraszewski: SFileLoad

BedrockServer.cpp
```diff
@@ -1862,3 +1914,87 @@ void BedrockServer::_acceptSockets() {
...
+                    char* buf = new char[fileSize];
+                    fread(buf, 1, fileSize, file);
+
+                    // Move the file into an SData object and pass it to the plugin it belongs to.
+                    secureData.deserialize(string(buf));
+                    plugin.first->loadSecureData(secureData);
```
tylerkaraszewski: Why do we make SData's out of these instead of passing the raw string to the plugin?

I'm thinking of a key that looks like an SSL cert or something.

BedrockServer.cpp
```diff
@@ -1862,3 +1914,87 @@ void BedrockServer::_acceptSockets() {
...
+                } else {
+                    SWARN("No secure data file " << fileName << " found for plugin " << pluginName << ".");
+                }
+            } else {
+                // Open a port to listen for secure data request
+                int p = S_socket(keyHost, true, true, true);  // tcp, port, blocking
+                SASSERT(p >= 0);
+
+                // Wait for a socket
+                SINFO("Waiting for secure data #" << fileName << " for plugin " << pluginName);
```
tylerkaraszewski: remove `#`

BedrockServer.cpp
```diff
@@ -1862,3 +1914,87 @@ void BedrockServer::_acceptSockets() {
...
+                sockaddr_in fromAddr;
+                int s = S_accept(p, fromAddr, true); // blocking
+                SASSERT(s >= 0);
+
+                // Receive everything
+                SINFO("Receiving secure data #" << fileName << " from " << fromAddr << " for plugin " << pluginName);
```
tylerkaraszewski: remove `#`

BedrockServer.cpp
```diff
@@ -1862,3 +1914,87 @@ void BedrockServer::_acceptSockets() {
...
+                    alive = S_recvappend(s, recvBuffer);
+                }
+
+                // Done with this socket
+                close(s);
+                s = 0;
+
+                // Send the plugin what we got
+                plugin.first->loadSecureData(secureData);
+                SINFO("Done receiving secure data #" << fileName << " for plugin " << pluginName);
```
tylerkaraszewski: remove `#`

libstuff/libstuff.cpp
```diff
@@ -1471,7 +1471,8 @@ string SGZip(const string& content) {
     }
 
     status = deflate(&stream, Z_FINISH);
-    if (status != Z_STREAM_END) {
+    if (status != Z_STREAM_END && status != Z_OK) {
+        SDEBUG("We deflated but we didn't get Z_STREAM_END or Z_OK, we got " << status);
```
tylerkaraszewski: SHMMM, or is this super common?

libstuff/libstuff.cpp
```diff
@@ -1910,6 +1911,31 @@ string SAESEncrypt(const string& buffer, unsigned char* iv, const string& key) {
     return encryptedBuffer;
 }
 
+string SAESEncrypt(char* buffer, const int bufferSize, const string& ivStr, const string& key) {
```
tylerkaraszewski: Can the existing `string SAESEncrypt(const string& buffer, unsigned char* iv, const string& key)` just call this one?

I.e.,
```
string SAESEncrypt(const string& buffer, unsigned char* iv, const string& key) {
    return SAESEncrypt(buffer.c_str(), buffer.size(), iv, key);
}
```

libstuff/libstuff.cpp
```diff
@@ -1948,6 +1975,34 @@ string SAESDecrypt(const string& buffer, const string& ivStr, const string& key)
     return SAESDecrypt(buffer, iv, key);
 }
 
+// These decrypt functions are used to return a value that still includes possible
+// padding, so it is up the caller to manage stripping the potential NULL chars off the end.
+string SAESDecryptNoStrip(const string& buffer, const size_t& bufferSize, unsigned char* iv, const string& key) {
```
tylerkaraszewski: Similar to above - can we have the stripping one call this one and then do the stripping?

main.cpp
```diff
@@ -283,6 +283,9 @@ int main(int argc, char* argv[]) {
         SDEBUG("Resetting database");
         string db = args["-db"];
         unlink(db.c_str());
+    } else if (args.isSet("-bootstrap")) {
+        // Allow for bootstraping a node with no database file in place.
+        SWARN("Loading in bootstrap mode, skipping check for database existance.");
```
tylerkaraszewski: SINFO

BedrockServer.cpp
coleaeason: As discussed offline, i'll just expose attach/detach to plugins set _detach() and back out this change so we stop double deserializing commands that come through here.

BedrockServer.cpp
coleaeason: Because we call `plugin->needsSecureData()` before we call `plugin->initialize()`, the plugins don't have a reference to args yet, so they don't know what flags are or are not set. I could pass a reference to `needsSecureData()` for args if you think that's better.

BedrockServer.cpp
coleaeason: Ah yeah, I think this is a good generalization of this code. All of our secure data is currently in the name:value format so SData makes sense, but I can see where a plugin would just load a single string instead.

BedrockServer.cpp
coleaeason: Well you have to call `_loadSecureData` before you call initialize, so we can't initialize all of the plugins above and then call _loadSecureData.

tylerkaraszewski: Yeah, the idea is don't do it above, just do it here.

BedrockPlugin.h
```diff
@@ -82,6 +87,9 @@ class BedrockPlugin {
     // Set to true if we don't want to log timeout alerts, and let the caller deal with it.
     virtual bool shouldSuppressTimeoutWarnings();
 
+    // Loads a given secure data object into the plugins local secure data map.
```
quinthar: Can you elaborate on what "secure data" means?  As someone coming into this cold, I can't figure out what it does and what makes it "secure".

tylerkaraszewski: Nothing really makes it secure, honestly. This is for key loading, but we aren't doing anything special with regards to security around it.

quinthar:
Can you add an explanation of what this whole thing is to the comment?

On Mon, Apr 30, 2018 at 5:45 PM, Tyler Karaszewski <notifications@github.com

BedrockPlugin.cpp
```diff
@@ -48,6 +48,11 @@ STable BedrockPlugin::getInfo() {
     return STable();
 }
 
+list<string> BedrockPlugin::needsSecureData(const SData& args) {
+    list<string> data;
+    return data;
```
tylerkaraszewski: I think you could just do `return {};` but it's functionally the same.

BedrockPlugin.h
```diff
@@ -33,6 +33,11 @@ class BedrockPlugin {
     // manner, as it will be called from a different thread than any processing commands.
     virtual STable getInfo();
 
+    // Returns the number of secure data entries required by this plugin.
```
tylerkaraszewski: Replace "number of" with "location of" or similar?

BedrockServer.cpp
```diff
@@ -41,7 +41,7 @@ void BedrockServer::acceptCommand(SQLiteCommand&& command, bool isNew) {
             SINFO("Forcing QUORUM consistency for command " << command.request.methodLine);
         }
         SINFO("Queued new '" << command.request.methodLine << "' command from bedrock node, with " << _commandQueue.size()
-              << " commands already queued.");
+                << " commands already queued.");
```
tylerkaraszewski: The original one matches our more common indentation for wrapped lines.

BedrockServer.cpp
```diff
@@ -1004,13 +1004,45 @@ BedrockServer::BedrockServer(const SData& args)
...
         BedrockPlugin* plugin = registeredPluginMap[SToLower(pluginName)];
         if (!plugin) {
             SERROR("Cannot find plugin '" << pluginName << "', aborting.");
         }
-        plugin->initialize(args, *this);
+
+        // Does this plugin need secured data? Plugin secure data is a method of
+        // storing sensitive data that you wouldn't want to commit into your code
+        // base (API keys, encryption keys, etc). Bedrock will allow a plugin to
+        // define a set of system file locations to load secure data from it (if
```
tylerkaraszewski: nitpicky: "it" in this sentence seems weird.

BedrockServer.cpp
```diff
@@ -1004,13 +1004,45 @@ BedrockServer::BedrockServer(const SData& args)
...
+        // the plugin requires any). These locations can be prefaced with `socket:`,
+        // in which case Bedrock will read the secure data from a socket (specified
```
tylerkaraszewski: What is prefixed with `socket`? Any random string, or the location of the socket to read from? If the second, what do `-secureDataHost` and the default do?

BedrockServer.cpp
tylerkaraszewski: Did we decide to do the thing in your previous comment?

BedrockServer.cpp
```diff
@@ -1862,3 +1909,80 @@ void BedrockServer::_acceptSockets() {
...
+
+    // Loop through the plugins that require secure data, if the file name
+    // is prefixed with socket:, load the data from a socket instead.
+    for (auto plugin : pluginSecureDataMap) {
+        const string& pluginName = SToLower(plugin.first->getName());
+
+        for (auto file : plugin.second) {
+            const string& fileName = file;
+
+            if (!SStartsWith(fileName, "socket")) {
```
tylerkaraszewski: should be `socket:` right?

BedrockServer.cpp
```diff
@@ -1862,3 +1909,80 @@ void BedrockServer::_acceptSockets() {
...
+                    plugin.first->loadSecureData(fileContents);
+
+                    // We need one less key for this plugin now.
+                    fclose(file);
+                } else {
+                    SWARN("No secure data file " << fileName << " found for plugin " << pluginName << ".");
+                }
+            } else {
+                // Open a port to listen for secure data request
+                int p = S_socket(secureDataHost, true, true, true);  // tcp, port, blocking
```
tylerkaraszewski: Like mentioned before, this seems like it should be whatever followed `socket:`

BedrockServer.cpp
```diff
@@ -1862,3 +1909,80 @@ void BedrockServer::_acceptSockets() {
...
+                if (SFileExists(fileName)) {
+                    FILE* file = fopen(fileName.c_str(), "r");
+                    if (!file) {
+                        SWARN("Couldn't open file secure data file " << fileName << " for reading. Error: " << errno << ", " << strerror(errno) << ".");
+                        continue;
+                    }
+                    SINFO("Successfully opened secure data file " << fileName << " for reading.");
+
+                    // Read the whole file into a string and pass it to the plugin it belongs to.
+                    string fileContents = SFileLoad(fileName);
```
tylerkaraszewski: It seems like it'd be easier for a plugin implementor to just call `SFileLoad` on a key file himself rather than implementing this API.

BedrockServer.h
```diff
@@ -203,6 +203,10 @@ class BedrockServer : public SQLiteServer {
     // Send a command to all of our peers. It will be wrapped appropriately.
     void broadcastCommand(const SData& message);
 
+    // Set the detach state of the server. Setting to True will cause the server to detach from the database and go
+    // into a sleep loop until this is called again with False
```
tylerkaraszewski: lowercase `false`

libstuff/libstuff.cpp
```diff
@@ -1928,24 +1996,20 @@ string SAESDecrypt(const string& buffer, unsigned char* iv, const string& key) {
...
                           (unsigned char*)decryptedBuffer.c_str());
-
-    // Trim off the padding.
-    int size = (int)decryptedBuffer.find('\0');
-    if (size != (int)string::npos) {
-        decryptedBuffer.resize(size);
-    }
     return decryptedBuffer;
+
+
```
tylerkaraszewski: Don't need these

libstuff/libstuff.cpp
```diff
@@ -2122,6 +2186,7 @@ string SHashSHA1(const string& buffer) {
 }
 
 string SHashSHA256(const string& buffer) {
+    // Just add and return
```
tylerkaraszewski: Add what?

PR #317, opened 2017-10-31
Demoting SQLite assert warning to hmmm

brandonhub:
@tylerkaraszewski pls review.

Demoting a log that was spamming us.

### Fixes issue
$ https://github.com/Expensify/Expensify/issues/63698

# Tests
1. Did a bunch of read only operations like fetching expense, reports, etc.
2. Checked the logs and made sure that I saw "setting query_only on" and "setting query_only off" indicating that the pragma was being turned on and off for those requests.
3. Also made sure each "setting query_only_off" was proceeded by a matching "setting query_only off" to make sure we never end up setting the flag and leaving it on
4. I left Bedrock on this branch throughout an entire day as I moved on to other issues and made sure no weirdness occurred.

# QA
none

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -298,9 +298,12 @@ string SQLite::read(const string& query) {
 }
 
 bool SQLite::read(const string& query, SQResult& result) {
+    
     // Execute the read-only query
+    if (!SContains(SToUpper(query), "UPDATE ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
+    }
     SASSERTWARN(!SContains(SToUpper(query), "INSERT "));
-    SASSERT(!SContains(SToUpper(query), "UPDATE "));
     SASSERTWARN(!SContains(SToUpper(query), "DELETE "));
     SASSERTWARN(!SContains(SToUpper(query), "REPLACE "));
     uint64_t before = STimeNow();
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -300,7 +300,7 @@ string SQLite::read(const string& query) {
 bool SQLite::read(const string& query, SQResult& result) {
     // Execute the read-only query
     SASSERTWARN(!SContains(SToUpper(query), "INSERT "));
-    SASSERTWARN(!SContains(SToUpper(query), "UPDATE "));
+    SASSERT(!SContains(SToUpper(query), "UPDATE "));
     SASSERTWARN(!SContains(SToUpper(query), "DELETE "));
     SASSERTWARN(!SContains(SToUpper(query), "REPLACE "));
     uint64_t before = STimeNow();
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -303,9 +303,15 @@ bool SQLite::read(const string& query, SQResult& result) {
     if (!SContains(SToUpper(query), "UPDATE ")) {
         SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
     }
-    SASSERTWARN(!SContains(SToUpper(query), "INSERT "));
-    SASSERTWARN(!SContains(SToUpper(query), "DELETE "));
-    SASSERTWARN(!SContains(SToUpper(query), "REPLACE "));
+    if (!SContains(SToUpper(query), "INSERT ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
+    }
+    if (!SContains(SToUpper(query), "DELETE ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
+    }
+    if (!SContains(SToUpper(query), "REPLACE ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
+    }
     uint64_t before = STimeNow();
     bool queryResult = !SQuery(_db, "read only query", query, result);
     _checkTiming("timeout in SQLite::read"s);
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -238,15 +238,6 @@ bool SQLite::beginConcurrentTransaction() {
     return _insideTransaction;
 }
 
-bool SQLite::setQueryOnlyPragma(bool on) {
-    if (on) {
-        SASSERT(!SQuery(_db, "setting query_only on", "PRAGMA query_only = true;"));
-    } else {
-        SASSERT(!SQuery(_db, "setting query_only off", "PRAGMA query_only = false;"));
-    }
-    return true;
-}
-
 bool SQLite::verifyTable(const string& tableName, const string& sql, bool& created) {
     // sqlite trims semicolon, so let's not supply it else we get confused later
     SASSERT(!SEndsWith(sql, ";"));
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -304,13 +304,13 @@ bool SQLite::read(const string& query, SQResult& result) {
         SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
     }
     if (!SContains(SToUpper(query), "INSERT ")) {
-        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"INSERT \")) != true");
     }
     if (!SContains(SToUpper(query), "DELETE ")) {
-        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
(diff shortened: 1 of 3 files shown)


brandonhub:
Updated
> You haven't demoted a warning to a hmmm, you've promoted it to a forced crash

Yeah, I guess I wasn't thinking there, thanks for pointing that out. I created an if statement which checked the same condition as the assertion but logs a HMMM if it fails instead of actually asserting. Let me know if that's not the customary way to do that.

>  Also, this is almost certainly not the right fix for this problem.

So in the [original issue](https://github.com/Expensify/Expensify/issues/63698), a demotion is what we agreed upon. I thought that the discussion was finished, but if you think we shouldn't demote then I'm happy to try something else!

sqlitecluster/SQLite.cpp
```diff
@@ -298,11 +298,20 @@ string SQLite::read(const string& query) {
...
     // Execute the read-only query
-    SASSERTWARN(!SContains(SToUpper(query), "INSERT "));
-    SASSERTWARN(!SContains(SToUpper(query), "UPDATE "));
-    SASSERTWARN(!SContains(SToUpper(query), "DELETE "));
-    SASSERTWARN(!SContains(SToUpper(query), "REPLACE "));
+    if (!SContains(SToUpper(query), "UPDATE ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
+    }
+    if (!SContains(SToUpper(query), "INSERT ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
```
coleaeason: copy paste error, `UPDATE` should be `INSERT`

sqlitecluster/SQLite.cpp
```diff
@@ -298,11 +298,20 @@ string SQLite::read(const string& query) {
...
+    }
+    if (!SContains(SToUpper(query), "DELETE ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
```
coleaeason: copy paste error, `UPDATE` should be `DELETE`

sqlitecluster/SQLite.cpp
```diff
@@ -298,11 +298,20 @@ string SQLite::read(const string& query) {
...
+    }
+    if (!SContains(SToUpper(query), "REPLACE ")) {
+        SHMMM("Assertion failed: (!SContains(SToUpper(query), \"UPDATE \")) != true");
```
coleaeason: copy paste error, `UPDATE` should be `REPLACE`

brandonhub:
updated, apologies for the careless errors

coleaeason:
Ths l

coleaeason:
This looks good to me, unless @tylerkaraszewski has a better solution.

The only thing I can think of is instead of turning these into IF statements we could make a new `SASSERTHMMM` to match `SASSERTWARN` to DRY this PR up.

sqlitecluster/SQLite.cpp
```diff
@@ -298,11 +298,20 @@ string SQLite::read(const string& query) {
 }
 
 bool SQLite::read(const string& query, SQResult& result) {
+    
     // Execute the read-only query
-    SASSERTWARN(!SContains(SToUpper(query), "INSERT "));
-    SASSERTWARN(!SContains(SToUpper(query), "UPDATE "));
-    SASSERTWARN(!SContains(SToUpper(query), "DELETE "));
-    SASSERTWARN(!SContains(SToUpper(query), "REPLACE "));
+    if (!SContains(SToUpper(query), "UPDATE ")) {
```
flodnv: Don't you want to log something if `SContains(SToUpper(query), "UPDATE ")` ? 😕 
I'm kind of even questioning why we'd want that. If you still want it, please use something other than exactly what `SASSERTWARN` would output. (eg `SHMMM("Query contains 'UPDATE'")`). There's not assertion that failed if you removed the assertion.

brandonhub: good point

coleaeason:
I have no idea why I closed this...I think i meant to click comment and clicked close and comment???

brandonhub:
> I have no idea why I closed this...I think i meant to click comment and clicked close and comment???

Oh whoops, I thought maybe you and @tylerkaraszewski had talked 1:1 or something.

> please use something other than exactly what SASSERTWARN would output. (eg SHMMM("Query contains 'UPDATE'")). There's not assertion that failed if you removed the assertion.

Agreed, I've updated the messages accordingly

>Don't you want to log something if SContains(SToUpper(query), "UPDATE ") ? 😕
I'm kind of even questioning why we'd want that. If you still want it

Yes, also updated.

quinthar:
A much better solution than that really weak check would be to use this pragma in peek: https://sqlite.org/pragma.html#pragma_query_only

flodnv:
Your comment got me thinking @quinthar -- isn't this code (currently in master) already going to fail if you do `DB::read("INSERT INTO something;")` ? I'd assume it will.. in which case, I am once again questioning the usefulness of this log.

tylerkaraszewski:
> isn't this code (currently in master) already going to fail if you do DB::read("INSERT INTO something;") ?

I think this might succeed (which is bad). If it fails, it doesn't fail until after we write the query and notice he DB changed, and then does so catastrophically (I remember this code, but can't currently find it). @quinthar's suggestion or something similar is much better.

brandonhub:
What do we think of simply adding two additional SQuery calls to `SQLite::read` function? Basically, we add a `PRAGMA query_only = true;` and a `PRAGMA query_only = false;` to the beginning and end of our query respectively. So SQLIte::read would look something like this:
```
    uint64_t before = STimeNow();
    SQuery(_db, "setting query_only on", "PRAGMA query_only = true;");
    bool queryResult = !SQuery(_db, "read only query", query, result);
    SQuery(_db, "setting query_only off", "PRAGMA query_only = false;");
    _checkTiming("timeout in SQLite::read"s);
    _readElapsed += STimeNow() - before;
    return queryResult;
```

Thoughts?

quinthar:
That seems great to me.

coleaeason:
Modifying the pragma for every read turns every read operation into three sqlite operations. Is there some increase in overhead to be considered here? We do a lot of reads.

quinthar:
Hm... good point.  Maybe we do it at the start of "peek" and undo at the
end?

brandonhub:
Sure there would be some increase in overhead...but we basically do the same thing in `SQLite::write` [here](https://github.com/Expensify/Bedrock/blob/master/sqlitecluster/SQLite.cpp#L356-L371) (albeit wer're reading a PRAGMA instead of setting on) and that doesn't seem to be causing us any issues.

tylerkaraszewski:
I doubt this adds significant overhead, we'd have to ask the sqlite team to make sure, but I doubt this pragma does much more than set a flag that the statement parser checks when it parses out a verb to act on. But I agree that adding it at the beginning (and turning it off at the end) of `peek` is abetter solution that adding it directly to `read`.

brandonhub:
To clarify, @tylerkaraszewski are you suggesting we would basically turn the pragma on and off by running `PRAGMA query_only = true/false;`  right before and after [this line](https://github.com/Expensify/Server-Expensify/blob/master/auth/Auth.cpp#L553)? (and I guess [this one](https://github.com/Expensify/Server-Expensify/blob/master/auth/Auth.cpp#L633) too).

tylerkaraszewski:
@brandonhub I would me the change [here](https://github.com/Expensify/Bedrock/blob/master/BedrockCore.cpp#L10) instead, and also make sure it gets unset in exception handlers.

brandonhub:
Cool makes sense, thanks

brandonhub:
Okay so given our discussion above, we concluded to turn the `query_only` flag on/off in the `bedrock::peekCommand` function. I thought it was weird to directly run an `SQuery` in the body of peekCommand because that would require me to import `libstuff.h` into `BedrockCore.cpp`, so instead I made a little function in` SQLite.cpp` called `setReadOnlyPragma` which takes a boolean and sets the `query_only` value on or off. Then I called `setReadOnlyPragma(true)` right after the transaction starts and `setReadOnlyPragma(false)` right after the command finishes.

I'm marking this as a WIP for now since we've sorta been iterating on this for a while.

Also, I'm happy to reach out to SQLite concerning the performance implications of setting that PRAGMA on each peek (although I doubt performance will be an issue)  if somebody gives me an email address to send something too, or perhaps somebody else can do that who has an actual relationship with the SQLite staff.

tylerkaraszewski:
Why call `SQuery` directly instead of `_db.read`? Also, we should turn off the pragma in the other (`...`) exception handler as well.

brandonhub:
Good catch, updated

brandonhub:
bump

tylerkaraszewski:
Code looks fine. Anyone else care to review?

coleaeason:
Should we also set the pragma to false at the top of process to make sure we never get to process without being able to write?

tylerkaraszewski:
I think all that would do is mask some bug where where we failed to un-set it in peek.

coleaeason:
What if we added it as an SASSERT? I'm trying to avoid a crash because we get to process and aren't able to write to the db.

tylerkaraszewski:
SASSERT causes a crash.

coleaeason:
After our short discussion offline, let's merge as is.

coleaeason:
Saving for high risk deploy window just incase my fears come true.

PR #363, opened 2018-01-31
Update bedrock config files for Ubuntu 14.04 and 16.04

swhi3635:
The package successfully installs the configs and the service starts without issue.  Tested both 14.04 and 16.04 packages.

For both 14.04 and 16.04, I added a dependency for `libstdc++6 > 7.1`, since the package that comes installed on 14.04 and 16.04 don't have `GLIBCXX_3.4.21` (required for bedrock).  This avoids the following error: 
```
version GLIBCXX_3.4.21 not defined in file libstdc++.so.6 with link time reference
```


configs/bedrock_prerequisites.service
```diff
@@ -0,0 +1,11 @@
...
+Description=Generates prerequisite files for bedrock.service, if absent
+Documentation=http://bedrockdb.com/
+
+[Service]
+Type=oneshot
+EnvironmentFile=/etc/bedrock/bedrock.conf
+
+# The following creates database dir and database file
+ExecStart=/bin/sh -c 'mkdir -p "$(dirname "${BEDROCK_DB_PATH}")"'
+ExecStart=/bin/sh -c 'test -f "${BEDROCK_DB_PATH}" || touch "${BEDROCK_DB_PATH}"'
```
coleaeason: is this needed? sqlite should create the database if non exists at the path.

configs/bedrock.service
```diff
@@ -0,0 +1,25 @@
...
+
+ExecStart=/usr/sbin/bedrock \
+        -nodeName ${THISNODE} \
+        -db ${BEDROCK_DB_PATH} \
+        -serverHost ${SERVER_HOST} \
+        -nodeHost ${NODE_HOST} \
+        -priority ${PRIORITY} \
+        -cache ${CACHE_SIZE} \
+        ${VERBOSE} ${PEERS} $EXTRA_FLAGS
+ExecStop=/bin/kill -HUP $MAINPID
```
coleaeason: `-HUP` won't kill the process, it's going to start a backup. This should be `-TERM` instead.

configs/bedrock.conf
```diff
@@ -0,0 +1,23 @@
...
+# Cache size in kilobytes
+CACHE_SIZE="10001"
+
+VERBOSE="-v"
+
+# Bind to network address
+SERVER_HOST="0.0.0.0:8888"
+NODE_HOST="0.0.0.0:8889"
+
+#EXTRA_FLAGS="-readThreads 1 -plugins db,jobs,cache,mysql"
```
coleaeason: This won't work, you can only have a value stored in these variables, not full flags and values.

Additionally, readThreads is deprecated, workerThreads should be used.

configs/bedrock.conf
```diff
@@ -0,0 +1,23 @@
...
+BEDROCK_DB_PATH="/var/lib/bedrock/bedrock.db"
+
+# Single node, unknown environment
+THISNODE="bedrock"
+PRIORITY=200
+
+# Cache size in kilobytes
+CACHE_SIZE="10001"
+
+VERBOSE="-v"
```
coleaeason: I don't think we should enable verbose logging by default.

configs/bedrock.service
```diff
@@ -0,0 +1,25 @@
...
+
+# The following creates database file if absent
+After=bedrock_prerequisites.service
+Requires=bedrock_prerequisites.service
+
+[Service]
+EnvironmentFile=/etc/bedrock/bedrock.conf
+PIDFile=/var/run/bedrock.pid
+
+Type=simple
```
coleaeason: This needs to be forking so the pid is managed properly.

configs/bedrock_prerequisites.service
swhi3635: Yes it is needed.  Bedrock/SQLite does not automatically create the database if it does not already exist.

configs/bedrock.conf
swhi3635: >This won't work, you can only have a value stored in these variables, not full flags and values.

Are you saying this doesn't follow some convention?  Because it definitely works.

swhi3635:
16.04 package works when built on Xenial.  I decided to test Trusty as well, and ran into this:
```
Selecting previously unselected package bedrock.
(Reading database ... 65168 files and directories currently installed.)
Preparing to unpack bedrock_2018.01.30~522e696-14.04_all-2.deb ...
Unpacking bedrock (2018.01.30~522e696-14.04) ...
Setting up bedrock (2018.01.30~522e696-14.04) ...
/var/lib/dpkg/info/bedrock.postinst: line 5: /etc/init.d/bedrock: Permission denied
dpkg: error processing package bedrock (--install):
 subprocess installed post-installation script returned error exit status 126
Processing triggers for ureadahead (0.100.0-16) ...
Errors were encountered while processing:
 bedrock
```

/etc/init.d/bedrock doesn't have execute permissions, so I need to sort that out before we're ready to do another release.

swhi3635:
Getting this when installing the Trusty package: 
```
 * Starting Expensify Bedrock Server bedrock                                                         /usr/sbin/bedrock: relocation error: /usr/sbin/bedrock: symbol _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEC1Ev, version GLIBCXX_3.4.21 not defined in file libstdc++.so.6 with link time reference
```

This package was built on a Xenial system. So I'll have to try building from Trusty (I assume that's the issue).  

So we'll probably have to separate our builds for 14.04 and 16.04, which is....not surprising.

coleaeason:
@swhi3635 yeah, you need to build on 14.04 for 14.04, and 16.04 for 16.04, (this is why we have the two different build servers.)

configs/bedrock.conf
```diff
@@ -0,0 +1,28 @@
...
+
+# Cache size in kilobytes.  Default: 1GB
+CACHE_SIZE="-cacheSize 10000"
+
+# Bind to network address
+SERVER_HOST="0.0.0.0:8888"
+NODE_HOST="0.0.0.0:8889"
+
+# Number of worker threads.  Default: # of CPUs on the system.
+WORKER_THREADS="-workerThreads 1"
```
coleaeason: we should leave this commented out, and let bedrock detect and default to the number of multithreaded available cpus.

configs/bedrock.init
```diff
@@ -36,7 +36,7 @@ if [ -n "$OTHERNODES" ]; then
     FLAGS="-peerList $OTHERNODES"
 fi
 
-DAEMON_OPTS="-fork -nodeName $THISNODE -db $BEDROCK_DB_PATH -serverHost 0.0.0.0:8888 -nodeHost 0.0.0.0:8889 -priority $PRIORITY -pidfile $PIDFILE -quorumCheckpoint 100 -readThreads 4 -plugins status,db,jobs,cache,mysql $FLAGS"
+DAEMON_OPTS="-fork -nodeName $THISNODE -db $BEDROCK_DB_PATH -serverHost 0.0.0.0:8888 -nodeHost 0.0.0.0:8889 -priority $PRIORITY -pidfile $PIDFILE -workerThreads 1 $FLAGS"
```
coleaeason: same as above, we should remove the workerThreads flag and let bedrock find the number of cores to use.

configs/bedrock.service
```diff
@@ -0,0 +1,31 @@
...
+ExecStart=/usr/sbin/bedrock \
+        -fork \
+        -pidFile /var/run/bedrock.pid \
+        -nodeName ${THISNODE} \
+        -db ${BEDROCK_DB_PATH} \
+        -serverHost ${SERVER_HOST} \
+        -nodeHost ${NODE_HOST} \
+        -priority ${PRIORITY} \
+        ${PLUGINS} \
+        ${WORKER_THREADS} \
```
coleaeason: we should remove this

swhi3635: This should stay, so the users can simply uncomment the worker threads option in the bedrock.conf file without having to also modify the service file.

coleaeason: Ah that makes sense.

coleaeason:
Before i merge can you please remove that extra WARN we found in mysql.h's `getPort()`?

PR #390, opened 2018-03-07
Demote some warnings

iwiznia:
This is already handled by BWM properly, so no need to warn.

Fixes:
$https://github.com/Expensify/Expensify/issues/72681
$https://github.com/Expensify/Expensify/issues/72634

Tests:
- Try to finish a job that's not running
- Try to finish a child job whose parent job is not paused.
- Try to fail a job that's not running.
- Check logs to see they don't generate a warning.


iwiznia:
Travis passed

PR #398, opened 2018-03-19
Do not use auto increment for job IDs

iwiznia:
@flodnv please review

One thing I just realized is that this will reuse jobIDs (when they are not used anymore). This is fine as nothing will break if we do that, but it will make finding a specific job run in the logs a bit harder. I think it's fine anyway.

Tests:
- Start bedrock without migrating the table
- Start bedrock after [migrating the table](https://github.com/Expensify/Expensify/issues/75339#issuecomment-374274942)
- Create a job, check it generates a correct ID.


iwiznia:
Fucking tests passed when retried.

PR #402, opened 2018-03-20
Use new version of sqlite with improvements on the sqlite_sequence table

iwiznia:
Updating the version of sqlite.


iwiznia:
Merge? Or are we holding for the high risk deploy window?

iwiznia:
Oh, BTW @timszot the comment when updating sqlite did not work 😞

PR #407, opened 2018-03-27
Add a SHA256 hash function

ctkochan22:
@flodnv @coleaeason   Please review
cc @tylerkaraszewski 

**Issue:**
We need to authenticate a webhook response from Stripe, and Cole and Flo wanted it to be done so in Auth. However, we only have methods in place for SHA1 hash functions.

**Solution:**
We already had a mbed_sha256 hash function already available, so I just created a wrapper method that uses that so that we can have a SHA256 hash function.

Overview (more for myself than you guys as a lot of this was self taught tonight)
- Block size is set to 64 bytes
- Inner padding is byte 0x36
- Outer padding is byte 0x5
- Output is 256bits/8bit = 32

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -8,6 +8,7 @@
 #include <mbedtls/aes.h>
 #include <mbedtls/base64.h>
 #include <mbedtls/sha1.h>
+#include <mbedtls/sha256.h>
 
 // Additional headers
 #include <netdb.h>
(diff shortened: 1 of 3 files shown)


ctkochan22:
Hey guys, so I went off of flo's hint of looking at `SHashSHA1` and created a `SHashSHA256`, which was super f*&king cool. But a lot of this was self-taught in the past hour or so... so please look over with care. 

On the bright side, i tested it against Stripe's signature hash vs the internal one created with ^^ method and it works! 

This is currenlty in WIP so that you guys can make sure I'm at least somewhat going in the right direction, and so I can write tests.

flodnv:
Cool! What do you think @tylerkaraszewski ? Handing the review to you 🙇

ctkochan22:
Just for transparency and easy tracking for myself, this PR was originally made for this issue: https://github.com/Expensify/Expensify/issues/75135

PR #409, opened 2018-03-28
[HOLD] Fix edge case for large payloads

coleaeason:

As we discussed, we shouldn't ASSERT on this non-catastrophic case where we don't send our entire payload, then try to remove only the part of it that we did send. I think the comment accurately describes my fix here, but let me know if you had something different in mind or a reason this won't work.

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -525,7 +525,6 @@ bool SParseList(const char* ptr, list<string>& valueList, char separator) {
 
 // --------------------------------------------------------------------------
 void SConsumeFront(string& lhs, ssize_t num) {
-    SASSERT((int)lhs.size() >= num);
     // If nothing, early out
     if (!num)
         return;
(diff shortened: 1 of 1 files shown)


libstuff/libstuff.cpp
```diff
@@ -525,7 +525,6 @@ bool SParseList(const char* ptr, list<string>& valueList, char separator) {
 
 // --------------------------------------------------------------------------
 void SConsumeFront(string& lhs, ssize_t num) {
-    SASSERT((int)lhs.size() >= num);
```
quinthar: Hm, why remove this?  The only time this alerts is if we think we sent *more* than the buffer size, which seems pretty extreme.

coleaeason: Yeah... Tyler and I are discussing in slack, we're not really sure how we died here now, as sending more than what's in the buffer shouldn't be possible.

quinthar: Ok, then let's not remove this as it sounds like we don't actually understand the real bug.  This is a protection; removing the protection isn't a great solution.

coleaeason: Agreed, we're looking for a better solution.

coleaeason:
We're gonna just close this until it happens again, as it's not at all clear what happened to cause us to hit this ASSERT and reproducing this edge case may be rather difficult and time consuming.

PR #412, opened 2018-04-05
[HOLD] Remove transition schema

iwiznia:
@flodnv please review. Holding a bit to give people to do the migration locally.

This removes the old schema check.


iwiznia:
Off hold

PR #415, opened 2018-04-09
Rename jobs priority param to jobPriority

iwiznia:
Renaming `priority` param of jobs plugin to `jobPriority` so it does not conflict with bedrock's param of the same name.


iwiznia:
Bump

PR #456, opened 2018-05-10
Make the BedrockServer object long lived, fix attach/detach

coleaeason:
@tylerkaraszewski Please review. This PR makes it so that the BedrockServer object isn't destroyed until the binary terminates. This allows us to ensure the server object never goes away while a plugin is depending on it. It also fixes attach/detach, which is currently broken and only works because after we hit the remainder of the 60 second gracefulshutdown timeout we destroy and recreate the bedrock server. 

## Tests
Tested `BeginBackup`, `Detach`,`Attach`, and shutdown, all function as expected.


coleaeason:
@tylerkaraszewski updated

PR #485, opened 2018-07-31
Fix partial index on jobsParentJobIDState

iwiznia:
@coleaeason please review. 

Needed for https://github.com/Expensify/Expensify/issues/84444 

Hopefully this will be lower (or throughput should increase):
image

Tests:
- Restart bedrock
- Stop bedrock.
- Run the migration.
- Start bedrock.

Queries tests:
```
sqlite> explain query plan SELECT jobID, data, state FROM jobs WHERE parentJobID != 0 AND parentJobID = 1 AND state IN ('FINISHED', 'CANCELLED');
QUERY PLAN
`--SEARCH TABLE jobs USING INDEX jobsParentJobIDState (parentJobID=? AND state=?)

sqlite> explain query plan DELETE FROM jobs WHERE parentJobID != 0 AND parentJobID=1 AND state IN ('FINISHED', 'CANCELLED');
QUERY PLAN
`--SEARCH TABLE jobs USING INDEX jobsParentJobIDState (parentJobID=? AND state=?)

`--SEARCH TABLE jobs USING INDEX jobsParentJobIDState (parentJobID=? AND state=?)
sqlite> explain query plan UPDATE jobs SET state='QUEUED' WHERE state='PAUSED' AND parentJobID != 0 AND parentJobID=1;
QUERY PLAN
`--SEARCH TABLE jobs USING INDEX jobsParentJobIDState (parentJobID=? AND state=?)

sqlite> explain query plan SELECT 1 FROM jobs WHERE parentJobID != 0 AND parentJobID=1 LIMIT 1;
QUERY PLAN
`--SEARCH TABLE jobs USING COVERING INDEX jobsParentJobIDState (parentJobID=?)

sqlite> explain query plan SELECT count(1) FROM jobs WHERE parentJobID != 0 AND parentJobID = 1 AND state IN ('QUEUED', 'RUNQUEUED', 'RUNNING');
QUERY PLAN
`--SEARCH TABLE jobs USING COVERING INDEX jobsParentJobIDState (parentJobID=? AND state=?)

sqlite> explain query plan SELECT 1 FROM jobs WHERE parentJobID != 0 AND parentJobID = 1 AND state IN ('QUEUED', 'RUNQUEUED', 'RUNNING', 'PAUSED') LIMIT 1;
QUERY PLAN
`--SEARCH TABLE jobs USING COVERING INDEX jobsParentJobIDState (parentJobID=? AND state=?)
```

Query plans on the DBs without the new index created:
```
[...]

PR #490, opened 2018-08-16
Optimize CancelJob peek query

pecanoro:
It was timing out for SSDos: 

$ https://github.com/Expensify/Expensify/issues/85542
$ https://github.com/Expensify/Expensify/issues/85541

Here is the old query plan:

```
ionatan@staging-www1.la:~$ sudo readdb.sh "explain query plan SELECT j.state, GROUP_CONCAT(jj.jobID), j.parentJobID FROM jobs j LEFT JOIN jobs jj ON jj.parentJobID = j.jobID WHERE j.jobID=726201773 GROUP BY j.jobID;"
QUERY PLAN
|--SEARCH TABLE jobs AS j USING INTEGER PRIMARY KEY (rowid=?)
`--SCAN TABLE jobs AS jj
```

As @iwiznia suggested, the solution is to add `parentJobID!=0`. I also simplified the whole query to make it more readable.

```
explain query plan SELECT j.state, j.parentJobID, (SELECT COUNT(1) FROM jobs WHERE parentJobID!=0 AND parentJobID=726201773) children FROM jobs j WHERE j.jobID=726201773;
----------  ----------  ----------  ----------------------------------------------------------
0           0           0           SEARCH TABLE jobs AS j USING INTEGER PRIMARY KEY (rowid=?)
0           0           0           EXECUTE SCALAR SUBQUERY 1                                 
1           0           0           SEARCH TABLE jobs USING COVERING INDEX jobsParentJobIDStat

real    0m0.006s
user    0m0.004s
sys    0m0.000s
```

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -335,12 +335,9 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
         int64_t jobID = request.calc64("jobID");
 
         SQResult result;
-        if (!db.read("SELECT j.state, GROUP_CONCAT(jj.jobID), j.parentJobID "
+        if (!db.read("SELECT j.state, j.parentJobID, (SELECT COUNT(1) FROM jobs WHERE parentJobID != 0 AND parentJobID=" + SQ(jobID) + ") children "
                      "FROM jobs j "
-                     "LEFT JOIN jobs jj ON jj.parentJobID = j.jobID "
-                     "WHERE j.jobID=" + SQ(jobID) + " "
-                        "AND jj.parentJobID != 0 "
-                     "GROUP BY j.jobID;",
+                     "WHERE j.jobID=" + SQ(jobID) + ";",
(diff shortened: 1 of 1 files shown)


iwiznia:
I missed this in https://github.com/Expensify/Bedrock/pull/485/

plugins/Jobs.cpp
```diff
@@ -335,22 +335,20 @@ bool BedrockPlugin_Jobs::peekCommand(SQLite& db, BedrockCommand& command) {
...
-                     "WHERE j.jobID=" + SQ(jobID) + " "
-                     "GROUP BY j.jobID;",
+                     "WHERE j.jobID=" + SQ(jobID) + ";",
                      result)) {
             STHROW("502 Select failed");
         }
 
         // Verify the job exists
-        if (result.empty()) {
+        if (result.empty() || result[0][0].empty()) {
```
iwiznia: NAB: I don't think this extra condition was needed

PR #512, opened 2018-10-10
Increase param size of name in GetJobs

pecanoro:
Do you want me to add a test for this or not really necessary?

## Issue: 

$ https://github.com/Expensify/Expensify/issues/86685

### DIFF
--- a/BedrockPlugin.h
+++ b/BedrockPlugin.h
@@ -21,10 +21,6 @@ class BedrockPlugin {
     static constexpr int64_t MAX_SIZE_BLOB = 1024 * 1024;
     static constexpr int64_t MAX_SIZE_SMALL = 255;
 
-    // We were using MAX_SIZE_SMALL in GetJob to check the job name, but now GetJobs accepts more than one job name,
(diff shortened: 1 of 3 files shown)


pecanoro:
I can also create a new constant with a new size, the one that I used seems too big

iwiznia:
> I can also create a new constant with a new size, the one that I used seems too big

It is technically too big, but I am not sure we can know the exact max length to use, right?

pecanoro:
> It is technically too big, but I am not sure we can know the exact max length to use, right?

Huh? I am not following. The current size is the maximum for a SQL query, we can set it to whatever we want between 0 and that size 😄

iwiznia:
lol, I know.. I meant that if 1024*1024 is the max sql query size, can we establish a specific number which will be good for the size of the name? I think not...

pecanoro:
I don't know, the average number of characters for a SSDos job is 40, are we ever going to have more than 100 queues? We could set it to 100*40 for example

iwiznia:
Sounds good to me.

pecanoro:
updated!

PR #514, opened 2018-10-15
Do not fail on empty data object

iwiznia:
Related to https://github.com/Expensify/Web-Expensify/pull/22695
Kind of needed for https://github.com/Expensify/Expensify/issues/88302

This allows the clients to send `{}` as the data.
In my tests of https://github.com/Expensify/Expensify/issues/88302 I tested with no data (null) and some data, but not an empty array... either that or I fucked up my tests before.


mattabullock:
Just to be clear, you mean an empty object, right? `{}` vs `[]`

iwiznia:
Yes

PR #526, opened 2018-11-07
retryAfter for repeatable jobs

mattabullock:
We're adding the ability to have retryAfter with repeatable jobs. The functionality will be that if you have retryAfter and a repeat and a job gets stuck, we'll pick whichever is sooner, the retryAfter or the repeat. If the retryAfter ends up running, it will offset any jobs that are done relative to queue or finish of the last run (and not beginning of the day) to be offset by the retryAfter time. Shouldn't be a big deal, just need to be aware.

# Tests
1. CreateJob with retryAfter `+5 SECONDS` and repeat `FINISHED, +10 SECONDS`
2. GetJob, confirm that nextRun is 5 seconds after lastRun.
3. FinishJob, confirm that nextRun is 10 seconds after you finished the job.

Also added some automated tests.

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -656,9 +656,9 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         string safeNumResults = SQ(max(request.calc("numResults"),1));
         bool mockRequest = command.request.isSet("mockRequest") || command.request.isSet("getMockedJobs");
         string selectQuery =
-            "SELECT jobID, name, data, parentJobID, retryAfter, created, repeat FROM ( "
+            "SELECT jobID, name, data, parentJobID, retryAfter, created, repeat, lastRun, nextRun FROM ( "
                 "SELECT * FROM ("
-                    "SELECT jobID, name, data, priority, parentJobID, retryAfter, created, repeat "
+                    "SELECT jobID, name, data, priority, parentJobID, retryAfter, created, repeat, lastRun, nextRun "
                     "FROM jobs "
                     "WHERE state IN ('QUEUED', 'RUNQUEUED') "
                         "AND priority=1000 "
@@ -669,7 +669,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                 ") "
             "UNION ALL "
                 "SELECT * FROM ("
-                    "SELECT jobID, name, data, priority, parentJobID, retryAfter, created, repeat "
+                    "SELECT jobID, name, data, priority, parentJobID, retryAfter, created, repeat, lastRun, nextRun "
                     "FROM jobs "
                     "WHERE state IN ('QUEUED', 'RUNQUEUED') "
                         "AND priority=500 "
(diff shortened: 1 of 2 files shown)


plugins/Jobs.cpp
```diff
@@ -791,10 +788,12 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
...
             SINFO("Updating jobs with retryAfter");
             for (auto job : retriableJobs) {
+                string currentTime = SCURRENT_TIMESTAMP();
+                string nextRunDateTime = _constructNextRunDATETIME(job["nextRun"], job["lastRun"] != "" ? job["lastRun"] : job["nextRun"], job["repeat"]);
                 string updateQuery = "UPDATE jobs "
                                      "SET state='RUNQUEUED', "
-                                         "lastRun=" + SCURRENT_TIMESTAMP() + ", "
-                                         "nextRun=DATETIME(" + SCURRENT_TIMESTAMP() + ", " + SQ(job["retryAfter"]) + ") "
+                                         "lastRun=" + currentTime + ", "
+                                         "nextRun=MIN(DATETIME(" + currentTime + ", " + SQ(job["retryAfter"]) + "), COALESCE(NULLIF(" + (nextRunDateTime != "" ? nextRunDateTime : "''") + ", ''), 'z'))" + " "
```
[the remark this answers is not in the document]
mattabullock: `NULLIF` is necessary because `nextRunDateTime` comes out as empty string (not null) and that is always the `MIN`, so we need it for the COALESCE to work correctly.

mattabullock:
updated

test/tests/jobs/CreateJobTest.cpp
```diff
@@ -266,11 +266,75 @@ struct CreateJobTest : tpunit::TestFixture {
...
+        command["repeat"] = repeatValue;
+        command["retryAfter"] = retryValue;
+
+        STable response = tester->executeWaitVerifyContentTable(command);
+        string jobID = response["jobID"];
+
+        SQResult originalJob;
+        tester->readDB("SELECT created, jobID, state, name, nextRun, lastRun, repeat, data, priority, parentJobID, retryAfter FROM jobs WHERE jobID = " + jobID + ";", originalJob);
+
+        ASSERT_EQUAL(originalJob[0][1], jobID);
```
iwiznia: NAB: I don't think these asserts are needed

test/tests/jobs/CreateJobTest.cpp
```diff
@@ -266,11 +266,75 @@ struct CreateJobTest : tpunit::TestFixture {
...
+        // Query the db and confirm that state, nextRun and lastRun are 1 second apart because of retryAfter
+        SQResult jobData;
+        tester->readDB("SELECT state, nextRun, lastRun FROM jobs WHERE jobID = " + jobID + ";", jobData);
+        ASSERT_EQUAL(jobData[0][0], "RUNQUEUED");
+        time_t nextRunTime = JobTestHelper::getTimestampForDateTimeString(jobData[0][1]);
+        time_t lastRunTime = JobTestHelper::getTimestampForDateTimeString(jobData[0][2]);
+        ASSERT_EQUAL(difftime(nextRunTime, lastRunTime), 1);
+
+        // Get the job, confirm error because 1 second hasn't passed
+        tester->executeWaitVerifyContent(command, "404 No job found");
```
iwiznia: This has flaky test written all over it...

PR #531, opened 2018-11-12
Return retryAfter in queryJob command

iwiznia:
@pecanoro please review.
 
Related to https://github.com/Expensify/Expensify/issues/90258
Used in https://github.com/Expensify/Web-Expensify/pull/22982

Tests:
Called queryJob and check retryAfter was returned.


pecanoro:
Tests failed, but also you can probably add a new test to QueryJob to make sure we return all parameters.

iwiznia:
Updated with the new tests, I have no idea why the build was failing, hope it passes now.

PR #550, opened 2018-12-04
Throw when trying to delete parent with children

pecanoro:
This can actually happen when one of the children fails and honestly, we don't want to just warn and move o, the parent should fail as well.

## Issues 

https://github.com/Expensify/Expensify/issues/94073

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -1059,7 +1059,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                 // At this point, all child jobs should already be deleted, but
                 // let's double check.
                 if (!db.read("SELECT 1 FROM jobs WHERE parentJobID != 0 AND parentJobID=" + SQ(jobID) + " LIMIT 1;").empty()) {
-                    SWARN("Child jobs still exist when deleting parent job, ignoring.");
+                    STHROW("405 Failed to delete a job with outstanding children");
                 }
             }
         }


iwiznia:
I am not sure I get this... when this happens exactly?

pecanoro:
- One of the children fails.
- Another children runs and finish, since there are no more QUEUED children, it resumes the parent.
- Parent resumes and finishes because it does not check for FAILED children,  it just deletes FINISHED or CANCELLED children. Leaving the FAILED one behind.

iwiznia:
hmmmm, but then shouldn't be the fix to delete the FAILED children? I think I still don't get it. What should happen to the parent when a child fails?

pecanoro:
The option that makes more sense to me is to FAIL as well. We can gather more opinions cc/ @cead22 @fnwbr

cead22:
With this change, will the parent job get stuck?

pecanoro:
Yes, it fails.

cead22:
I think that makes sense since we'll need to unstick the child job in order to unstick the parent job, right?

Also, does this happen mostly for SSDos or are the other scenarios where we see this?

pecanoro:
But FAILED means we have to investigate, otherwise, you may have been creating more bugs but letting the parent resumed without having completed children.

> Also, does this happen mostly for SSDos or are the other scenarios where we see this?

It has never happened before, only during the fire, but by not failing the parent, it just went ahead and finished receipts with incomplete data.

cead22:
> But FAILED means we have to investigate, otherwise, you may have been creating more bugs but letting the parent resumed without having completed children.

👍 It sounds like we're on the same but let me know if I'm misunderstanding

pecanoro:
Yeah, yeah, we agree, I was just adding more context

fnwbr:
Agreed 👍 	- @iwiznia any concerns that you'd see with Rocio's proposal to also mark the parent as `FAILED`?

iwiznia:
> any concerns that you'd see with Rocio's proposal to also mark the parent as FAILED?

I think this is better yes. As it is right now the parent job would remain in RUNNING right? And if it has a retryAfter it would even be run again...?

pecanoro:
RIght now the parent will simply finish leaving the FAILED child behind.

iwiznia:
Yeah, then I like more the approach of leaving the parent in FAILED state too, same as the children.

pecanoro:
So are you going to merge it then?

iwiznia:
Ummm maybe I am misunderstanding something, but this PR does not leave the parent in FAILED state, does it? Or is it because after throwing here, BWM will mark the parent as FAILED?

iwiznia:
>  Or is it because after throwing here, BWM will mark the parent as FAILED?

yes

PR #572, opened 2019-01-03
Fix/c++14

TomMD:
These changes are just a bit of standardization:

* Move from gnu++14 to c++14
* Use cxa_demangle in a manner consistent with the spec.
* Call gcc and g++ instead of an explicit version of the compiler


quinthar:
Awesome, we'll take a look, thank you!!

PR #567, opened 2018-12-21
Log request email

iwiznia:
@tylerkaraszewski please review. /cc @flodnv 

# Fixes 
https://github.com/Expensify/Expensify/issues/93741

# Tests
Checked the logs:
Auth request `Auth::call('SetNameValuePair', ['name' => 'caca', 'value' => 'caca', 'requestEmail' => 'ionatan@expensify.com', 'authToken' => authToken()])`:
```
2018-12-21T14:25:13.779799+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (BedrockServer.cpp:1452) postPoll [main] [info] Waiting for 'SetNameValuePair' to complete.
2018-12-21T14:25:13.780233+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (BedrockServer.cpp:1491) postPoll [main] [info] Queued new 'SetNameValuePair' command from local client, with 0 commands already queued.
2018-12-21T14:25:13.781034+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (BedrockServer.cpp:686) worker [worker2] [info] Dequeued command SetNameValuePair in worker, 0 commands in  queue.
2018-12-21T14:25:13.781317+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (BedrockCore.cpp:65) peekCommand [worker2] [dbug] Peeking at 'SetNameValuePair' with priority: 500
2018-12-21T14:25:13.781507+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (SQLite.cpp:417) beginConcurrentTransaction [worker2] [dbug] [concurrent] Beginning transaction
2018-12-21T14:25:13.782253+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (libstuff.cpp:2355) SQuery [worker2] [dbug] BEGIN CONCURRENT
2018-12-21T14:25:13.782574+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (libstuff.cpp:2355) SQuery [worker2] [dbug] PRAGMA query_only = true;
2018-12-21T14:25:13.783439+00:00 expensidev bedrock: xK6shX ionatan@expensify.com (Auth.cpp:658) peekCommand [worker2] [dbug] {:Auth} Peeking at 'SetNameValuePair'
[...]

PR #591, opened 2019-01-18
Revert "Log request email"

coleaeason:
Reverts Expensify/Bedrock#567

@righdforsa I can't think of any reason this would be drastically slower, but it's the only change we haven't rolled back and tested with.


iwiznia:
Spent a long time trying to figure out why I wasn't seeing any emails in requests... I assume we can turn this back on.

PR #599, opened 2019-01-22
Revert "Revert "Log request email""

iwiznia:
Reverts Expensify/Bedrock#591

I assume that was a test and that we can re-enable this.


coleaeason:
Yeah this was a test, and yes we can re-enable it.

PR #607, opened 2019-02-01
Have any job that finishes with delete:true in their data object delete itself

yuwenmemon:
@tylerkaraszewski @iwiznia please review

Coming from https://github.com/Expensify/Web-Expensify/pull/23921#discussion_r253056529, we want the ability for a recurring job that has `"delete": true` in its data object to be able to delete itself.

Tests:
- I did the test steps in https://github.com/Expensify/Web-Expensify/pull/23921#issue-249411586, and made sure the same results happened with this change as opposed to the `UPDATE` query being in that PR.


yuwenmemon:
Added!

PR #613, opened 2019-03-06
Update Expensify branding for BedrockDB.com

shawnborton:
@coleaeason will you please review? This adds new logos/fonts/colors for the BedrockDB.com website.

image

I just ran `jekyll serve` to test these changes locally, and it looks like I ended up with lots of `.html` changes. Let me know if those shouldn't be committed or not - admittedly I'm not too familiar with how this works.


shawnborton:
Cool, thanks for the guidance. Updated!

PR #614, opened 2019-03-06
Exclude RUNQUEUED from fail job prevention list

ctkochan22:
@tylerkaraszewski @flodnv @iwiznia 
cc @pecanoro @chiragsalian 

Right now, any job that is created with a `retryAfter` set, will never be set to `RUNNING`, it will always be `RUNQUEUED`. However, the problem with this is that we only fail jobs that are in the Running state. 

This means, if a job that throws an error and we try to run `FailJob` on it, it will fail to do so. It will not set the state to `FAILED` and it will remain as `RUNQUEUED`, which means that if job is failing for a good reason, we will continue to pick it up and run it forever. 

So I'm arguing that we include `RUNQUEUED` along with `RUNNING` as state where we allow it to fail.

### Fixed Issue:
$ https://github.com/Expensify/Expensify/issues/100261

# Tests
- Pick any job, and make sure it fails. 
    e.g. I picked the job in a PR i'm working on. I manually break it by passing in a messageID (for intercom) that doesn't exist so that it never successfully pulls a conversation. And will retry 5 times before throwing an ExpError. 
- From `nc auth 8888` create a job with a retryAfter value
e.g. 
```
CreateJob
name: www-prod/IntercomCreateConversation
data: {"email":"ctseng@expensify.com", "adminID":"521022", "subject":"test", "messageType":"email", "message":"test test", "assigneeID":"1079954"}
retryAfter: +1 MINUTES
```
- With this change, the job should successfully fail once teh ExpError is thrown

You can also simply test it like so
image

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -1139,9 +1139,9 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         }
         const string& state = result[0][0];
 
-        // Make sure we're failing a job that's actually running
-        if (state != "RUNNING") {
-            SINFO("Trying to fail job#" << request["jobID"] << ", but isn't RUNNING (" << state << ")");
+        // Make sure we're failing a job that's actually running or running with a retryAfter
+        if (state != "RUNNING" && state != "RUNQUEUED") {
+            SINFO("Trying to fail job#" << request["jobID"] << ", but isn't RUNNING or RUNQUEUED (" << state << ")");
             STHROW("405 Can only fail RUNNING jobs");
         }
 
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -1142,7 +1142,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         // Make sure we're failing a job that's actually running or running with a retryAfter
         if (state != "RUNNING" && state != "RUNQUEUED") {
             SINFO("Trying to fail job#" << request["jobID"] << ", but isn't RUNNING or RUNQUEUED (" << state << ")");
-            STHROW("405 Can only fail RUNNING jobs");
+            STHROW("405 Can only fail RUNNING or RUNQUEUED jobs");
         }
 
         // Are we updating the data too?
--- a/test/tests/jobs/FailJobTest.cpp
+++ b/test/tests/jobs/FailJobTest.cpp
@@ -47,6 +47,7 @@ struct FailJobTest : tpunit::TestFixture {
         tester->executeWaitVerifyContent(command, "405 Can only fail RUNNING or RUNQUEUED jobs");
     }
 
+    // Fail job in RUNNING state
     void failJobInRunningState() {
         // Create a job
         SData command("CreateJob");
@@ -72,6 +73,7 @@ struct FailJobTest : tpunit::TestFixture {
         ASSERT_EQUAL(result[0][0], "FAILED");
     }
 
+    // Fail job in RUNQUEUED state
     void failJobInRunqueuedState() {
         // Create a job
         SData command("CreateJob");
--- a/test/tests/jobs/FailJobTest.cpp
+++ b/test/tests/jobs/FailJobTest.cpp
@@ -40,7 +40,7 @@ struct FailJobTest : tpunit::TestFixture {
         STable response = tester->executeWaitVerifyContentTable(command);
         string jobID = response["jobID"];
 
-        // Finish it
+        // Fail it
         command.clear();
         command.methodLine = "FailJob";
         command["jobID"] = jobID;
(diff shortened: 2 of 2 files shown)


plugins/Jobs.cpp
```diff
@@ -1139,9 +1139,9 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         }
         const string& state = result[0][0];
 
-        // Make sure we're failing a job that's actually running
-        if (state != "RUNNING") {
-            SINFO("Trying to fail job#" << request["jobID"] << ", but isn't RUNNING (" << state << ")");
+        // Make sure we're failing a job that's actually running or running with a retryAfter
+        if (state != "RUNNING" && state != "RUNQUEUED") {
+            SINFO("Trying to fail job#" << request["jobID"] << ", but isn't RUNNING or RUNQUEUED (" << state << ")");
             STHROW("405 Can only fail RUNNING jobs");
```
pecanoro: Can you update the exception? `RUNNING OR RUNQUEUED`

ctkochan22: Updated :) resolving

ctkochan22:
updated :D

chiragsalian:
P.S There's an easier way to test this than creating an actual job and having it throw experror and running bwm.sh

You could just do all of the necessary testing in netcat, 
On master,
image

In your branch,
image

plugins/Jobs.cpp
```diff
@@ -1139,10 +1139,10 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         }
         const string& state = result[0][0];
 
-        // Make sure we're failing a job that's actually running
-        if (state != "RUNNING") {
-            SINFO("Trying to fail job#" << request["jobID"] << ", but isn't RUNNING (" << state << ")");
-            STHROW("405 Can only fail RUNNING jobs");
+        // Make sure we're failing a job that's actually running or running with a retryAfter
+        if (state != "RUNNING" && state != "RUNQUEUED") {
```
chiragsalian: @flodnv shouldn't we check timestamp here to identify if its a currently active RUNQUEUED job to ensure we don't fail a job that RUNQUEUED but not active?

pecanoro: My opinion is that it's fine not to check it since we don't do it for `CancelJob` either.

ctkochan22:
Added tests, but I see in other test files that we test parent and child stuff. Should I add those as well?

plugins/Jobs.cpp
flodnv: @chiragsalian What's a non-active RUNQUEUED job? RUNQUEUED is effectively the same as RUNNING, except with more features.

flodnv:
Not needed @ctkochan22 , you're not really touching that.

flodnv:
>You could just do all of the necessary testing in netcat,

@chiragsalian much better to do it via automated tests, this way they are ran for you forever on every and each build :)

test/tests/jobs/FailJobTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    // Throw an error if the job is not in RUNNING or REQUEUED state
+    void notInRunningRunqueuedState() {
+        // Create a job
+        SData command("CreateJob");
+        command["name"] = "job";
+        STable response = tester->executeWaitVerifyContentTable(command);
+        string jobID = response["jobID"];
+
+        // Finish it
```
flodnv: bad commment

test/tests/jobs/FailJobTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+        // Create a job
+        SData command("CreateJob");
+        command["name"] = "job";
+        command["retryAfter"] = "+1 MINUTES";
+        STable response = tester->executeWaitVerifyContentTable(command);
+        string jobID = response["jobID"];
+
+        command.clear();
+        command.methodLine = "Query";
+        command["query"] = "UPDATE jobs SET state = 'RUNNING' WHERE jobID = " + jobID + ";";
```
flodnv: This is not a very real test. Instead, run a GetJob to grab this job and make the plugin switch its state to RUNNING. Heh - just like you did below 😄

plugins/Jobs.cpp
pecanoro: He means that a RUNQUEUED job may be RUNNING or it may be QUEUED. Therefore, I think he is asking if it's ok to fail a RUNQUEUED job that may be just QUEUED waiting to be run, since we don't know if it's actually running or not.

chiragsalian: Yup what rocio said, RUNQUEUED is basically a job determined as RUNNING or QUEUED depending on its nextRun time. Hence the question being is it alright that we can fail a RUNQUEUED job thats effectively the same state as QUEUED i.e., a job that can be retried via getJob.

ctkochan22:
Updated the tests :)

plugins/Jobs.cpp
flodnv: Oh ok I see what you mean. 

> is it alright that we can fail a RUNQUEUED job

Yes, it is already past the "raw" QUEUED state, it is running, if you want to fail it, you can ; just like you can finish it or cancel it. If you don't do anything, it'll run again, and again, and again. In some situations you don't want to finish nor cancel it, so we need to allow for failing it.

PR #620, opened 2019-03-22
489 create file during clean

qpfiffer:
Makes `-clean` leave a blank file after it has run. I saw this in the `README` and figured it'd be a good excuse to check out some of the code. Lemme know if this PR doesn't make sense, for example should Bedrock always create a blank file? 

Fixes 489, 24.


qpfiffer:
> Dose this duplicate just calling `SFileSave` with an empty content buffer? I'm not sure we need `SFileCreate` if we already have overlapping functionality.

It's slightly different in that it uses `a+` instead of `wb` as an argument to `fopen()`; I didn't want the file to be discarded if it exists. I'm not really partial one way or the other, it just seemed presumptive to clear any file that is passed into the commandline.

PR #631, opened 2019-05-14
Add new SSECURITY log line

mattabullock:
We're going to begin tagging security related log lines in rsyslog, so we need them to follow a specific format each time.

# Test
With https://github.com/Expensify/Server-Expensify/pull/3479 and 
1. Log into account as support
2. Using ES, see if event.tags is populated as security.


mattabullock:
That's what I was thinking as I was making the PR... all our CPP logging tools are here though. Any suggestions?

coleaeason:
My suggestion is to keep this in the private repos, even if that means creating some new library helpers, which should be trivial.

mattabullock:
Okay I'll hack something up.

PR #643, opened 2019-05-17
Support for RequeueJobs to update the job name if provided

chiragsalian:
@tylerkaraszewski or @iwiznia or @cead22 , could you review this?
cc @fnwbr 

~**WIP** - i'll add tests~ (Done)

As per the discussion [here](https://github.com/Expensify/Expensify/issues/103633#issuecomment-493415723), I'd like to update RequeueJobs to support updating job name as well.

Sample nc command,
```
RequeueJobs
jobIDs: 5307009005176589558, 5721840509781560221, 83435286685058090 
name: manual/ChatbotUnprocessedInput?queue=c4b
```

# Tests
Automated

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -1209,6 +1209,14 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
             if (!db.writeIdempotent(updateQuery)) {
                 STHROW("502 RequeueJobs update failed");
             }
+
+            // If we want to update the name, let's do that
+            const string& name = request["name"];
+            if (!name.empty()) {
+                if (!db.writeIdempotent("UPDATE jobs SET name = " + SQ(name) + " WHERE jobID IN(" + SQList(jobIDs)+ ");")) {
+                    STHROW("502 Failed to update job name");
+                }
+            }
         }
 
         return true;
--- a/test/tests/jobs/RequeueJobsTest.cpp
+++ b/test/tests/jobs/RequeueJobsTest.cpp
@@ -1,0 +1,198 @@
+#include <test/lib/BedrockTester.h>
+#include <test/tests/jobs/JobTestHelper.h>
+
+struct RequeueJobsTest : tpunit::TestFixture {
+    RequeueJobsTest()
+        : tpunit::TestFixture("RequeueJobs",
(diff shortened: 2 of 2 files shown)


chiragsalian:
Added automated tests. Ready for review.

plugins/Jobs.cpp
```diff
@@ -1209,6 +1209,14 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
             if (!db.writeIdempotent(updateQuery)) {
                 STHROW("502 RequeueJobs update failed");
             }
+
+            // If we want to update the name, let's do that
+            const string& name = request["name"];
```
iwiznia: Why not try to make this update in the query above instead of doing 2 separate queries?

fnwbr: IMO what Ionatan is suggesting should be a blocker before we start making use of this new option. Or are there any downsides to doing it in the same query?

chiragsalian: So for this I tried to follow how the structure was written elsewhere in this file and it seems like we run multiple update commands depending if the param was passed. Another example of how name was updated is [here](https://github.com/Expensify/Bedrock/blob/e857107be743af28490fc109af6d2144223b6939/plugins/Jobs.cpp#L994). 

Do you feel like having it combined is definitely better? If so i can update this. Not sure why it would be a blocker though but whatever you guys feel.

iwiznia: I don't think this is a blocker, but will let Florian decide that.
As a general rule, we should always try to make commands do the least amount of queries we can... I have no idea why we did 2 in that example you pasted, maybe an oversight or maybe it made the code harder to read/understand (as it is only applying to one command), not sure really.

chiragsalian: Cool sounds good. I'll combine the queries and keep in mind that for future we'd prefer fewer queries over here :+1:

chiragsalian: @fnwbr mind if i update the code in a follow up PR so we don't block the Web-E PR on this.

fnwbr: Heh, well, my point would be do update this _before_ we start making use of this.

But I can't judge about the performance. Maybe the current average load is so small that we wouldn't even notice if it's 2 queries or only 1.

I'd say: Let's not hold Web, but prioritize getting the follow-up Bedrock PR out there.

PR #665, opened 2019-09-26
Log normalized slow query

migfabio:
@tylerkaraszewski @iwiznia @flodnv 

This PR will make Bedrock logging the normalised slow queries

# Test
1) Manually change the threshold to 1ms
2) `makeAll`
3) Check the log for Slow query

### DIFF
--- a/libstuff/libstuff.h
+++ b/libstuff/libstuff.h
@@ -772,11 +772,10 @@ void SQueryLogOpen(const string& logFilename);
 void SQueryLogClose();
 
 // Returns an SQLite result code.
-int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result,
-           int64_t warnThreshold = 2000 * STIME_US_PER_MS, bool skipWarn = false);
-inline int SQuery(sqlite3* db, const char* e, const string& sql, int64_t warnThreshold = 2000 * STIME_US_PER_MS, bool skipWarn = false) {
+int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result, bool skipWarn = false);
+inline int SQuery(sqlite3* db, const char* e, const string& sql, bool skipWarn = false) {
     SQResult ignore;
-    return SQuery(db, e, sql, ignore, warnThreshold, skipWarn);
+    return SQuery(db, e, sql, ignore, skipWarn);
 }
 
 bool SQVerifyTable(sqlite3* db, const string& tableName, const string& sql);
--- a/libstuff/libstuff.h
+++ b/libstuff/libstuff.h
@@ -773,10 +773,7 @@ void SQueryLogClose();
 
 // Returns an SQLite result code.
 int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result, bool skipWarn = false);
-inline int SQuery(sqlite3* db, const char* e, const string& sql, bool skipWarn = false) {
-    SQResult ignore;
-    return SQuery(db, e, sql, ignore, skipWarn);
-}
+int SQuery(sqlite3* db, const char* e, const string& sql, bool skipWarn = false);
 
 bool SQVerifyTable(sqlite3* db, const string& tableName, const string& sql);
 bool SQVerifyTableExists(sqlite3* db, const string& tableName);
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -116,25 +116,25 @@ SQLite::SQLite(const string& filename, int cacheSize, bool enableFullCheckpoints
     SASSERT(!sqlite3_open_v2(filename.c_str(), &_db, DB_WRITE_OPEN_FLAGS, NULL));
 
     // WAL is what allows simultaneous read/writing.
-    SASSERT(!SQuery(_db, "enabling write ahead logging", "PRAGMA journal_mode = WAL;"));
+    SASSERT(!timedSQuery("enabling write ahead logging", "PRAGMA journal_mode = WAL;"));
 
     if (mmapSizeGB) {
-        SASSERT(!SQuery(_db, "enabling memory-mapped I/O", "PRAGMA mmap_size=" + to_string(mmapSizeGB * 1024 * 1024 * 1024) + ";"));
+        SASSERT(!timedSQuery("enabling memory-mapped I/O", "PRAGMA mmap_size=" + to_string(mmapSizeGB * 1024 * 1024 * 1024) + ";"));
     }
 
     // PRAGMA legacy_file_format=OFF sets the default for creating new databases, so it must be called before creating
     // any tables to be effective.
-    SASSERT(!SQuery(_db, "new file format for DESC indexes", "PRAGMA legacy_file_format = OFF"));
+    SASSERT(!timedSQuery("new file format for DESC indexes", "PRAGMA legacy_file_format = OFF"));
 
     // Check if synchronous has been set and run query to use a custom synchronous setting
     if (!synchronous.empty()) {
-        SASSERT(!SQuery(_db, "setting custom synchronous commits", "PRAGMA synchronous = " + SQ(synchronous)  + ";"));
+        SASSERT(!timedSQuery("setting custom synchronous commits", "PRAGMA synchronous = " + SQ(synchronous)  + ";"));
     } else {
         DBINFO("Using SQLite default PRAGMA synchronous");
     }
 
     // These other pragmas only relate to read/write databases.
-    SASSERT(!SQuery(_db, "disabling change counting", "PRAGMA count_changes = OFF;"));
+    SASSERT(!timedSQuery("disabling change counting", "PRAGMA count_changes = OFF;"));
 
     // Do our own checkpointing.
     sqlite3_wal_hook(_db, _sqliteWALCallback, this);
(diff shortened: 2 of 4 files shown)


flodnv:
Please get in the habit of using pull panda 🙇

flodnv:
Seems like we forgot to configure it to be enabled on this repo, can you do it please? 😄

migfabio:
@flodnv Oh I will, I just need the first round of review because I didn't put any test

sqlitecluster/SQLite.cpp
```diff
@@ -18,6 +18,8 @@ atomic<int> SQLite::fullCheckpointPageMin(25000); // Approx 100mb (pages are ass
 // Tracing can only be enabled or disabled globally, not per object.
 atomic<bool> SQLite::enableTrace(false);
 
+atomic<sqlite3_stmt*> SQLite::_lastQuery;
```
tylerkaraszewski: I think this should be a string, and we should just store `sqlite3_normalized_sql((sqlite3_stmt*)p)` in `_sqliteTraceCallback`. Based on the docs ([here](https://www.sqlite.org/c3ref/exec.html) and [here](https://www.sqlite.org/c3ref/stmt.html)), this object is destroyed when `sqlite3_exec` inside `SQuery`, and so we shouldn't be able to access it after `SQuery` returns.

Also, it shouldn't need to be atomic, only one thread has access to this object.

sqlitecluster/SQLite.cpp
```diff
@@ -405,7 +409,7 @@ bool SQLite::beginTransaction(bool useCache, const string& transactionName) {
     _sharedData->blockNewTransactionsCV.notify_one();
     SDEBUG("Beginning transaction");
     uint64_t before = STimeNow();
-    _insideTransaction = !SQuery(_db, "starting db transaction", "BEGIN TRANSACTION");
+    _insideTransaction = !timedSQuery("starting db transaction", "BEGIN TRANSACTION");
```
tylerkaraszewski: I'd skip the timing on these "standard" queries that we run before/after other queries, and on startup/shutdown. I don't think we need to time these. Only "user-supplied" queries should need timing.

sqlitecluster/SQLite.cpp
```diff
@@ -1067,6 +1071,15 @@ bool SQLite::getUpdateNoopMode() const {
     return _noopUpdateMode;
 }
 
+int SQLite::timedSQuery(const char *e, const string &sql, SQResult &result, int64_t warnThreshold, bool skipWarn) {
+    uint64_t startTime = STimeNow();
+    int queryResult = SQuery(_db, e, sql, result, skipWarn);
+    uint64_t elapsed = STimeNow() - startTime;
```
tylerkaraszewski: I'm going to suggest using the new std::chrono functions like we use here: https://github.com/Expensify/Bedrock/blob/master/main.cpp#L315-L350

But if you like it this way feel free to ignore me.

sqlitecluster/SQLite.cpp
```diff
@@ -1067,6 +1071,15 @@ bool SQLite::getUpdateNoopMode() const {
...
+    if ((int64_t)elapsed > warnThreshold)
+        SWARN("Slow query (" << elapsed / 1000 << "ms) " << sql.length() << ": " << sqlite3_normalized_sql(_lastQuery));
```
tylerkaraszewski: `length()` in here doesn't seem to tell us much.

sqlitecluster/SQLite.cpp
```diff
@@ -1067,6 +1071,15 @@ bool SQLite::getUpdateNoopMode() const {
     return _noopUpdateMode;
 }
 
+int SQLite::timedSQuery(const char *e, const string &sql, SQResult &result, int64_t warnThreshold, bool skipWarn) {
```
tylerkaraszewski: Let's just call it `timedQuery`, I think. It's not in libstuff.

sqlitecluster/SQLite.h
```diff
@@ -192,6 +198,8 @@ class SQLite {
     // Enable/disable SQL statement tracing.
     static atomic<bool> enableTrace;
 
+    static atomic<sqlite3_stmt*> _lastQuery;
```
tylerkaraszewski: Like above, this should probably be a string and doesn't need to be atomic, but specifically it needs to *not* be `static`. Otherwise there's no way for each `SQLite` object to know it's looking at the right value.

migfabio: mh, the only way to get a `sqlite3_stmt` is in `_sqliteTraceCallback` which is static. Is there another way to get it?

tylerkaraszewski: We register the callback function here: https://github.com/Expensify/Bedrock/blob/master/sqlitecluster/SQLite.cpp#L141

You can see the fourth argument to `sqlite3_trace_v2` is `this`, a pointer to the current SQLite object.

So, based on the (somewhat confusing) docs here: https://www.sqlite.org/c3ref/trace_v2.html the `c` argument to `_sqliteTraceCallback` here: https://github.com/Expensify/Bedrock/blob/master/sqlitecluster/SQLite.cpp#L233 is the same as `this`, so you can do:

```
c->_lastQuery
```

To access the last query for this particular object.

migfabio:
Thanks again @tylerkaraszewski 
I think I've addressed your comments. Can you double-check?

libstuff/libstuff.h
```diff
@@ -772,11 +772,10 @@ void SQueryLogOpen(const string& logFilename);
 void SQueryLogClose();
 
 // Returns an SQLite result code.
-int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result,
-           int64_t warnThreshold = 2000 * STIME_US_PER_MS, bool skipWarn = false);
-inline int SQuery(sqlite3* db, const char* e, const string& sql, int64_t warnThreshold = 2000 * STIME_US_PER_MS, bool skipWarn = false) {
+int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result, bool skipWarn = false);
+inline int SQuery(sqlite3* db, const char* e, const string& sql, bool skipWarn = false) {
```
tylerkaraszewski: Actually, one small change, drop the `inline` here, we've stopped using it and let the compiler decide on it's own optimizations. I'd actually put the implementation in `SQLite.cpp`, too, but that's minor.

migfabio:
> This looks really good! Only thing I'd like to see is some example log lines to indicate it's working! I don't think we need automated tests for this, since we don't have a good way to check logs in the tests, so if you could just post some in the issue indicating that this is working, that'd be good.

@tylerkaraszewski gonna follow what @iwiznia said [here](https://github.com/Expensify/Expensify/issues/94508#issuecomment-534503237) if that's okay

migfabio:
Set the threshold to 1ms

Screenshot 2019-09-27 at 14 54 45

libstuff/libstuff.h
migfabio: I don't know why, but I get errors if I drop the `inline`
maybe I should do something else, like changing all the SQuery calls from SQLite?

tylerkaraszewski: I would move the method definition into the cpp file, so you have only:
```
int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result, bool skipWarn = false);
int SQuery(sqlite3* db, const char* e, const string& sql, bool skipWarn = false);
```

In SQLite.h, and the definitions for both functions on SQLite.cpp.

tylerkaraszewski:
This looks good to me. I'm fine to merge.

migfabio:
Removed WIP and added tests if you like to do test this.
Waiting for Travis too

sqlitecluster/SQLite.h
```diff
@@ -192,6 +197,8 @@ class SQLite {
     // Enable/disable SQL statement tracing.
     static atomic<bool> enableTrace;
 
+    string _lastQuery;
```
flodnv: does the underscore indicate this should be private? in any case, we prefer not prefixing with underscore for privates (that's an old thing -- @tylerkaraszewski correct me if I'm wrong)

tylerkaraszewski: We don't use underscores for private variables in auth but we still do in bedrock.

migfabio: is it not? my bad... gonna fix it

migfabio:
Travis failed. let me check the test

flodnv:
I think we are stuck in this weird test where travis passes but fails? Does that ring a bell @tylerkaraszewski ?

tylerkaraszewski:
I'm just trying it again.

migfabio:
travis is happy now!

tylerkaraszewski:
This was reverted for performance reasons, see discussion here:
https://expensify.slack.com/archives/C094TGUTZ/p1569889671057900

migfabio:
bummer, but I need some clarification

@coleaeason said
> pretty sure there’s some serious overhead to normalizing every single query we run…

In this PR we just normalize the Slow query and we normalize all queries `enableTrace` is `true` which existed before this PR

so, are we sure it is PR? if yes, do you have any idea of how to fix this?

coleaeason:
>In this PR we just normalize the Slow query and we normalize all queries if enableTrace is true which existed before this PR

Hm this is not true, You're normalizing all queries, then only printing them if `enableTrace` is `true`:

https://github.com/Expensify/Bedrock/pull/665/files#diff-e0fd399d80d036c72780c20678a97b54R237

migfabio:
true, my bad.
then I need to store the raw sqlite query and normalize when it needs. do you see some issue doing that?
gonna make this change soon(tm) in case

coleaeason:
That should be fine as far as I can tell. Although, i'm not sure i actually like this change. Seeing the normalized slow queries was way less helpful than seeing slowing full queries, IMO.

migfabio:
gotcha
any opinion from the issue creator about it?
cc @iwiznia @flodnv

iwiznia:
Well, if the query is long we were truncating them, so you don't see the full query, but also I don't see why it is less useful, can you clarify?

PR #676, opened 2019-11-11
Add orderByCreated parameter to GetJob

alex-mechler:
@mattabullock  

Related issue: https://github.com/Expensify/Expensify/issues/118854

Context for this solution: https://github.com/Expensify/Expensify/issues/118854#issuecomment-550572350

# Tests
Automated

Can be manually tested with these PRs:
https://github.com/Expensify/Web-Expensify/pull/26251
https://github.com/Expensify/Bedrock-PHP/pull/142

# Query Timings/Plans
No new query, just adding the option to change an `ORDER BY` to be based on created, not just nextRun

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -661,6 +661,13 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         const list<string> nameList = SParseList(request["name"]);
         string safeNumResults = SQ(max(request.calc("numResults"),1));
         bool mockRequest = command.request.isSet("mockRequest") || command.request.isSet("getMockedJobs");
+        bool orderByCreated = command.request.isSet("orderByCreated") ? request.test("orderByCreated") : false;
+        string orderByQuery = "ORDER BY ";
+        if (orderByCreated) {
+            orderByQuery += "created ASC";
+        } else {
+            orderByQuery += "nextRun ASC";
+        }
         string selectQuery;
         if (request.isSet("jobPriority")) {
             selectQuery =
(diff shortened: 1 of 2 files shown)


plugins/Jobs.cpp
```diff
@@ -661,6 +661,13 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         const list<string> nameList = SParseList(request["name"]);
         string safeNumResults = SQ(max(request.calc("numResults"),1));
         bool mockRequest = command.request.isSet("mockRequest") || command.request.isSet("getMockedJobs");
+        bool orderByCreated = command.request.isSet("orderByCreated") ? request.test("orderByCreated") : false;
```
mattabullock: request.test will handle checking if it's set and setting it to false I believe, so the ternary is redundant

alex-mechler:
Closing this because we have a better way forward

plugins/Jobs.cpp
```diff
@@ -661,6 +662,13 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         const list<string> nameList = SParseList(request["name"]);
         string safeNumResults = SQ(max(request.calc("numResults"),1));
         bool mockRequest = command.request.isSet("mockRequest") || command.request.isSet("getMockedJobs");
+        bool orderByCreated = request.test("orderByCreated");
+        string orderByQuery = "ORDER BY ";
+        if (orderByCreated) {
```
chiragsalian: lets,
```
orderByQuery += orderByCreated ? "created ASC" : "nextRun ASC"
```

chiragsalian: actually why not,
```
safeOrderBy = orderByCreated ? "created ASC" : "nextRun ASC"
```
and then  in the query you would,
```
"ORDER BY " + safeOrderBy + " LIMIT " + safeNumResults + ";";
```

PR #691, opened 2020-01-29
Add jobPriority parameter to UpdateJob

alex-mechler:
@mattabullock can you review this please?

cc: @coleaeason @stitesExpensify in case something would be wrong with this

Needed for https://github.com/Expensify/Expensify/issues/123018

Simple update to the `UpdateJob` command to also allow it to update the jobs priority.

Can be tested with `nc` by having a job in the db, then running

```
UpdateJob
jobID:<the existing jobID>
data:<the existing jobData>
priority:<a different priority>
```

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -842,6 +842,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         //     - jobID - ID of the job to delete
         //     - data  - A JSON object describing work to be done
         //     - repeat - A description of how to repeat (optional)
+        //     - priority - The priority of the job (optional)
         //
         verifyAttributeInt64(request, "jobID", 1);
         verifyAttributeSize(request, "data", 1, MAX_SIZE_BLOB);
@@ -854,6 +855,14 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                 request.erase("repeat");
             } else if (!_validateRepeat(request["repeat"])) {
                 STHROW("402 Malformed repeat");
+            }
+        }
+
+        // If a priority is provided, validate it
+        if (request.isSet("priority")) {
+            int64_t priority = request.calc64("priority");
+            if (priority != 0 && priority != 500 && priority != 1000) {
+                STHROW("402 Invalid priority value");
             }
         }
 
@@ -881,6 +890,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                                 SQ(request["data"]) + " " +
                                 (request.isSet("repeat") ? ", repeat=" + SQ(SToUpper(request["repeat"])) : "") +
                                 (!newNextRun.empty() ? ", nextRun=" + newNextRun : "") +
+                                (request.isSet("priority") ? ", priority=" + request.calc64("priority") : "") +
                                 "WHERE jobID=" +
                                 SQ(request.calc64("jobID")) + ";")) {
             STHROW("502 Update failed");
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -842,7 +842,7 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
         //     - jobID - ID of the job to delete
         //     - data  - A JSON object describing work to be done
         //     - repeat - A description of how to repeat (optional)
-        //     - priority - The priority of the job (optional)
+        //     - jobPriority - The priority of the job (optional)
         //
         verifyAttributeInt64(request, "jobID", 1);
         verifyAttributeSize(request, "data", 1, MAX_SIZE_BLOB);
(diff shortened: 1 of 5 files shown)


alex-mechler:
Added in a basic test file that tests the core functionality of UpdateJob.

test/tests/jobs/UpdateJobTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+                                  TEST(UpdateJobTest::updateJob),
+                                  AFTER(UpdateJobTest::tearDown),
+                                  AFTER_CLASS(UpdateJobTest::tearDownClass)) { }
+
+    BedrockTester* tester;
+
+    void setupClass() { tester = new BedrockTester(_threadID, {{"-plugins", "Jobs,DB"}}, {});}
+
+    // Reset the jobs table
+    void tearDown() {
```
coleaeason: you shouldn't need this, every tester gets it's own db, etc. so when this tested is deleted in "tearDownClass()` all of the data is deleted.

test/tests/jobs/UpdateJobTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+    UpdateJobTest()
+            : tpunit::TestFixture("UpdateJob",
+                                  BEFORE_CLASS(UpdateJobTest::setupClass),
+                                  TEST(UpdateJobTest::updateJob),
+                                  AFTER(UpdateJobTest::tearDown),
+                                  AFTER_CLASS(UpdateJobTest::tearDownClass)) { }
+
+    BedrockTester* tester;
+
+    void setupClass() { tester = new BedrockTester(_threadID, {{"-plugins", "Jobs,DB"}}, {});}
```
coleaeason: can you add some line breaks and style this as not a single line?

test/tests/jobs/UpdateJobTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    // Reset the jobs table
+    void tearDown() {
+        SData command("Query");
+        command["query"] = "DELETE FROM jobs WHERE jobID > 0;";
+        tester->executeWaitVerifyContent(command);
+    }
+
+    void tearDownClass() { delete tester; }
```
coleaeason: line breaks please

test/tests/jobs/UpdateJobTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    // Simple UpdateJob with all parameters
+    void updateJob() {
+        // Create the job
+        SData command("CreateJob");
+        string jobName = "job";
```
coleaeason: no need for this, can just do

```c++
command["name"] = "job"
```
on the next line

test/tests/jobs/UpdateJobTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+        command["name"] = jobName;
```
coleaeason: shouldn't you set the priority here too? to 500 or something? then after you update it do an assert that `oldPriority != new Priority`

test/tests/jobs/UpdateJobTest.cpp
mattabullock: does this not break uncrustify?

plugins/Jobs.cpp
```diff
@@ -881,7 +890,8 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
                                 SQ(request["data"]) + " " +
                                 (request.isSet("repeat") ? ", repeat=" + SQ(SToUpper(request["repeat"])) : "") +
                                 (!newNextRun.empty() ? ", nextRun=" + newNextRun : "") +
-                                "WHERE jobID=" +
+                                (request.isSet("jobPriority") ? ", priority=" + SQ(request.calc64("jobPriority")) : "") +
+                                " WHERE jobID=" +
```
mattabullock: No space after the quote

plugins/Jobs.cpp
```diff
@@ -857,6 +858,14 @@ bool BedrockPlugin_Jobs::processCommand(SQLite& db, BedrockCommand& command) {
             }
         }
 
+        // If a priority is provided, validate it
+        if (request.isSet("jobPriority")) {
```
mattabullock: You copied this from somewhere else, right? Maybe throw it into a function?

test/tests/jobs/UpdateJobTest.cpp
mattabullock: I imagine you added it in case there are more tests added? but yeah, don't add it if you're not adding more tests. (tearDown is between function calls here @coleaeason , not between classes)

Make sure you remove it from line 9 as well.

coleaeason: >(tearDown is between function calls here @coleaeason , not between classes)

this makes a lot more sense

test/tests/jobs/UpdateJobTest.cpp
coleaeason: well it doesn't run in this repo yet so no :D

alex-mechler:
Updated to fix all the comments above

PR #701, opened 2020-02-13
Fix atomicity in write cache

coleaeason:
@tylerkaraszewski This fixes an atomicity problem in `WriteCache`.

If two threads run `WriteCache` at same time with one entry in the cache, this line:
```
const string& name = (_lruMap.empty() ? db.read("SELECT name FROM cache LIMIT 1") : _lruMap.popLRU());
```
Will have `_lruMap.empty()` return false in both threads, then the first thread will call `_lruMap.popLRU()`, delete the last entry in the cache. Then the second thread calls `_lruMap.popLRU()` which causes us to try to pop a now empty cache hitting the assert:
`SASSERT(!empty());`.

This makes this operation atomic by checking for empty and doing the pop inside a single mutex lock.

I also did some style changes.


agzakula:
Oh, yep.  Atomicity is hard.

PR #712, opened 2020-02-26
Moving method to libstuff so thats its accessible in server-e

chiragsalian:
## Fixed Issues
Helps with this [comment](https://github.com/Expensify/Server-Expensify/pull/4256#discussion_r382755434)

## Web Tests
None

## Mobile Tests
None

## Automated Auth Tests
None

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2529,3 +2529,27 @@ void STerminateHandler(void) {
     // And we're out.
     abort();
 }
+
+bool SIsValidSQLiteDateModifier(const string& modifier) {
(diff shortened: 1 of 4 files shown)


libstuff/libstuff.cpp
```diff
@@ -2544,3 +2544,27 @@ void STerminateHandler(void) {
     // And we're out.
     abort();
 }
+
+bool SIsValidSQLiteDateModifier(const string& modifier) {
```
chiragsalian: I felt like the underscore in the name was weird so i updated it to `SIsValidSQLiteDateModifier`.

PR #719, opened 2020-03-03
Fix query shorthand syntax

coleaeason:
@tylerkaraszewski Please review. we accidentally made the `query:` syntax case sensitive when we merged the command refactor. This fixes that and adds automated tests.

also sorry, but i renamed all of the `status` variables to `query` because it was poor naming and clearly a copy paste.


coleaeason:
no idea what is up with the commit history on this branch...let me know if you want me to fix it

PR #744, opened 2020-04-03
Revert "Need to rollback to notify checkpoint thread it can do something"

coleaeason:
Reverts Expensify/Bedrock#742


cead22:
Skipping travis due to fire

PR #745, opened 2020-04-03
Revert "Check for new exception in SQLiteNode"

coleaeason:
Reverts Expensify/Bedrock#739


cead22:
Skipping travis due to fire

PR #788, opened 2020-05-14
Add Dockerfile

dinedal:
Creates a Dockerfile for Bedrock.

This is using `phusion/baseimage` because it's ubuntu xenial to match travis and syslogd built in, which Bedrock uses for logging.

Example usage:
```bash
mkdir -p db
touch db/bedrock.db
docker run --env DB=/db/bedrock.db --env SERVER_HOST=0.0.0.0:8888 -v `pwd`/db:/db -p 8888:8888 -p 3306:3306 dinedal/bedrockdb:latest
```

The env vars to command line args is a bit clever for my tastes, even, however it does allow flexibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,64 @@
+FROM phusion/baseimage:0.11 AS builder
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y software-properties-common \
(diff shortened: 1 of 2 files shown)


fukawi2:
Docker support has been added in https://github.com/Expensify/Bedrock/pull/1906

PR #798, opened 2020-05-19
[WIP] Log sync total from freshest

righdforsa:
There are periodic circumstances where we can see "to go" commit totals that are wildly off. AFAIK, this only happens after a network partition when a node and its neighbor are both restoring from backup, and one chooses the other as its `_syncPeer`. In this case, a node can be ~ "1000 commits to go" for an hour or more. Seems like we can do better. Not sure this is a great solution. By creating an internal `_freshestPeer` variable, we solve the problem, but we introduce a new set of "freshest peer" calculations that are separate from the ones done under the `SEARCHING` state. Even though they're functionally distinct, that seems like it's introducing confusion into the codebase, so I'd appreciate any suggestions on whether this makes sense.

Also, it wasn't immediately obvious to me how to do a proper test of this in a low-latency environment like a cluster in a dev vm where choosing the _syncPeer is going to be more or less random. Any ideas there would be appreciated as well.

Tests: ran `Bedrock/test` and `Bedrock/test/clustertest -repeatCount 3` with no errors

### DIFF
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -1347,6 +1347,9 @@ void SQLiteNode::_onMESSAGE(Peer* peer, const SData& message) {
                 _changeState(SEARCHING);
             } else {
                 // Otherwise, more to go
+                if (_syncPeer->calcU64("CommitCount") < _freshestPeer->calcU64("CommitCount")) {
+                    peerCommitCount = _freshestPeer->calcU64("CommitCount");
+                }
                 SINFO("Synchronization underway, at commitCount #"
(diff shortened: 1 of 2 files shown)


coleaeason:
Is there an issue to track this? I don't think there is an actual problem here. It **is** logging the correct information, it just isn't what you would expect it to be. Would love to know what the problem is as defined in an issue.

righdforsa:
No, logging "to go" based on the sync peer is not "what you would expect" until you understood why it was doing it. Maybe a shorter path to mitigating my concern would be adjusting the log line to be clearer on the relationship. But it seems like it should be relatively easy to track the highest commit peer even if this is not the place to do it.

Expensify internal issue is here. https://github.com/Expensify/Expensify/issues/103529

PR #799, opened 2020-05-20
Allow to update nextRun in UpdateJob

iwiznia:
Needed for https://github.com/Expensify/Expensify/issues/131401

Added automated tests, more tests in the web PR

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -828,6 +828,7 @@ void BedrockJobsCommand::process(SQLite& db) {
         //     - data  - A JSON object describing work to be done
         //     - repeat - A description of how to repeat (optional)
         //     - jobPriority - The priority of the job (optional)
+        //     - nextRun - The time to run the job (optional)
         //
         BedrockPlugin::verifyAttributeInt64(request, "jobID", 1);
         BedrockPlugin::verifyAttributeSize(request, "data", 1, BedrockPlugin_Jobs::MAX_SIZE_BLOB);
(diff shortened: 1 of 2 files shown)


plugins/Jobs.cpp
```diff
@@ -863,8 +864,13 @@ void BedrockJobsCommand::process(SQLite& db) {
         const string& nextRun = result[0][1];
         const string& lastRun = result[0][2];
 
-        // Are we rescheduling?
-        const string& newNextRun = request["repeat"].size() ? _constructNextRunDATETIME(nextRun, lastRun, request["repeat"]) : "";
+        // Passed next run takes priority over the one computed via the repeat feature
+        string newNextRun;
```
alex-mechler, line 868: We should do some validation on what newNextRun contains, since a malformed nextRun could cause the query to fail.

iwiznia: True, but we don't really validate this in any other command that receives it, for example [here](https://github.com/Expensify/Bedrock/blob/5987b1e312250f79bca46d988cd16ef5ca0f1794/plugins/Jobs.cpp#L1053) or [here](https://github.com/Expensify/Bedrock/blob/5987b1e312250f79bca46d988cd16ef5ca0f1794/plugins/Jobs.cpp#L509)

iwiznia: Also, not quite true, the query would not fail, it would leave the job with a weird nextRun value

alex-mechler: Good old sqlite everything is a string

Given that we aren't validating timestamps anywhere, and that this wouldnt break the query, this is fine

PR #812, opened 2020-06-09
Cole perftest

quinthar:
Is there any reason why we don't merge this in with master?


rafecolton:
Should probably merge master into the branch and make sure tests are passing

PR #832, opened 2020-08-04
Add SuppressCommandPort and ClearCommandPort as non-secure control commands

robertjchen:
Thanks for the review! The `SuppressCommandPort`, `ClearCommandPort` control commands should be permitted to be run from other hosts, at least for now.

## Tests
- Ensure other existing control commands e.g. `BeginBackup` are still restricted from being run through other hosts AND still can be successfully run from localhost.
- Ensure `SuppressCommandPort` and `ClearCommandPort` can be run from from localhost and external hosts.

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -1169,8 +1169,9 @@ bool BedrockServer::_handleIfStatusOrControlCommand(unique_ptr<BedrockCommand>& 
         _reply(command);
         return true;
     } else if (_isControlCommand(command)) {
-        // Control commands can only come from localhost (and thus have an empty `_source`).
-        if (command->request["_source"].empty()) {
+        // Control commands can only come from localhost (and thus have an empty `_source`)
+        // with the exception of non-secure control commands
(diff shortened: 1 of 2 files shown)


BedrockServer.cpp
```diff
@@ -2057,6 +2058,16 @@ bool BedrockServer::_isControlCommand(const unique_ptr<BedrockCommand>& command)
     return false;
 }
 
+bool BedrockServer::_isNonSecureControlCommand(const unique_ptr<BedrockCommand>& command) {
+    // A list of non-secure control commands that can be run from another host
+    if (SIEquals(command->request.methodLine, "SuppressCommandPort") ||
```
cead22, line 2063: `return SIEquals(command->request.methodLine, "SuppressCommandPort") || SIEquals(command->request.methodLine, "ClearCommandPort");`

robertjchen:
Thanks, updated and re-tested :+1:

cead22:
Looks good, I think we could do `SContains(command->request.methodLine, "CommandPort")` but I like the simplicity of this

PR #862, opened 2020-08-26
Switch to c++17 to get new mutex types and other new features

coleaeason:
@tylerkaraszewski Please review. Switches us to c++17 std, which gives us `shared_mutex`, some new string functions like `starts_with()` and `ends_with()` and more.

## Tests
Compiles, tests run 🤷‍♂️

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ endif
 GIT_REVISION = $(shell git rev-parse --short HEAD)
 PROJECT = $(shell git rev-parse --show-toplevel)
 INCLUDE = -I$(PROJECT) -I$(PROJECT)/mbedtls/include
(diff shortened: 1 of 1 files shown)


tylerkaraszewski:
I think we'll need to do this in all of our other repos as well.

quinthar:
What new mutex type does this open up that we care about?

coleaeason:
>What new mutex type does this open up that we care about?

It gives us `shared_mutex` as opposed to having to use `shared_timed_mutex` which has features we don't use. It also gives us new string functions that we could use from the standard library instead of our own. 

>I think we'll need to do this in all of our other repos as well.

Yeah, I have those PRs coming too.

tylerkaraszewski:
C++17 also adds [std::filesystem](https://en.cppreference.com/w/cpp/filesystem), some parallel algorithms that might be useful (though I don't have a current use case for them).

tylerkaraszewski (approved):

PR #916, opened 2020-10-26
Add optional overwrite parameter to `CreateJob` and `CreateJobs`

alex-mechler:
@iwiznia can you review please?

Part of https://github.com/Expensify/Expensify/issues/143805

This adds an optional parameter to `CreateJob` and `CreateJobs` where we will not overwrite any data when when attempting to create a unique job that already exists. This optional parameter is assumed to be `true` when not included to not break any existing behavior

## Tests

Added a new automated test, will be further tested in followup prs

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -505,10 +505,7 @@ void BedrockJobsCommand::process(SQLite& db) {
                 // If we found a job, but the data was different, we'll need to update it.
                 if (!result.empty()) {
                     updateJobID = SToInt64(result[0][0]);
-                    continue;
                 }
-
-                SINFO("ALEX NO DUPE FOUND");
             }
 
             // Record whether or not this job is scheduling itself in the future. If so, it's not suitable for
@@ -584,7 +581,6 @@ void BedrockJobsCommand::process(SQLite& db) {
                         STHROW("502 update query failed");
                     }
                 }
-
 
                 // If we are calling CreateJob, return early, there are no more jobs to create.
                 if (SIEquals(requestVerb, "CreateJob")) {
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -402,7 +402,7 @@ void BedrockJobsCommand::process(SQLite& db) {
         //     - jobPriority - High priorities go first (optional, default 500)
         //     - unique - if true, it will check that no other job with this name already exists, if it does it will
         //                return that jobID
-        //     - overwrite - Only applicable when submit is is true. When set to true it will overwrite the existing job
+        //     - overwrite - Only applicable when unique is is true. When set to true it will overwrite the existing job
         //                   with the new jobs data
         //     - parentJobID - The ID of the parent job (optional)
         //     - retryAfter - Amount of auto-retries before marking job as failed (optional)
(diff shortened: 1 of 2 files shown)


plugins/Jobs.cpp
```diff
@@ -402,6 +402,8 @@ void BedrockJobsCommand::process(SQLite& db) {
         //     - jobPriority - High priorities go first (optional, default 500)
         //     - unique - if true, it will check that no other job with this name already exists, if it does it will
         //                return that jobID
+        //     - overwrite - if true, and unique is true, it will overwrite the existing job with the new jobs data
```
iwiznia, line 405:
```suggestion
        //     - overwrite - Only applicable when submit is is true. When set to true it will overwrite the existing job with the new jobs data
```

plugins/Jobs.cpp
```diff
@@ -566,14 +570,16 @@ void BedrockJobsCommand::process(SQLite& db) {
...
             if (updateJobID) {
-                // Update the existing job.
-                if(!db.writeIdempotent("UPDATE jobs SET "
-                                         "repeat   = " + SQ(SToUpper(job["repeat"])) + ", " +
-                                         "data     = JSON_PATCH(data, " + safeData + "), " +
-                                         "priority = " + SQ(priority) + " " +
-                                       "WHERE jobID = " + SQ(updateJobID) + ";"))
-                {
-                    STHROW("502 update query failed");
+                if (!SContains(job, "overwrite") || job["overwrite"] == "true") {
```
iwiznia, line 573: I think you also want to check if `overwrite` is `==` and treat it as true too?

plugins/Jobs.cpp
```diff
@@ -566,14 +570,16 @@ void BedrockJobsCommand::process(SQLite& db) {
...
+                    // Update the existing job.
+                    if(!db.writeIdempotent("UPDATE jobs SET "
+                                             "repeat   = " + SQ(SToUpper(job["repeat"])) + ", " +
+                                             "data     = JSON_PATCH(data, " + safeData + "), " +
```
iwiznia, line 577: hmmm you sure about PATCH? Don't we want to replace the whole data?

plugins/Jobs.cpp
alex-mechler: I'm not quite following. What would we check if `overwrite` is `==` to?

plugins/Jobs.cpp
alex-mechler: This is the same query as before, just indented. I agree it probably should not be a `PATCH`, but I don't want to change existing behavior (in case any relies on this)

iwiznia: Oh... 🤦  you are right

plugins/Jobs.cpp
iwiznia: Damn sorry, I mean is equal to `''`

alex-mechler: Ah good catch, will add that

alex-mechler:
Updated

iwiznia (approved):

iwiznia:
All looks good, I did not hit the merge button because it seems real tests have not been done yet... feel free to merge when you do.

alex-mechler:
Ah, this was tested with https://github.com/Expensify/Web-Expensify/pull/29514, going to hit the merge button

PR #960, opened 2021-02-10
Preserve mockRequest when updating a job

alex-mechler:
# Fixed Issue
$ https://github.com/Expensify/Expensify/issues/151702

# Tests
New automatic test

Manually tested by
1. Enabling LOAD_TEST on dev
2. Requested a call from the inbox
3. Completed calls in expensiworks
4. _Verified_ I only got 1 call, not multiple that were mocked.

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -848,7 +848,7 @@ void BedrockJobsCommand::process(SQLite& db) {
 
         // Verify there is a job like this
         SQResult result;
-        if (!db.read("SELECT jobID, nextRun, lastRun "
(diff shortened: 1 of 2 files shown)


iwiznia (commented):
Do we need to cover the opposite case when job was not mocked and we are updating its data?

alex-mechler:
I don't think that is a case that will come up, as the call to `UpdateJob` will have to explicitly include a `mockRequest` in the data, but it is probably safer to add that. Will add that in

alex-mechler:
Updated

iwiznia (approved):

PR #964, opened 2021-02-24
Fix to compile for gcc-10, g++-10 (Debian Sid)

johnaman:


johnaman:
Of course this push should not be merged until gcc-10 becomes standard.

PR #970, opened 2021-03-04
[WIP] Break down postPoll

righdforsa:
This breaks down the timing inside postPoll() to try to get a clearer picture where it's spending time under stress. 

Tests:
Spammed with 10000 "Ping" commands and waited 30s to close the sockets, to see if the logs looked reasonable.
[WIP: they don't]


righdforsa:
We decided to move forward with parallel bedrock networking 👋

PR #976, opened 2021-03-19
Use private apt mirror for travis tests

rafecolton:
@coleaeason I think this will make it impossible to run tests from forks. I think we still want this due to the issue we had the other day, but just checking...


rafecolton:
Confirmed [here](https://travis-ci.com/github/Expensify/Bedrock/builds/220968822) that if the travis commit is in a fork (and the commit is also not included in a branch that is not a fork), it will not be provided with the secure env vars and will do the alternate build instead. This is ready for review.

iwiznia (approved):

PR #979, opened 2021-03-25
Make Bedrock clang++ compatible

TomMD:
Bedrock is only two small changes away from clang++ compatibility, which is fixed in this PR.

Many analysis tools, such as Infer used by Muse, lean on Clang to compile the code into clang IR for analysis. By making bedrock compatible we can also use Infer and similar tools on downstream projects that include bedrock.

@flodnv This is the PR we previously discussed.

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2602,7 +2602,7 @@ string SGetCurrentExceptionName()
     int status = 0;
     size_t length = 1000;
     char buffer[length];
-    memset(buffer,0,length);
+    memset(buffer, 0, length);
 
     // Demangle the name of the current exception.
     // See: https://libcxxabi.llvm.org/spec.html for details on this ABI interface.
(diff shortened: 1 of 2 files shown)


libstuff/libstuff.cpp
```diff
@@ -2601,7 +2601,8 @@ string SGetCurrentExceptionName()
     // exception name.
     int status = 0;
     size_t length = 1000;
-    char buffer[length] = {0};
+    char buffer[length];
+    memset(buffer,0,length);
```
flodnv, line 2605: Let's do this update to follow our style: 
```suggestion
    memset(buffer, 0, length);
```

flodnv (commented):
Thanks for this! @tylerkaraszewski are you ok with this?

tylerkaraszewski (approved):

flodnv (approved):

flodnv:
@TomMD unfortunately you'll have to GPG sign your commits for us to merge this. If it's easier for you I can just submit another PR with these same changes?

TomMD:
@flodnv Perhaps you making the commits would be for the best.

flodnv:
NP, done [here](https://github.com/Expensify/Bedrock/pull/980), thanks again for your help!

PR #993, opened 2021-04-13
[WIP] working multithread client net first commit

righdforsa:
Using this for saving working code states while I'm developing the multithread design.

Currently handles:
- launching a thread for each accepted socket
- shutting down a socket after command completion
- closing open empty sockets for shutdown

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -17,6 +17,12 @@ void BedrockServer::acceptCommand(SQLiteCommand&& command, bool isNew) {
     acceptCommand(make_unique<SQLiteCommand>(move(command)), isNew);
 }
 
+void BedrockServer::acceptAndLaunchThread() {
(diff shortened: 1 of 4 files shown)


righdforsa:
lol 🤦 first commit with "working" commit message won't build.

righdforsa:
> ~lol 🤦 first commit with "working" commit message won't build. ~

Fixed the previous commit and force pushed. Works now. Builds, starts, handles new connections, closes completed connections, (evaluated with netstat, no accumulation of sockets in the normal case) and shuts down cleanly with sockets that were opened but not written to. (evaluated with a bash script that writes 100 connections, and on it % 10 == 0, opens one connection with no data in the background then keeps going.)

righdforsa:
Added automated test with higher level of multiple simultaneous requests than other existing parallel tests.

PR #1009, opened 2021-04-27
Make Bedrock Tests ASSERT/EXPECT failures print to cout with details - Take 2

johnmlee101:
Reverts Expensify/Bedrock#1007

But also fixes working with all types.


test/lib/Comparator.h
```diff
@@ -0,0 +1,73 @@
+#pragma once
+#include <libstuff/libstuff.h>
```
tylerkaraszewski, line 2: Let's not make tpunit++ dependent on libstuff, just include the specific headers we'll need here.

test/lib/tpunit++.hpp
```diff
@@ -75,22 +76,22 @@ using namespace std;
...
-#define EXPECT_NOT_EQUAL(lhs, rhs) if((lhs) != (rhs)) { PASS(); } else { FAIL(); }
-#define ASSERT_GREATER_THAN(lhs, rhs) if((lhs) > (rhs)) { PASS(); } else { ABORT(); }
-#define EXPECT_GREATER_THAN(lhs, rhs) if((lhs) > (rhs)) { PASS(); } else { FAIL(); }
-#define ASSERT_GREATER_THAN_EQUAL(lhs, rhs) if((lhs) >= (rhs)) { PASS(); } else { ABORT(); }
-#define EXPECT_GREATER_THAN_EQUAL(lhs, rhs) if((lhs) >= (rhs)) { PASS(); } else { FAIL(); }
-#define ASSERT_LESS_THAN(lhs, rhs) if((lhs) < (rhs)) { PASS(); } else { ABORT(); }
-#define EXPECT_LESS_THAN(lhs, rhs) if((lhs) < (rhs)) { PASS(); } else { FAIL(); }
-#define ASSERT_LESS_THAN_EQUAL(lhs, rhs) if((lhs) <= (rhs)) { PASS(); } else { ABORT(); }
-#define EXPECT_LESS_THAN_EQUAL(lhs, rhs) if((lhs) <= (rhs)) { PASS(); } else { FAIL(); }
+#define ASSERT_TRUE(condition) if(condition) { PASS(); } else { testLog("assertion failed: " + SToStr(#condition) + " resolved to FALSE"); ABORT(); }
```
tylerkaraszewski, line 79: You don't need `SToStr` around the stringification operator, you can just do:
```
"assertion failed: " #condition " resolved to FALSE"
```

test/lib/tpunit++.hpp
```diff
@@ -75,22 +76,22 @@ using namespace std;
...
+#define EXPECT_TRUE(condition) if(condition) { PASS(); } else { testLog("expect failed: " + SToStr(#condition) + " resolved to FALSE"); FAIL(); }
+#define ASSERT_FALSE(condition) if(condition) { testLog("assertion failed: " + SToStr(#condition) + " resolved to TRUE"); ABORT(); } else { PASS(); }
+#define EXPECT_FALSE(condition) if(condition) { testLog("expect failed: " + SToStr(#condition) + " resolved to TRUE"); FAIL(); } else { PASS(); }
+#define ASSERT_EQUAL(lhs, rhs) if((lhs) == (rhs)) { PASS(); } else { EqualComparator(lhs, rhs); ABORT(); }
+#define EXPECT_EQUAL(lhs, rhs) if((lhs) == (rhs)) { PASS(); } else { EqualComparator(lhs, rhs); FAIL(); }
+#define ASSERT_NOT_EQUAL(lhs, rhs) if((lhs) != (rhs)) { PASS(); } else { EqualComparator(lhs, rhs); ABORT(); }
+#define EXPECT_NOT_EQUAL(lhs, rhs) if((lhs) != (rhs)) { PASS(); } else { EqualComparator(lhs, rhs); FAIL(); }
+#define ASSERT_GREATER_THAN(lhs, rhs) if((lhs) > (rhs)) { PASS(); } else { testLog("incorrectly asserted " + SToStr(lhs) + " > " + SToStr(rhs)); ABORT(); }
```
tylerkaraszewski, line 87: We can use `to_string()` instead of `SToStr()` here, it's built-in.

test/lib/tpunit++.hpp
```diff
@@ -75,22 +76,22 @@ using namespace std;
...
-#define EXPECT_GREATER_THAN_EQUAL(lhs, rhs) if((lhs) >= (rhs)) { PASS(); } else { FAIL(); }
-#define ASSERT_LESS_THAN(lhs, rhs) if((lhs) < (rhs)) { PASS(); } else { ABORT(); }
-#define EXPECT_LESS_THAN(lhs, rhs) if((lhs) < (rhs)) { PASS(); } else { FAIL(); }
-#define ASSERT_LESS_THAN_EQUAL(lhs, rhs) if((lhs) <= (rhs)) { PASS(); } else { ABORT(); }
-#define EXPECT_LESS_THAN_EQUAL(lhs, rhs) if((lhs) <= (rhs)) { PASS(); } else { FAIL(); }
+#define ASSERT_TRUE(condition) if(condition) { PASS(); } else { testLog("assertion failed: " + SToStr(#condition) + " resolved to FALSE"); ABORT(); }
+#define EXPECT_TRUE(condition) if(condition) { PASS(); } else { testLog("expect failed: " + SToStr(#condition) + " resolved to FALSE"); FAIL(); }
+#define ASSERT_FALSE(condition) if(condition) { testLog("assertion failed: " + SToStr(#condition) + " resolved to TRUE"); ABORT(); } else { PASS(); }
+#define EXPECT_FALSE(condition) if(condition) { testLog("expect failed: " + SToStr(#condition) + " resolved to TRUE"); FAIL(); } else { PASS(); }
+#define ASSERT_EQUAL(lhs, rhs) if((lhs) == (rhs)) { PASS(); } else { EqualComparator(lhs, rhs); ABORT(); }
```
tylerkaraszewski, line 83: It occurs to me we don't actually need the comparison in `EqualComparator`, all we want is the generic printing that falls back to "unprintable" instead of failing to compile. The actual comparison has already been done here.

We could probably simplify all the template cone that way.

tylerkaraszewski (changes requested):

test/lib/tpunit++.hpp
johnmlee101: Oh true, so we type-cast with the class first then print?

johnmlee101: Actually not sure what you mean here.

tylerkaraszewski: We already know (for this line) that if we hit the `else`, these are not equal (otherwise we would have run `PASS()`).

So where we do `EqualComparator(lhs, rhs);` we could just do something like: `cout << "Not equal " << Print(lhs, rhs);` where `Print` was a simpler version of `EqualComparator` that didn't need to actually do any comparison, just the printing.

If that was the case we could replace:
```
        template <typename T, enable_if_t<is_integral<T>::value, bool> = true, typename U>
        EqualComparator(T a, U b)
        {
            // Integral case.
            if (Equal<T>(a) != (T) b) {
                cout << "(Integral): " << a << " != " << b << "\n";
            } else {
                cout << "(Integral): " << a << " == " << b << "\n";
            }
        }
```


With:
```
        template <typename T, enable_if_t<is_integral<T>::value, bool> = true, typename U>
        EqualComparator(T a, U b)
        {
            cout << "(Integral): " << a << " == " << b << "\n";
        }
```

And then we don't even need `class Equal` anymore.

test/lib/PrintEquality.h
```diff
@@ -0,0 +1,57 @@
...
+                cout << "equal (unhandled type: " << buffer << ")" << "\n";
+            }
+        }
+
+        template <typename U>
+        PrintEquality(string a, U b, bool isEqual)
+        {
+            // Non-integer base case.
+            if (!isEqual) {
+                cout << "(string): \"" << a << "\" != \""<< b << "\"" << "\n";
```
tylerkaraszewski, line 42: Missing space before `<<` (repeated on several lines below).

test/lib/PrintEquality.h
```diff
@@ -0,0 +1,57 @@
...
+                cout << "Not equal (unhandled type: " << buffer << ")" << "\n";
+            } else {
+                cout << "equal (unhandled type: " << buffer << ")" << "\n";
+            }
+        }
+
+        template <typename U>
+        PrintEquality(string a, U b, bool isEqual)
+        {
+            // Non-integer base case.
```
tylerkaraszewski, line 40: Wrong comment.

test/lib/tpunit++.hpp
```diff
@@ -75,22 +76,22 @@ using namespace std;
...
+#define ASSERT_TRUE(condition) if(condition) { PASS(); } else { testLog("assertion failed: " #condition " resolved to FALSE"); ABORT(); }
+#define EXPECT_TRUE(condition) if(condition) { PASS(); } else { testLog("expect failed: " #condition " resolved to FALSE"); FAIL(); }
+#define ASSERT_FALSE(condition) if(condition) { testLog("assertion failed: " #condition " resolved to TRUE"); ABORT(); } else { PASS(); }
+#define EXPECT_FALSE(condition) if(condition) { testLog("expect failed: " #condition " resolved to TRUE"); FAIL(); } else { PASS(); }
+#define ASSERT_EQUAL(lhs, rhs) if((lhs) == (rhs)) { PASS(); } else { PrintEquality(lhs, rhs, false); ABORT(); }
+#define EXPECT_EQUAL(lhs, rhs) if((lhs) == (rhs)) { PASS(); } else { PrintEquality(lhs, rhs, false); FAIL(); }
+#define ASSERT_NOT_EQUAL(lhs, rhs) if((lhs) != (rhs)) { PASS(); } else { PrintEquality(lhs, rhs, true); ABORT(); }
+#define EXPECT_NOT_EQUAL(lhs, rhs) if((lhs) != (rhs)) { PASS(); } else { PrintEquality(lhs, rhs, true); FAIL(); }
+#define ASSERT_GREATER_THAN(lhs, rhs) if((lhs) > (rhs)) { PASS(); } else { testLog("incorrectly asserted " + to_string(lhs) + " > " + to_string(rhs)); ABORT(); }
```
tylerkaraszewski, line 87: Why are we not doing the same thing for `GREATER_THAN`, `LESS_THAN` etc that we did for `==` and `!=`? Just cause these are almost certainly numbers?

test/lib/tpunit++.hpp
```diff
@@ -305,21 +308,27 @@ namespace tpunit {
                                          const std::list<std::string>& before, const std::list<std::string>& after, int threads,
                                          std::function<void()> threadInitFunction);
 
+         /**
+          * This method writes to a temporary buffer and formats the message nicely for debugging
+          * purposes if a test were to fail.
+          */
+         void testLog(const string& newLog);
```
tylerkaraszewski, line 315: What if we called this `TESTINFO` or something that looks more like our normal logging functions?

tylerkaraszewski (changes requested):

test/lib/PrintEquality.h
```diff
@@ -0,0 +1,57 @@
...
+            if (!isEqual) {
+                cout << "(string): \"" << a << "\" != \""<< b << "\"" << "\n";
+            } else {
+                cout << "(string): \"" << a << "\" == \""<< b << "\"" << "\n";
+            }
+        }
+
+        template <typename U>
+        PrintEquality(const char* a, U b, bool isEqual) {
+            // Non-integer base case.
```
tylerkaraszewski, line 50: Wrong comment here, too.

test/lib/tpunit++.hpp
johnmlee101: Correct, it never caused issues with compilation so far. Going to assume its not quite as prone to casting issues as equality chcecks

test/lib/tpunit++.hpp
johnmlee101: Sure!

test/lib/PrintEquality.h
```diff
@@ -0,0 +1,57 @@
...
+                cout << "Not equal (unhandled type: " << buffer << ")" << "\n";
+            } else {
+                cout << "equal (unhandled type: " << buffer << ")" << "\n";
+            }
+        }
+
+        template <typename U>
+        PrintEquality(string a, U b, bool isEqual)
+        {
+            // Non-integral base case.
```
tylerkaraszewski, line 40: I meant that this was not the "base case" but rather a string specialization. The "base case" prints "unhandled type". :)

tylerkaraszewski (changes requested):
O left one more comment on the comments in the change. One other thing to do before finalizing it is to run all the tests for the other repos (auth, fuzzybot, expensifybackupmanager) against this to make sure you haven't broken them. Can post that they passed in the body of the PR.

test/lib/tpunit++.cpp
```diff
@@ -396,6 +403,18 @@ void tpunit::TestFixture::tpunit_detail_do_tests(TestFixture* f) {
...
 }
 
+void tpunit::TestFixture::TESTINFO(const string& newLog) {
+    lock_guard<recursive_mutex> lock(*(_mutex));
+
+    // Format the buffer with an indent as we print it out.
+    testOutputBuffer += "    " + newLog + "\n";
+}
+
+void tpunit::TestFixture::printTestBuffer() {
```
tylerkaraszewski, line 413: This should probably lock the mutex as well.

test/lib/PrintEquality.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include <iostream>
+#include <cxxabi.h>
```
tylerkaraszewski, line 4: We need at least `<string>` here as well.

tylerkaraszewski (commented):
Almost there. Two small comments and waiting for the test results for the other repos.

test/lib/PrintEquality.h
johnmlee101: 🤔 weird that its compiling fine for this?

tylerkaraszewski: It's compiling fine because you only use this file here:
https://github.com/Expensify/Bedrock/pull/1009/files#diff-0049b41cb8098a521abb593b7921df73b67adeadfa658315a972292110895479R32

And that's below an include of `<string>` in that file. It will stop compiling if someone removes that include from tpunit++.hpp, or includes this file in some other place that doesn't include `string`.

johnmlee101: Ah gotcha. Makes sense. Added.

tylerkaraszewski (approved):
Approving the code but lets test that the auth, fuzzybot, and expensifybackupmanager tests all compile before we merge.

johnmlee101:
Yep!

johnmlee101:
### auth
image

### expensifybackup
image


### fuzzybot
all compiles but its been broken for me. I might ask for you to test this.
```
/usr/bin/ld: externalLib/snowball/libstemmer.o(libstemmer.o): relocation R_X86_64_32 against `.data' can not be used when making a shared object; recompile with -fPIC
externalLib/snowball/libstemmer.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Makefile:96: recipe for target 'concierge.so' failed
make: *** [concierge.so] Error 1
```

tylerkaraszewski:
Doesn't look like you built the auth tests, jsut auth.

johnmlee101:
### Auth test
image
image
ran it successfully

johnmlee101:
Do you think I should have someone with a working Fuzzybot test this @tylerkaraszewski ?

johnmlee101:
Tyler tested Fuzzybot, all good!

PR #1027, opened 2021-05-24
Updates query parsing in peek to make it robust

MonilBhavsar:
### Details

Updates query parsing in `peek` function to execute a query using `db.read()`. If it fails, it escalates it to `process`. If it's a bad query, it will fail in `process` too. 

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/161171

### Tests

- All Bedrock `Write` tests
- Identifies this query as a read and executes it in `peek`, which was escalated to `process` before

```
query: PRAGMA cache_size = 0; select 'caca';
format: json

200 OK
commitCount: 1874
nodeName: bedrock
peekTime: 3156
totalTime: 4363
unaccountedTime: 1007
Content-Length: 40

{"headers":["'caca'"],"rows":[["caca"]]}
``` 
_________

cc @flodnv

### DIFF
--- a/plugins/DB.cpp
+++ b/plugins/DB.cpp
@@ -64,30 +64,19 @@ bool BedrockDBCommand::peek(SQLite& db) {
         STHROW("502 Query aborted");
     }
 
-    if (SStartsWith(upperQuery, "SELECT ")) {
-        // Seems to be read-only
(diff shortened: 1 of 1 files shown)


flodnv (approved):
Thanks!

tylerkaraszewski (approved):

PR #1028, opened 2021-05-25
Fixes PHP Integration Tests caused by UPDATE query parsing

MonilBhavsar:
### Details
Context here - https://expensify.slack.com/archives/C03TQ48KC/p1621953197258100
### Fixed Issues
PHP Integration tests were failing because of the query like `INSERT INTO foo(a, b) VALUES('I update a', 'query')` was aborted as it was identified as `UPDATE` query without `WHERE` clause.

So, instead of using `SContains()` used `SStartWith()`
### Tests
- All Write Tests
- PHP Integration EgenciaFeedTest
_________

### DIFF
--- a/plugins/DB.cpp
+++ b/plugins/DB.cpp
@@ -69,7 +69,7 @@ bool BedrockDBCommand::peek(SQLite& db) {
     int preChangeCount = db.getChangeCount();
     if (!db.read(query, result)) {
         if (shouldRequireWhere &&
-            (SContains(upperQuery, "UPDATE ") || SContains(upperQuery, "DELETE ")) &&
+            (SStartsWith(upperQuery, "UPDATE") || SStartsWith(upperQuery, "DELETE")) &&
             !SContains(upperQuery, " WHERE ")) {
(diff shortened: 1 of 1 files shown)


flodnv (changes requested):
Can you please add a test query in WriteTest.cpp that replicates the problem (a test that fails before this change)?

tylerkaraszewski (approved):
I'm going to approve this even though it's basically a hack, only because the original code is an equal level of hackiness. We should really do these checks with the sqlite3 authorizer: https://www.sqlite.org/c3ref/c_alter_table.html

MonilBhavsar:
Will look into this and lets get it done the right way.

flodnv:
Yes, but for now let's get this merged so we can unblock the deploy 🙃 @MonilBhavsar maybe create 2 followup issues?
1. Add tests covering this
2. Do it better, the way @tylerkaraszewski is suggesting

iwiznia (approved):

flodnv (approved):

PR #1032, opened 2021-05-26
Added test case where SQL keywords are passed as values

MonilBhavsar:
### Details
Added a couple of tricky test cases which can bypass the query parser 

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/165143

### Tests
All `Write` Tests

_________

### DIFF
--- a/test/tests/WriteTest.cpp
+++ b/test/tests/WriteTest.cpp
@@ -184,10 +184,11 @@ struct WriteTest : tpunit::TestFixture {
         SData query2("query: INSERT INTO stuff VALUES ( NULL, 12, 'Do not delete the test' );");
         tester->executeWaitVerifyContent(query2);
 
-        // Monil B - 05/26/21
-        // Uncomment this test once https://github.com/Expensify/Expensify/issues/165207 is solved
-        // SData query3("query: UPDATE stuff SET info = 'This is not a where clause';");
-        // tester->executeWaitVerifyContent(query3, "502 Query aborted");
+        // This is a false test case, This query should't get executed. This is currently a limitation of our parsing
+        // As "nowhere" parameter is not provided, this query should get aborted and prevent all rows from updating
+        // Change the expected result to "502 Query aborted" once https://github.com/Expensify/Expensify/issues/165207 is solved
+        SData query3("query: UPDATE stuff SET info = 'This is not a where clause';");
+        tester->executeWaitVerifyContent(query3);
     }
 
 } __WriteTest;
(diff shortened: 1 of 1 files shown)


test/tests/WriteTest.cpp
```diff
@@ -18,14 +18,15 @@ struct WriteTest : tpunit::TestFixture {
...
+                              TEST(WriteTest::keywordsAsValue),
                               AFTER_CLASS(WriteTest::tearDown)) { }
 
     BedrockTester* tester;
 
     void setup() {
         tester = new BedrockTester({}, {
             "CREATE TABLE foo (bar INTEGER);",
-            "CREATE TABLE stuff (id INTEGER PRIMARY KEY, value INTEGER);",
+            "CREATE TABLE stuff (id INTEGER PRIMARY KEY, value INTEGER, info TEXT);",
```
flodnv, line 29: curious why this change and the one below?

test/tests/WriteTest.cpp
```diff
@@ -176,4 +177,17 @@ struct WriteTest : tpunit::TestFixture {
...
     }
 
+    void keywordsAsValue() {
+        SData query("query: INSERT INTO stuff VALUES ( NULL, 11, 'Please update the test' );");
+        tester->executeWaitVerifyContent(query);
+
+        SData query2("query: INSERT INTO stuff VALUES ( NULL, 12, 'Do not delete the test' );");
+        tester->executeWaitVerifyContent(query2);
+
+        // Monil B - 05/26/21
```
flodnv, line 187: no need for this comment (git blame does this for us 🙃 )

test/tests/WriteTest.cpp
```diff
@@ -176,4 +177,17 @@ struct WriteTest : tpunit::TestFixture {
...
+        // Uncomment this test once https://github.com/Expensify/Expensify/issues/165207 is solved
+        // SData query3("query: UPDATE stuff SET info = 'This is not a where clause';");
+        // tester->executeWaitVerifyContent(query3, "502 Query aborted");
```
flodnv, line 190: Maybe uncomment this test and make it pass with the (unexpected) current response, and leave this commented line explaining why it should be that way instead?

flodnv (changes requested):
Thanks!

test/tests/WriteTest.cpp
MonilBhavsar: We want to include a test where SQL keywords like UPDATE, DELETE are inserted into the table as a values. There was not any column to add TEXT data, so created a column here.
Below in the other query, passed NULL to match the table schema.

test/tests/WriteTest.cpp
MonilBhavsar: Okay removing this.

test/tests/WriteTest.cpp
MonilBhavsar: Yes, can be done. It will also make more sense.

MonilBhavsar:
Updated

flodnv (approved):
Thanks!!

PR #1073, opened 2021-07-09
Remove reference to installing Bedrock from our apt repo

rafecolton:
It is being decommissioned, per https://github.com/Expensify/Expensify/issues/170058

cc @quinthar


righdforsa (approved):
👍  We are decomming that apt server.

PR #1080, opened 2021-07-23
Revert "Skip munmap call"

righdforsa:
Reverts Expensify/Bedrock#1078

Servers are crashing suspiciously today. This is the only code that was deployed yesterday.


johnmlee101 (approved):

righdforsa:
Our leading hypothesis is that other places in the sqlite code use the munmap call, and that by bypassing it, we created a memory leak.

PR #1089, opened 2021-09-13
[HOLD]Revert "Skip final munmap call that takes minutes."

justinpersaud:
Reverts Expensify/Bedrock#1086

Just in case, may not be needed.


fukawi2 (approved):
Hopefully we won't need it 🤞

fukawi2:
Looks like we don't need this 🎉

PR #1105, opened 2021-10-13
Mark infinitely retrying jobs as FAILED

flodnv:
### Details

Some jobs can retry forever (if no one ever calls FinishJob / FailJob, but keeps calling GetJob(s)). This fixes that by adding a maximum number of 10 retries. 

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/179455

### Tests

Automated

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -787,16 +787,29 @@ void BedrockJobsCommand::process(SQLite& db) {
         if (!retriableJobs.empty()) {
             for (auto job : retriableJobs) {
                 SINFO("Updating job with retryAfter " << job["jobID"]);
+                STable jobData = SParseJSONObject(job["data"]);
+                if (SToInt(jobData["retryAfterCount"]) >= 10) {
+                    SINFO("Job " << job["jobID"] << " has retried 10 times, marking it as FAILED.");
+                    string failQuery = "UPDATE jobs "
+                                       "SET state='FAILED' "
+                                       "WHERE jobID = " + SQ(job["jobID"]) + ";";
+                    if (!db.writeIdempotent(failQuery)) {
+                        STHROW("502 Update failed");
+                    }
+                    // TODO: Remove job from returned jobs...?
+                    continue;
+                }
                 string currentTime = SUNQUOTED_CURRENT_TIMESTAMP();
                 string retryAfterDateTime = "DATETIME(" + SQ(currentTime) + ", " + SQ(job["retryAfter"]) + ")";
                 string repeatDateTime = _constructNextRunDATETIME(job["nextRun"], currentTime, job["repeat"]);
                 string nextRunDateTime = repeatDateTime != "" ? "MIN(" + retryAfterDateTime + ", " + repeatDateTime + ")" : retryAfterDateTime;
                 bool isRepeatBasedOnScheduledTime = SToUpper(job["repeat"]).find("SCHEDULED") != string::npos;
                 string updateQuery = "UPDATE jobs "
-                                     "SET state='RUNQUEUED', "
-                                         "lastRun=" + SQ(currentTime) + ", " +
-                                         "nextRun=" + nextRunDateTime + " " +
-                                         (isRepeatBasedOnScheduledTime ? ", data=JSON_SET(data, '$.originalNextRun', " + SQ(job["nextRun"]) + ") ": "") +
+                                     "SET state = 'RUNQUEUED', "
+                                         "lastRun = " + SQ(currentTime) + ", " +
+                                         "nextRun = " + nextRunDateTime + ", " +
+                                         "data = JSON_SET(data, '$.retryAfterCount', COALESCE(JSON_EXTRACT(data, '$.retryAfterCount'), 0) + 1" + // Set this so we don't retry infinitely (see above)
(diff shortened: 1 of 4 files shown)


plugins/Jobs.cpp
```diff
@@ -787,16 +787,29 @@ void BedrockJobsCommand::process(SQLite& db) {
...
+                STable jobData = SParseJSONObject(job["data"]);
+                if (SToInt(jobData["retryAfterCount"]) >= 10) {
+                    SINFO("Job " << job["jobID"] << " has retried 10 times, marking it as FAILED.");
+                    string failQuery = "UPDATE jobs "
+                                       "SET state='FAILED' "
+                                       "WHERE jobID = " + SQ(job["jobID"]) + ";";
+                    if (!db.writeIdempotent(failQuery)) {
+                        STHROW("502 Update failed");
+                    }
+                    // TODO: Remove job from returned jobs...?
```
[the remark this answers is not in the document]
iwiznia: I think we need to remove it, no? Otherwise when it runs and tries to finish/fail it, it will fail?

plugins/Jobs.cpp
```diff
@@ -787,16 +787,29 @@ void BedrockJobsCommand::process(SQLite& db) {
...
                 bool isRepeatBasedOnScheduledTime = SToUpper(job["repeat"]).find("SCHEDULED") != string::npos;
                 string updateQuery = "UPDATE jobs "
-                                     "SET state='RUNQUEUED', "
-                                         "lastRun=" + SQ(currentTime) + ", " +
-                                         "nextRun=" + nextRunDateTime + " " +
-                                         (isRepeatBasedOnScheduledTime ? ", data=JSON_SET(data, '$.originalNextRun', " + SQ(job["nextRun"]) + ") ": "") +
+                                     "SET state = 'RUNQUEUED', "
+                                         "lastRun = " + SQ(currentTime) + ", " +
+                                         "nextRun = " + nextRunDateTime + ", " +
+                                         "data = JSON_SET(data, '$.retryAfterCount', COALESCE(JSON_EXTRACT(data, '$.retryAfterCount'), 0) + 1" + // Set this so we don't retry infinitely (see above)
```
iwiznia, line 811: You already have the data, so you could directly set it, instead of doing the coalesce and + 1 in sql? NAB

test/tests/jobs/InifiniteRetryAfterTest.cpp
```diff
@@ -0,0 +1,47 @@
...
+     * This tests that a job with a retryAfter won't get requeued forever.
+     * We want it to fail after 10 tries.
+     */
+    void test() {
+        BedrockTester tester = BedrockTester({{"-plugins", "Jobs,DB"}}, {});
+
+        // Create a job
+        SData createJob("CreateJob");
+        createJob["name"] = "infinite-job";
+        createJob["retryAfter"] = "+1 SECOND";
```
iwiznia, line 22:
```suggestion
        createJob["retryAfter"] = "QUEUED +1 SECOND";
```
?

plugins/Jobs.cpp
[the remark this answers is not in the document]
tylerkaraszewski: I think the same as @iwiznia , we shouldn't return it or it'll run again and probably throw some error when we see it's already failed.

plugins/Jobs.cpp
```diff
@@ -787,16 +787,29 @@ void BedrockJobsCommand::process(SQLite& db) {
         if (!retriableJobs.empty()) {
             for (auto job : retriableJobs) {
                 SINFO("Updating job with retryAfter " << job["jobID"]);
+                STable jobData = SParseJSONObject(job["data"]);
+                if (SToInt(jobData["retryAfterCount"]) >= 10) {
```
bondydaa, line 791: I was wondering if we needed to check if `jobData["retryAfterCount"]` exists but looking at what `SToInt` does: https://github.com/Expensify/Bedrock/blob/9b8fe5493e3ae6787830a65c6930cc631ec9164b/libstuff/libstuff.cpp#L2739-L2741

we're fine right? This is because the `val.c_str()` will return a c string that is null terminated (`\0`) https://www.cplusplus.com/reference/string/string/c_str/ and `atoi` will return 0 if the passed c string has only whitespace characters or a non valid integer value. 

so if `jobData["retryAfterCount"]` isn't set for a job then the c string returned by `val.c_string()` will just be the null terminated value and `atoi` will return 0 for that. 

Just rubber ducking this out to make sure I'm following it properly.

bondydaa (dismissed):
I think we're good here, I'm very unsure about if we should or shouldn't be returning the job where you've called it out but should be an easy fix if we need to not return it so I'm fine as is.

test/tests/jobs/InifiniteRetryAfterTest.cpp
flodnv: No? 

image

flodnv:
Updated:
- Fixed unrelated test
- Remove job from returned jobs... I am not a fan of this solution, the other option is a small refactor (because we store strings inside the `jobList`...
- Last night after pushing this PR I realized it had a massive flaw: it would've broken all repeating jobs with a retryAfter 😅 I fixed that in the last commit: In FinishJob, remove the retryAfterCount value.

PR #1137, opened 2021-12-08
Fix shutdown race condition

righdforsa:
### Details
To prevent a failed shutdown when Auth servers switch from STANDINGDOWN to SEARCHING with commands in the early part of the "processCommand" function, we need to catch an additional edge case, otherwise we get stuck with the command endlessly retrying.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/176212

### Tests
Ran GracefulFailoverTest repeatedly. 
- To reproduce the failure, I changed the Graceful shutdown timeout `_gracefulShutdownTimeout.alarmDuration` from 60s to 15s and added a random 30s sleep to the top of `BedrockCore::processCommand`. This fix resolved that issue. Ran the test 10x to make sure it was good and fixed.
- Unwound the 15s shutdown timeout and ran the test again 10x to make sure it was still good.
- Unwound the 30s random processCommand timeout and all debug logging and ran the test again 10x to make sure it was still good. Interestingly, this one had a shutdown failure in one of the follower nodes on the 7th test pass, but it was in a different part of the code, so I believe it's distinct, and a repeat of the tests passed all 10x.

_________

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,9 @@ sqlitecluster/sqlitecluster
 libstuff/libstuff.d
 libstuff/libstuff.h.gch
 .idea
+
+*.[do]
+perftest/perftest
+*.db
+*.db-wal
+*.db-shm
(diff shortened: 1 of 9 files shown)


righdforsa:
Aww, poop, somehow this has unrelated perftest branch changes. I'll try to unwind.

righdforsa:
Rebased and made a new PR

PR #1138, opened 2021-12-08
Fix shutdown race condition

righdforsa:
Note: rebased version of https://github.com/Expensify/Bedrock/pull/1137

### Details
To prevent a failed shutdown when Auth servers switch from STANDINGDOWN to SEARCHING with commands in the early part of the "processCommand" function, we need to catch an additional edge case, otherwise we get stuck with the command endlessly retrying.

### Fixed Issues
Fixes [Expensify/Expensify#176212](https://github.com/Expensify/Expensify/issues/176212)

### Tests
Ran GracefulFailoverTest repeatedly.

To reproduce the failure, I changed the Graceful shutdown timeout _gracefulShutdownTimeout.alarmDuration from 60s to 15s and added a random 30s sleep to the top of BedrockCore::processCommand. This fix resolved that issue. Ran the test 10x to make sure it was good and fixed.
Unwound the 15s shutdown timeout and ran the test again 10x to make sure it was still good.
Unwound the 30s random processCommand timeout and all debug logging and ran the test again 10x to make sure it was still good. Interestingly, this one had a shutdown failure in one of the follower nodes on the 7th test pass, but it was in a different part of the code, so I believe it's distinct, and a repeat of the tests passed all 10x.

Had issues with BadCommandTest and StatusHandlingCommandsTest failing intermittently. Most recent test I did 10x with "-only BadCommand,StatusHandlingCommands" and "-repeatCount 10", so 100 total runs, with no issues. Testing with full clustertest again now.
_________

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -816,7 +816,7 @@ void BedrockServer::worker(int threadId)
                 continue;
             }
 
-            // We just spin until the node looks ready to go. Typically, this doesn't happen expect briefly at startup.
+            // We just spin until the node looks ready to go. Typically, this doesn't happen expect briefly at startup or shutdown.
             while (_upgradeInProgress ||
                    (_replicationState.load() != SQLiteNode::LEADING &&
                     _replicationState.load() != SQLiteNode::FOLLOWING &&
@@ -824,7 +824,7 @@ void BedrockServer::worker(int threadId)
             ) {
                 // Make sure that the node isn't shutting down, leaving us in an endless loop.
                 if (_shutdownState.load() != RUNNING) {
-                    SWARN("Sync thread shut down while were waiting for it to come up. Discarding command '"
+                    SWARN("Got a command but node is shutting down. Discarding command '"
                           << command->request.methodLine << "'.");
                     return;
                 }
@@ -1079,10 +1079,10 @@ void BedrockServer::worker(int threadId)
                         // Nothing to do in this case, `command->complete` will be set and we'll finish as we fall out
                         // of this block.
                     } else if (result == BedrockCore::RESULT::SERVER_NOT_LEADING) {
-                        // We won't write regardless.
+                        // Server not leading, we won't write regardless.
                         core.rollback();
 
-                        // If there are no HTTPS requests, we can just re-queue this command, otherwise, we will
+                        // If there are outstanding HTTPS requests, we can't just re-queue this command, we would
                         // potentially run the same HTTPS requests twice.
                         if (command->httpsRequests.size()) {
                             SALERT("Server stopped leading while running command with HTTPS requests!");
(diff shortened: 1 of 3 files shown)


righdforsa:
Since BadCommandTest failed twice, and StatusHandlingCommands failed in Travis, (and locally with some prodding) and both of them failed in ASSERT statements regarding SQLiteNode states (e.g. LEADING, FOLLOWING, STANDINGDOWN) I'm going to take these as "things that need to be identified or fixed to move forward". Working on them.

BedrockServer.cpp
```diff
@@ -1079,20 +1079,28 @@ void BedrockServer::worker(int threadId)
...
+                                SINFO("State changed before 'processCommand' and we're shutting down. Abandoning " << command->request.methodLine << ".");
+                                command->response.methodLine = "500 Leader stopped leading";
+                                command->complete = true;
+                            // If we're not shutting down, we can likely recover from the current state changes.
                             // Allow for an extra retry and start from the top, like with ABANDONED_FOR_CHECKPOINT.
-                            SINFO("State changed before 'processCommand' but no HTTPS requests so retrying.");
-                            ++retry;
+                            } else {
+                                SINFO("State changed before 'processCommand' but no HTTPS requests so retrying.");
+                                ++retry;
```
flodnv, line 1102: Discussed about this with @righdforsa:

Why would we keep this `++retry;` here -- in which situation would retrying infinitely (see new line 1125) would ever benefit us? This log line hasn't been logged in over 1 week. Can we not just give this very rare flow the regular amount of retries @tylerkaraszewski ? `++retry` in `ABANDONED_FOR_CHECKPOINT` conceptually makes theoretical sense, but I just can't make any sense of it here.

tylerkaraszewski: We did not count any of the exception cases (like this one, the state changed while we were running the command) toward the retry count. The purpose of the retry count is to avoid conflicts, when we run out of retries, we push the command onto the blocking commit queue. Why should this state change affect the conflict-handling behavior?

Yes, we could remove it and it would likely make little difference but we could also leave it and it would likely make little difference because it's a weird exception handling case that almost never happens.

Sending the command to the blocking commit queue is actually really weird if we have just changed from LEADING to FOLLOWING (I think it will actually still work, though).

flodnv: >we could also leave it and it would likely make little difference because it's a weird exception handling case that almost never happens.

Right, but my point is that this can result, and has resulted, in infinite loops. So if we can just avoid this possibility, we're not really making anything worse, are we?

tylerkaraszewski: Are we sure this in particular has resulted in an infinite loop? I'm not actually sure how that would happen.

flodnv: @righdforsa seemed sure when he explained it to me yesterday 😅

tylerkaraszewski: I did not hear that explanation. :)

righdforsa: > Are we sure this in particular has resulted in an infinite loop? I'm not actually sure how that would happen.

That's the whole point of the ticket that spawned this PR. When we got into shutdown, there was a command running that happened to take just long enough that it began processing before the leader node got past standingdown but when "processCommand()" hit the "is this node leading?" block, it was no longer leading, and returned a "not leading" result. Because of the not leading result, we hit this retry increment, and the command looped over and over and never recovered, which caused the sync thread to not be able to shut down, which left the node port open, which caused peers to think it was still available, and caused a 5 minute site outage.

righdforsa: Now that that condition is handled (in this PR) in order to get there again, we'd have to disconnect from the cluster but not be shutting down. I don't know if we want to loop endlessly in that case.

tylerkaraszewski: > Because of the not leading result, we hit this retry increment, and the command looped over and over and never recovered

Ah, looking at it now, I see how it could do that. That's not how it's intended to work. The reason it gets stuck in the loop is because it doesn't notice it's not leading at the top of the loop.

This block of code determines if this command can write:
https://github.com/Expensify/Bedrock/blob/master/BedrockServer.cpp#L924-L946

But it's only run before the retry loop.

We could move it inside the loop, or we could just set `canWriteParallel` to false when we notice we're no longer leading.

flodnv:
bump @righdforsa ?

righdforsa:
I'm actively working on getting this up to date with the current state of
the Bedrock repo. Unfortunately, I've been having some stability problems
with the clustertest tests that I've been trying to resolve. This fix
works, but it's not totally clear to me what is being proposed as the
alternative approach, so maybe it's better.

PR #1157, opened 2022-01-12
Revert "Making `poll` runnable per command instead of per httpsManager."

flodnv:
Reverts Expensify/Bedrock#1108


botify:
**Beware!**

flodnv:
cc @rafecolton

righdforsa (approved):
👍

justinpersaud (approved):

PR #1163, opened 2022-01-14
Parallelone

rafecolton:
### Fixed Issues
Fixes GH_LINK

### Tests

_________


botify:
**Beware!**

rafecolton:
Closing this as it was just a POC. Will reuse some of this code later, I'm sure.

PR #1181, opened 2022-01-31
Fix nextRun < created - Flakey tests

aldo-expensify:
### Details

Fix rare case when `job.nextRun` can be one second earlier than `job.created`, which make some tests fail.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/193740

### Tests

Make sure that tests run
_________

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -500,8 +500,10 @@ void BedrockJobsCommand::process(SQLite& db) {
                 SINFO("Job specified run time or repeat, not suitable for immediate scheduling.");
             }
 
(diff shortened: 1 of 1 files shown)


francoisl (approved):
👍 makes sense.

For additional context on the change, we can say that this ensures that the `created` and `nextRun` are equal, which was failing sometimes [here](https://github.com/Expensify/Bedrock/blob/4c385d3b095b18859ab3d1b0bd4b8ec7edbe94a3/test/tests/jobs/CreateJobTest.cpp#L59).

PR #1187, opened 2022-02-05
Revert "Remove blocking commit queue"

righdforsa:
Reverts Expensify/Bedrock#1177 - we've been having crashes in the webrock cluster all night


botify:
**Beware!**

tylerkaraszewski (approved):
Noting that this is not a guaranteed fix, but it based on a correlation between this change and an increased TCP established on webrock.

tylerkaraszewski:
Seems to have fixed it. I'll investigate why tomorrow.
Screen Shot 2022-02-06 at 11 04 51 AM

PR #1211, opened 2022-03-07
Handle jobs with improperly formatted times

MariaHCD:
### Details
When retrieving a job, if the retryAfter update query fails for whatever reason, mark that job as failed and log a BugBot

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/199535

### Tests
New test case in GetJobTest.cpp
_________

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -852,18 +852,13 @@ void BedrockJobsCommand::process(SQLite& db) {
                                      "WHERE jobID = " + SQ(job["jobID"]) + ";";
 
                 try {
-                    db.writeIdempotent(updateQuery);
-                } catch (const SQLite::constraint_error& e) {
-                    // Something went wrong with the query to update the retryAfter,
-                    // Let's not throw an exception here but instead update the job to failed and log a Bugbot.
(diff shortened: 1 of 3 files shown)


plugins/Jobs.cpp
```diff
@@ -1419,6 +1426,15 @@ void BedrockJobsCommand::_validatePriority(const int64_t priority) {
     }
 }
 
+void BedrockJobsCommand::_handleFailedRetryAfterQuery(SQLite& db, const string& jobID) {
+    SALERT("ENSURE_BUGBOT Query error when updating job with retryAfter");
```
alex-mechler, line 1430: NAB: Adding the jobID to this line would make it slightly quicker for the person looking at this blocker to figure out the jobID so they can be debugging.

plugins/Jobs.cpp
```diff
@@ -1447,4 +1463,4 @@ void BedrockJobsCommand::handleFailedReply() {
         cmd->initiatingClientID = -1;
         _plugin->server.acceptCommand(move(cmd));
     }
-}
+}
```
alex-mechler, line 1466: NAB: bring the newline back

alex-mechler (approved):

tylerkaraszewski (approved):

PR #1328, opened 2022-07-20
Add a Travis check to ensure we remember to reset SQLITE_MAX_PAGE_COUNT

rafecolton:
cc @bondydaa @tylerkaraszewski 

### Details
If we update sqlite3.c, this value won't be updated in the mainline, so we need to ensure it gets updated or else we will run into problems.

### Fixed Issues
Part of https://github.com/Expensify/Expensify/issues/219373

### Tests
Added a check to the Travis tests, they should pass.
_________

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -57,6 +57,23 @@ travis_fold() {
   echo -en "travis_fold:${action}:${name}\r${ANSI_CLEAR}"
 }
 
+enforce_max_page_count() {
+    local fold_name="enforce_max_page_count"
+    travis_fold start $fold_name
+    travis_time_start
+    local intended_max_page_count="4294967295"
+    local max_page_count="$(grep -E 'define SQLITE_MAX_PAGE_COUNT [0-9]+' libstuff/sqlite3.c | awk '{print $4}')"
+    if [[ "$max_page_count" != "$intended_max_page_count" ]] ; then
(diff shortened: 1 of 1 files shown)


bondydaa (approved):
looks like it worked https://app.travis-ci.com/github/Expensify/Bedrock/builds/253436994

Do we care that there's no output during successful runs? I think it's fine so approving.
image

rafecolton:
I did notice that but I don't think it's a big deal. There will be output if it fails.

tylerkaraszewski:
I was thinking that instead of having travis fail if this is wrong, we'd make a build script for sqlite3.c that always does it correctly, and use that to generate our sqlite files.

rafecolton:
@tylerkaraszewski it does that in the build script on Jenkins for the CLI (separate PR), but building it for Bedrock here is somewhat manual. Until we script that part too, I added it to the SO instructions, which is hopefully sufficient. Added this check here because we all know nobody reads.

rafecolton:
I see your other comment about building it with that flag in the first place. I agree that would be better.

Gonals (approved):

rafecolton:
@tylerkaraszewski I meant to put this on hold but didn't get to it fast enough. Do you think we should revert and include it in the build flags instead?

bondydaa:
I think that'd be worthwhile too but also think this is a nice easy win to
get in to ensure that we don't screw this up while getting the build script
figured out.

PR #1332, opened 2022-07-27
Set SQLITE_MAX_PAGE_COUNT using build flag instead of travis check

rafecolton:
Reverts Expensify/Bedrock#1328

Part of https://github.com/Expensify/Expensify/issues/219373

Tested by editing the code to print out the value of `PRAGMA max_page_count` on startup.


travis.sh
```diff
@@ -57,23 +57,6 @@ travis_fold() {
   echo -en "travis_fold:${action}:${name}\r${ANSI_CLEAR}"
 }
 
-enforce_max_page_count() {
```
bondydaa, line 60: would we still leave this check but instead have it run after we build and start bedrock and then run 
```
nc localhost 8888
Query: pragma max_page_count
```

and verify it's the correct number?

bondydaa (approved):
just one thing but not really a blocker for me

travis.sh
rafecolton: I don't see why we would, we don't verify that any of the other build flags worked as expected. I think if we find a pattern of this breaking or not working as expected, we can revisit.

bondydaa: fair enough but I think this not being set properly is probably a little more catastrophic than some others since it could prevent leader from writing any new transactions and basically grind all functionality to a halt. agree we can always add it later if that happens though.

PR #1355, opened 2022-08-24
Add milliseconds to all timestamps

roryabraham:
### Details
This adds milliseconds to almost all timestamps.

### Fixed Issues
This is necessary because we are adding a reportActionID column to the reportActions table as the primary key. However, there are many places where we do `ORDER BY ROWID`, where `ROWID` will no longer be sequential. Instead, we need to `ORDER BY created`. In order to make that more predictable/precise, we will add ms to the `created` timestamp.

While we _could_ do that in only one place, I opted for the slightly more drastic option of just adding ms everywhere ... because it _should_ be pretty safe to do?

### Tests
1. Recompile Bedrock and Auth
1. Run Auth tests
_________

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2905,7 +2905,7 @@ int SQuery(sqlite3* db, const char* e, const string& sql, int64_t warnThreshold,
 }
 
 string SUNQUOTED_TIMESTAMP(uint64_t when) {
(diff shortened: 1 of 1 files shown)


coleaeason:
IMO this seems like overkill and likely to break shit, I would guess we don't expect this anywhere else in the code today but we DO expect the current format

roryabraham:
Yeah, agree it's a bit extreme. Going to close this

PR #1377, opened 2022-09-20
Default _isCommandPortLikelyBlocked to false on init for broken tests

johnmlee101:
### Details
This was never initiated to a default value and bools are set to garbage by default. See 
https://app.travis-ci.com/github/Expensify/Auth/jobs/583377405 for proof 

This caused flakey parts of tests which became very apparent today
See this giant thread for context: https://expensify.slack.com/archives/C094TGUTZ/p1663671773822939

### Fixed Issues
Fixes auth tests not working on the 20.04.5 ubuntu version

### Tests

_________

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -1121,6 +1121,7 @@ BedrockServer::BedrockServer(const SData& args_)
     _lastQuorumCommandTime(STimeNow()), _pluginsDetached(false), _socketThreadNumber(0),
     _outstandingSocketThreads(0), _shouldBlockNewSocketThreads(false)
 {
+    _isCommandPortLikelyBlocked = false;
     _version = VERSION;
 
     // Enable the requested plugins, and update our version string if required.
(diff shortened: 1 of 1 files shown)


iwiznia (approved):

BedrockServer.cpp
```diff
@@ -1121,6 +1121,7 @@ BedrockServer::BedrockServer(const SData& args_)
     _lastQuorumCommandTime(STimeNow()), _pluginsDetached(false), _socketThreadNumber(0),
     _outstandingSocketThreads(0), _shouldBlockNewSocketThreads(false)
 {
+    _isCommandPortLikelyBlocked = false;
```
johnmlee101, line 1124: Does this need the mutex lock for the port threads? Probably not since we're initializing it.

johnmlee101:
@tylerkaraszewski whenever you get the chance to look at this

BedrockServer.cpp
johnmlee101: I tried to add this to the initializers above but it wouldn't compile, probably because its an atomic type

johnmlee101: actually I can, my b 1 sec

johnmlee101:
wait for this test branch to pass as well, which references this branch of Bedrock
https://app.travis-ci.com/github/Expensify/Auth/jobs/583376505

johnmlee101:
Tests passed!

PauloGasparSv (approved):

tgolen (approved):

tylerkaraszewski (commented):
Thanks for fixing!

PR #1391, opened 2022-09-30
Tyler fix broken travis 2

tylerkaraszewski:
### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/231211

### Tests

_________


rafecolton (approved):
Approving since this fixes the underlying issue, but noting that using tags instead of releases could come back to bite us one day if we ever decide to use a tag for anything else.

PR #1412, opened 2022-12-20
Optionalize checkpoint mode feature

robertjchen:
### Details

Optionalize the changes from https://github.com/Expensify/Bedrock/pull/1403 as it may not be stable for certain specific workloads/use cases.

### Fixed Issues
Fixes GH_LINK

### Tests

_________


mountiny (approved):
This looks good to me @robertjchen However, I might be missing something, so to clarify, by default we set the optional flag to true and nowhere we set it to false. If we know these changes turned on everywhere cause issues, shouldnt we turn this off for some instances

Oh haha, I see it is set in the Salt PR, linking it here for posteriority - https://github.com/Expensify/Salt/pull/11746

PauloGasparSv (approved):

tylerkaraszewski (approved):

PR #1534, opened 2023-07-13
Add notifyPlugins method

iwiznia:
### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/299489

### Tests
None, but will start to work on the auth PR and we will be able to test this

_________


BedrockServer.h
```diff
@@ -497,4 +497,6 @@ class BedrockServer : public SQLiteServer {
     // syncNode while the sync thread exists, it's a shared pointer to allow for the last socket thread using it to
     // destroy the pool at shutdown.
     shared_ptr<SQLitePool> _dbPool;
+
+    void notifyPlugins(SQLiteNodeState newState);
```
danieldoglas, line 501:
```suggestion
    void notifyPlugins(SQLiteNodeState newState) override;
```

Making sure this won't compile if it's not overriding the original function

sqlitecluster/SQLiteNode.cpp
```diff
@@ -1809,6 +1809,7 @@ void SQLiteNode::_changeState(SQLiteNodeState newState) {
     _localCommitNotifier.notifyThrough(_db.getCommitCount());
 
     if (newState != _state) {
+//        _server.canStand
```
danieldoglas, line 1812:
```suggestion
```

sqlitecluster/SQLiteServer.h
```diff
@@ -14,4 +14,6 @@ class SQLiteServer : public STCPManager {
 
     // When a node connects to the cluster, this function will be called on the sync thread.
     virtual void onNodeLogin(SQLitePeer* peer) = 0;
+
```
danieldoglas, line 17: Please add a comment

sqlitecluster/SQLiteNode.cpp
danieldoglas: Also missing the call to notifyPlugins

danieldoglas (changes requested):

sqlitecluster/SQLiteNode.cpp
iwiznia: Yeah, this is not done yet 😄

iwiznia:
I was trying to add tests, but not sure how to... like how can I know that the callback was called?

iwiznia:
hmmmmm, I just realized that the `Auth` plugin does not really have a reference to the DB... I think I will need to modify this to also pass the db handle to this method so we can run queries.

iwiznia:
ok, added the db as param

BedrockServer.cpp
```diff
@@ -2356,3 +2356,9 @@ void BedrockServer::waitForHTTPS(unique_ptr<BedrockCommand>&& command) {
 const atomic<SQLiteNodeState>& BedrockServer::getState() const {
     return _nodeStateSnapshot == SQLiteNodeState::UNKNOWN ? _replicationState : _nodeStateSnapshot;
 }
+
+void BedrockServer::notifyStateChangeToPlugins(SQLite& db, SQLiteNodeState newState) {
+    for (auto plugin : plugins) {
+        plugin.second->stateChanged(db, newState);
+    }
+}
```
johnmlee101, line 2364: EOL

johnmlee101 (changes requested):
End of line is missing but otherwise looks good

BedrockServer.cpp
iwiznia: Why does that even matter?

johnmlee101: This is a standardization we apply across our entire codebase, and github shows this
image

johnmlee101: *EOF

iwiznia: > This is a standardization we apply across our entire codebase

Why? And why is it not handled by the linter?

> and github shows this

Who cares?

johnmlee101 (approved):

danieldoglas (approved):

sqlitecluster/SQLiteNode.cpp
```diff
@@ -1809,6 +1809,9 @@ void SQLiteNode::_changeState(SQLiteNodeState newState) {
     _localCommitNotifier.notifyThrough(_db.getCommitCount());
 
     if (newState != _state) {
+        // First, we notify all plugins about the state change
+        _server.notifyStateChangeToPlugins(_db, newState);
+
```
danieldoglas, line 1814: @iwiznia why did we notify the state change to the plugins before actually changing the state?

iwiznia: Hmmmm I don't know. Is it causing issues?

danieldoglas: Not exactly. I'm working on a PR that will need to check the state, but because we're notifying before changing the state, I'll have to check both the previous state and current state.

So I just wanted to understand why we did it this way. But I think it makes sense that we notify before we start processing things as leader, for example

iwiznia: Yeah for onyx updates we definitely want to do it before we start processing things. We could maybe move this to a bit down the method, though not sure if that would help you in any way for your issue

PR #1544, opened 2023-07-27
Give the plugin its own db handle for stateChanged

coleaeason:
### Details
Currently, for the `stateChanged` function, we pass the plugin the sync thread's database handle. This works most of the time however because there is no locking around this handle it is possible that a plugin and the sync thread try to operate on the database at the same time. This is not supported, and causes SQLite to report database corruption for the second caller. 

This change creates a new database handle specifically for the stateChanged handler, so that the sync thread may use its own handle and the plugin gets it's own as well. This should prevent the crashes we've seen. 

### Fixed Issues
Multiple crashes during deploys.

### Tests
The automated tests still pass. Since this bug is triggered by a race condition, I was unable to replicate the crashes we've seen in dev. However I'm reasonably sure this is the fix, and the downside of deploying this is essentially none. 
_________

### DIFF
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -129,6 +129,10 @@ SQLiteNode::SQLiteNode(SQLiteServer& server, shared_ptr<SQLitePool> dbPool, cons
 {
     SASSERT(_originalPriority >= 0);
     onPrepareHandlerEnabled = false;
+    
+    // We create a copy of the database handle here so that the sync node can operate on its handle and the plugin gets
+    // its own handle to operate on. This avoids conflicts where the sync thread and the plugin are trying to both run
+    // queries at the same time. This also avoids the need to create any share locking between the two.
     pluginDB = new SQLite(_db);
     SINFO("[NOTIFY] setting commit count to: " << _db.getCommitCount());
     _localCommitNotifier.notifyThrough(_db.getCommitCount());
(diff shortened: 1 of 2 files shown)


danieldoglas (approved):
LGTM

sqlitecluster/SQLiteNode.cpp
```diff
@@ -1815,7 +1824,7 @@ void SQLiteNode::_changeState(SQLiteNodeState newState) {
 
     if (newState != _state) {
         // First, we notify all plugins about the state change
-        _server.notifyStateChangeToPlugins(_db, newState);
+        _server.notifyStateChangeToPlugins(*pluginDB, newState);
```
joelbettner, line 1827: I think we should move the creation of `pluginDB` into this conditional statement, that way we only create it when we need it.

```cpp
if (newState != _state) {
    // First, we notify all plugins about the state change
        // We create a copy of the database handle here so that the sync node can operate on its handle and the plugin gets
    // its own handle to operate on. This avoids conflicts where the sync thread and the plugin are trying to both run
    // queries at the same time. This also avoids the need to create any share locking between the two.
    pluginDB = new SQLite(_db);
    _server.notifyStateChangeToPlugins(*pluginDB, newState);
```

joelbettner (changes requested):

sqlitecluster/SQLiteNode.cpp
coleaeason: I explained this in slack but will post here for everyone's benefit:

I originally had it there but I moved it to make the lifecycle line up with the life cycle of the SQLiteNode object since I am deleting it in the deconstructor. Additionally it actually doesn’t matter, because functionally it it the same thing. We change the state immediately upon bring a node up. So with the way I have it now, the object probably exists milliseconds longer.

bondydaa (approved):

sqlitecluster/SQLiteNode.cpp
joelbettner: Discussed here: https://expensify.slack.com/archives/C094TGUTZ/p1690480820049479

joelbettner (approved):

PR #1553, opened 2023-07-31
Revert "Use a handle from the existing pool instead of a new one"

coleaeason:
Reverts Expensify/Bedrock#1551

This is broken when used with Auth, I'll try to figure out a fix, but for now let's just revert so we can deploy. The old code worked fine.


johnmlee101 (approved):

johnmlee101:
Straight revert, merged without tests

PR #1615, opened 2023-11-13
Support view in verifyTable

iwiznia:
### Fixed Issues
Related https://github.com/Expensify/Expensify/issues/336109

### Tests
Start auth and restart auth, check the view is correctly created
_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -385,13 +385,13 @@ bool SQLite::beginTransaction(TRANSACTION_TYPE type) {
     return _insideTransaction;
 }
 
-bool SQLite::verifyTable(const string& tableName, const string& sql, bool& created) {
+bool SQLite::verifyTable(const string& tableName, const string& sql, bool& created, const string& type) {
     // sqlite trims semicolon, so let's not supply it else we get confused later
     SASSERT(!SEndsWith(sql, ";"));
 
     // First, see if it's there
     SQResult result;
-    SASSERT(read("SELECT sql FROM sqlite_master WHERE type='table' AND tbl_name=" + SQ(tableName) + ";", result));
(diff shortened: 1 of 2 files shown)


sqlitecluster/SQLite.cpp
```diff
@@ -385,13 +385,13 @@ bool SQLite::beginTransaction(TRANSACTION_TYPE type) {
     return _insideTransaction;
 }
 
-bool SQLite::verifyTable(const string& tableName, const string& sql, bool& created) {
+bool SQLite::verifyTable(const string& tableName, const string& sql, bool& created, const string& type) {
```
danieldoglas, line 388: Should we maybe create a verifyView instead? We have verifyIndex as well, so I think that keeping the types separate would be good. Then separate the code into a third abstract method that received the type. To simplify the call..

iwiznia: verifyIndex has specific code for indexes, but in this case views and tables are the same and the code would be identical. I could create a `verifyView` and leave `verifyTable` and both could call another method but then what would method be called? And is it really better than having this?

danieldoglas: I feel like for consistency, it would be better to have specific methods. It's kind of confusing to say `verifyTable of type view`. But then, views are indeed virtual tables, so maybe we don't really care about that?

danieldoglas: I won't block on this, since I think it's not super important, it's just a matter of definition. If we consider that virtual tables are the same as tables, verifyTable type view is not a problem. Approving the PR

danieldoglas (approved):

sqlitecluster/SQLite.cpp
iwiznia: Cool, I'll take it because views are indeed tables and because I would not know how to name the methods otherwise 😄

PR #1697, opened 2024-04-16
Move AutoTimer after transaction

aldo-expensify:
### Details

If a command is going to run `peek` in blocking commit, and it has to wait for other commands that are already running in blocking commit, that wait time was being added to the `peek` time, which makes it look like this command used more time than it really used in the blocking commit thread.

See https://expensify.slack.com/archives/C05CBC62HGW/p1712357941670379
Another examples found here: https://github.com/Expensify/Expensify/issues/387423#issuecomment-2057690158

I want to fix this because this error in our performance measuring tools makes us focus on things that are not really a problem.

### Fixed Issues
Fixes https://expensify.slack.com/archives/C05CBC62HGW/p1712357941670379

### Tests

N/A

_________

### DIFF
--- a/BedrockCore.cpp
+++ b/BedrockCore.cpp
@@ -145,7 +145,7 @@ BedrockCore::RESULT BedrockCore::peekCommand(unique_ptr<BedrockCommand>& command
                 STHROW("501 Failed to begin " + (exclusive ? "exclusive"s : "shared"s) + " transaction");
             }
 
-            // We start the timer here to avoid including the time spent acquiring the the lock _sharedData.commitLock
+            // We start the timer here to avoid including the time spent acquiring the lock _sharedData.commitLock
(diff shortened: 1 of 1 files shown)


aldo-expensify:
cc @tylerkaraszewski @flodnv in case you see something wrong with this

flodnv (approved):
Nice, thanks!

tylerkaraszewski (commented):
This only times these lines now: https://github.com/Expensify/Bedrock/blob/e2b984b21dfbdd79b8c1654ad429214948e65713/BedrockCore.cpp#L151-L164

I mean, that's *most* of peek, but are you sure that's what we want?

aldo-expensify:
> I mean, that's most of peek, but are you sure that's what we want?

An alternative could be to measure the time acquiring the lock (beginTransaction) and subtract that from the peek time as we were measuring it before. We could subtract it when we are processing the timing info. That would make it include the rest of the code that now we are missing.

Not sure if there is value in doing that though, the rest of the code looks like it should be fast stuff.

PR #1705, opened 2024-04-30
Add `SParseSet`

jasperhuangg:
SParseSet parses a comma separated string into a set of strings.

i.e. `"a", "b", "c"` becomes `<set<string>>{"a", "b", "c"}`

### Details
Needed for https://github.com/Expensify/Auth/pull/10560

### Fixed Issues
Related to https://github.com/Expensify/App/issues/35240

### Tests
Tested with https://github.com/Expensify/Auth/pull/10560, I compiled it against my changes, and the automated tests in that PR call the new code added here.

_________

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -702,7 +702,7 @@ bool SParseList(const char* ptr, list<string>& valueList, char separator) {
 }
 
 // --------------------------------------------------------------------------
-set<int64_t> SParseSet(const string& value, char separator) {
+set<string> SParseSet(const string& value, char separator) {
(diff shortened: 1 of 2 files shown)


neil-marcellini (approved):
Looks fine to me, and I trust your tests on the Auth PR.

bondydaa:
as noted in slack, this seems unnecessary to me you can easily take a list and create a set from it (examples here https://www.tutorialspoint.com/cplusplus-program-to-convert-list-to-set)

so why not just do that directly in your code? it's not like this saves having to call `SParseList` it just does it for you.

PR #1738, opened 2024-05-17
[WIP] Allow testing a single test method from a test fixture

marcaaron:
### Details

- This gives us a way to test a specific method in a test case.
- Pretty rough PoC for now just to prove it works to see if we might want this
- I am assuming we are gonna be perpetually maintaining our fork of tpunit in this repo and this is a reasonable place to raise this change.

cc @youssef-lr

### Fixed Issues

Related [convo](https://expensify.slack.com/archives/C03TQ48KC/p1715964660816439)

### Tests

1. Build Auth 
2. Run a test like `./test/authtest -only RemoveFromGroupChat -includeMethods testCommand`

Verify only the one test method runs:

_________


marcaaron:
not a priority saving it for a rainy day.

PR #1760, opened 2024-05-31
Move to GH Actions

johnmlee101:
Need to test releases, but that requires this be merged.


.github/workflows/bedrock.yml
```diff
@@ -0,0 +1,90 @@
...
+        tar -xf ccache-4.8.3-linux-x86_64.tar.xz
+        sudo cp ccache-4.8.3-linux-x86_64/ccache /usr/local/bin/ccache
+        ccache -V
+        ln -s /usr/local/bin/ccache /usr/local/bin/gcc-13
+        ln -s /usr/local/bin/ccache /usr/local/bin/g++-13
+
+    - name: Setup tmate session
+      if: runner.debug == '1'
+      uses: mxschmitt/action-tmate@v3
+      timeout-minutes: 60
```
AndrewGable, line 70: This seems kinda long for a big runner, but maybe we rarely debug?

.github/workflows/bedrock.yml
```diff
@@ -0,0 +1,90 @@
...
+      run: "./ci_tests.sh"
+
+    - uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # v0.1.15
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        GITHUB_REPOSITORY: "${{ github.repository }}"
+      with:
+        files: |-
+          ./bedrock
+      if: "${{ startsWith(github.ref, 'refs/tags/') }}"
```
AndrewGable, line 89: NAB - Usually see `if` at the top

.github/workflows/bedrock.yml
johnmlee101: yeah we rarely debug, and I haven't seen anyone stretch this as of yet

.github/workflows/bedrock.yml
```diff
@@ -0,0 +1,90 @@
+name: Bedrock Test Suite
+on:
+  push:
+    branches: # this ignores tag pushes, and only looks at branches.
+      - '**'
+  release:
+    types: [prereleased]
```
johnmlee101, line 7: This is because we only trigger pre-releases here: https://github.com/Expensify/Bedrock/releases

AndrewGable (approved):

PR #1806, opened 2024-07-11
Use mold linker in GH actions

flodnv:
### Fixed Issues
For https://github.com/Expensify/Expensify/issues/409979

### DIFF
--- a/.github/workflows/bedrock.yml
+++ b/.github/workflows/bedrock.yml
@@ -19,6 +19,7 @@ jobs:
     timeout-minutes: 30
     steps:
 
+    - uses: rui314/setup-mold@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/bedrock.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     timeout-minutes: 30
     steps:
 
+    - uses: rui314/setup-mold@v1
```
johnmlee101, line 22:
```suggestion
    - name: Install the Mold Linker
      uses: rui314/setup-mold@v1
      
```

johnmlee101 (changes requested):

johnmlee101 (approved):

PR #1861, opened 2024-09-02
Add shouldGenerateCrashCommand

aldo-expensify:
### Details

Adds a function that commands can override to avoid generating a CRASH_COMMAND we don't want

This function will be overridden for `Get` here: https://github.com/Expensify/Auth/pull/12323

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/422785

### Tests

Setup: To force the two different crashes, I modify the code as follows:

In Bedrock `BedrockCore::peekCommand`:

image

In Auth `Get::peek`:

image

1. Connect to bedrock using `nc auth 4444`
2. Send a command to crash it before validating the authToken:
    ```
    Get
    authToken:<autoToken>
    doubleTransaction:true
    ```
3. Look at the logs and verify that a `CRASH_COMMAND` **was not** created
4. Restart bedrock/auth
5. Connect to bedrock using `nc auth 4444`
2. Send a command to crash it after validating the authToken (inside Get::peek):
    ```
    Get
    authToken:<autoToken>
    forceSegFault:true
    ```
3. Look at the logs and verify that a `CRASH_COMMAND` **was** created, and that it has an `accountID=xxxx`

_________

### DIFF
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -389,3 +389,7 @@ string BedrockCommand::serializeData() const {
 
 void BedrockCommand::deserializeData(const string& data) {
 }
+
+bool BedrockCommand::shouldGenerateCrashCommand() {
+    return true;
+}
--- a/BedrockCommand.h
+++ b/BedrockCommand.h
@@ -187,6 +187,8 @@ class BedrockCommand : public SQLiteCommand {
     };
     CrashMap crashIdentifyingValues;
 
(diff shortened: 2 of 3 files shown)


aldo-expensify:
cc @tylerkaraszewski (in case you think there is a better approach)

tylerkaraszewski (changes requested):
I don't think we need this. I will comment on the auth PR.

aldo-expensify:
Changing approach to using IP

PR #1878, opened 2024-09-24
Redact the "edits" field in any logged queries

marcaaron:
### Details

Discovered while investigating: https://github.com/Expensify/Expensify/issues/381602

We are logging deleted report comments when people delete them.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/430811

### Tests

Apply this diff since the queries need to be over 10ms to show on dev and probably won't be

```diff
diff --git a/libstuff/libstuff.cpp b/libstuff/libstuff.cpp
index 7275931f..56b27131 100644
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2667,7 +2667,7 @@ int SQuery(sqlite3* db, const char* e, const string& sql, SQResult& result, int6
     }

     uint64_t elapsed = STimeNow() - startTime;
-    if (!skipInfoWarn && ((int64_t)elapsed > warnThreshold || (int64_t)elapsed > 10000)) {
+    if (!skipInfoWarn && true) {
         // Avoid logging queries so long that we need dozens of lines to log them.
         string sqlToLog = sql.substr(0, 20000);
         SRedactSensitiveValues(sqlToLog);
```

1. Compile Auth
2. Update a comment
3. Tail logs for `UPDATE reportActions SET message =`
4. Verify that "edits" says `edits: ["REDACTED"]`
_________

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2913,6 +2913,9 @@ void SRedactSensitiveValues(string& s) {
 
     // Remove anything inside "html" because we intentionally don't log chats.
     s = SREReplace("\"html\":\".*\"", s, "\"html\":\"<REDACTED>\"");
+
+    // Remove anything inside "edits" because these are also chats.
+    s = SREReplace("\"edits\":\".*\"", s, "\"edits\":\"<REDACTED>\"");
 }
 
 SStopwatch::SStopwatch() {
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2915,7 +2915,7 @@ void SRedactSensitiveValues(string& s) {
     s = SREReplace("\"html\":\".*\"", s, "\"html\":\"<REDACTED>\"");
 
     // Remove anything inside "edits" because these are also chats.
-    s = SREReplace("\"edits\":\\[\".*?\"\\]", s, "\"edits\":[\"REDACTED\"]");
+    s = SREReplace(R"(\"edits\":\[.*?\])", s, "\"edits\":[\"REDACTED\"]");
 }
 
 SStopwatch::SStopwatch() {
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2915,7 +2915,7 @@ void SRedactSensitiveValues(string& s) {
     s = SREReplace("\"html\":\".*\"", s, "\"html\":\"<REDACTED>\"");
 
     // Remove anything inside "edits" because these are also chats.
-    s = SREReplace("\"edits\":\".*\"", s, "\"edits\":\"<REDACTED>\"");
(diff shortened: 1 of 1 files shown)


libstuff/libstuff.cpp
```diff
@@ -2913,6 +2913,9 @@ void SRedactSensitiveValues(string& s) {
 
     // Remove anything inside "html" because we intentionally don't log chats.
     s = SREReplace("\"html\":\".*\"", s, "\"html\":\"<REDACTED>\"");
+
+    // Remove anything inside "edits" because these are also chats.
+    s = SREReplace("\"edits\":\\[\".*?\"\\]", s, "\"edits\":[\"REDACTED\"]");
```
Beamanator, line 2918: I coulddddd be wrong, BUT regex 101 is telling me:

Screenshot 2024-09-24 at 8 38 02 PM

The weird parts to me are:

1. `\\ matches the character \ with index 9210 (5C16 or 1348) literally (case sensitive)`
    - I don't think normally arrays in json look like `"edits":\...` right?
2. `Match a single character present in the list below [\".*?\"\\]`
    - I thinkkk you were going for trying to match the `[` character, then stuff inside of it right? It looks like this is doing the standard regex character grouping match thing, not just matching `[` (the start of an array) right?

Please tell me if I'm confused or a genius

Beamanator (changes requested):

libstuff/libstuff.cpp
marcaaron: I see your confusion. I think it will click after I lay these points out for ya...

- We can't just pass `[*.]` as a regex string because brackets need to be escaped in regex if you are matching on them literally (otherwise they are special characters)
- a `\` character is _already_ an escape character as far as C++ strings go
- So, we have to escape twice.

That would get you something like this:



Maybe it would be better to use raw string literal? Then we would not need to double escape the `[` character only single escape and then it would be less weird or something?

marcaaron: I think we can get rid of the quotes though and maybe simplify it to:

```c++
R"(\"edits\":\[.*?\])"
```

marcaaron: That worked too so I pushed it.

flodnv (approved):

Beamanator (approved):
ooh that looooooks much more like what i'd expect, thanks for the update! 🙏 

NAB - do we have any tests for redacting stuff? If so, maybe we could add one for this new change as a follow-up?

libstuff/libstuff.cpp
Beamanator: loveeeeeein that version, thanks for updating - looks much clearer to me!

marcaaron:
>do we have any tests for redacting stuff?

I could not find any.

>maybe we could add one for this new change as a follow-up?

I like that idea.

PR #1882, opened 2024-09-30
Removing conflict check when the last conflict happened on a journal

danieldoglas:
### Details

The normal flow for a write command is to try to commit it. If that fails because of a conflict, we will retry the commits three times before sending them to the blocking queue.

When we have a conflict on a page that belongs to a table, that means that some other command is also changing that same page.  We'll store the page that conflicted:

https://github.com/Expensify/Bedrock/blob/67b55ff5c9da80d1452d5e36d60da88226323028/sqlitecluster/SQLite.cpp#L286

And then we'll wait until that page is not locked anymore before we try to execute the command again:

https://github.com/Expensify/Bedrock/blob/67b55ff5c9da80d1452d5e36d60da88226323028/BedrockServer.cpp#L910-L913

When that happens for any tables in the system, it is not a problem. But when we have a conflict that happened on a journal table, then we're blocking the next commit from happening without reason. The journals are chosen when the commit is being executed, and the chances that we select the same journal on the next commit are low.  that considered, we don't need to do a PageLockGuard if the previous conflict was on a journal table.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/432114

### Tests

Ran the ConflictSpam test and confirmed that the changed log line is also logging the tables:

```
2024-09-30T18:35:02.974407+00:00 expensidev2004 bedrock10019: xxxxxx (SQLite.cpp:757) commit [replicate517] [info] part of last conflict page: 31, conflict table: test
```

then I applied these changes to make the cluster use page conflicts

```
diff --git a/BedrockServer.cpp b/BedrockServer.cpp
index 544d3e5b..59469959 100644
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -1029,6 +1029,9 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
                                 // don't need to lock our next commit on this page conflict.
                                 if (!SStartsWith(lastConflictTable, "journal")) {
                                     lastConflictPage = db.getLastConflictPage();
+                                    SINFO(format("Not skipping lock on journal table {}", lastConflictTable));
+                                } else {
+                                    SINFO(format("Skipping lock on journal table {}", lastConflictTable));
                                 }
                             }
                         }
diff --git a/test/clustertest/BedrockClusterTester.h b/test/clustertest/BedrockClusterTester.h
index d8d11bc4..fb8dbe54 100644
--- a/test/clustertest/BedrockClusterTester.h
+++ b/test/clustertest/BedrockClusterTester.h
@@ -129,6 +129,7 @@ ClusterTester<T>::ClusterTester(ClusterSize size,
             {"-peerList", peerString},
             {"-plugins", pluginsToLoad},
             {"-overrideProcessName", "bedrock" + to_string(nodePort)},
+            {"-enableConflictPageLocks", "true"},
         };
 
         // Now, if any args were supplied, we'll use those instead of these ones (note that this overwrites our
```

And confirmed that the logs were successfully doing what we wanted them to do:

```
2024-09-30T19:04:37.866701+00:00 expensidev2004 bedrock10013: xGhV1t (BedrockServer.cpp:1032) runCommand [socket27] [info] Not skipping lock on journal table test
2024-09-30T19:04:37.866977+00:00 expensidev2004 bedrock10013: QZeN1R (BedrockServer.cpp:1034) runCommand [socket32] [info] Skipping lock on journal table journal0004
2024-09-30T19:04:37.867260+00:00 expensidev2004 bedrock10013: DHoH5U (BedrockServer.cpp:1032) runCommand [socket5] [info] Not skipping lock on journal table test
```
_________


sqlitecluster/SQLite.cpp
```diff
@@ -284,6 +285,20 @@ void SQLite::_sqliteLogCallback(void* pArg, int iErrCode, const char* zMsg) {
...
+        // Check if the tableOffset exists since not all conflicts are on tables
+        if (tableOffset) {
+            // Based on the SQLite log line, we should always have ';' after the table name,
+            // so let's use it to finish this loop.
+            int i = 0;
+            while (tableOffset[i] != ';') {
+                _conflictTable += tableOffset[i];
+                i++;
+            }
+        }
```
danieldoglas, line 301: I've confirmed this works as expected:

image

danieldoglas: I'm using this log line as the base for this logic: https://github.com/Expensify/Bedrock/blob/676f804ed01ac09ffaedb2f3732069f0bbf50500/libstuff/sqlite3.c#L84720

sqlitecluster/SQLite.cpp
```diff
@@ -284,6 +285,20 @@ void SQLite::_sqliteLogCallback(void* pArg, int iErrCode, const char* zMsg) {
         // 17 is the length of "conflict at page" and the following space.
         const char* offset = strstr(zMsg, "conflict at page") + 17;
         _conflictPage = atol(offset);
+
+        // 17 is the length of "part of db table" and the following space.
+        const char* tableOffset = strstr(zMsg, "part of db table") + 17;
```
tylerkaraszewski, line 290: You can start from `offset`, we already know `tablerOffset` will be after that.

Also, renaming `offset` to `pageOffset` might be clearer.

BedrockServer.cpp
```diff
@@ -1021,7 +1022,10 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
                         } else {
                             SINFO("Conflict or state change committing " << command->request.methodLine << " on worker thread.");
                             if (_enableConflictPageLocks) {
-                                lastConflictPage = db.getLastConflictPage();
+                                lastConflictTable = db.getLastConflictTable();
+                                if (!SStartsWith(lastConflictTable, "journal")) {
```
flodnv, line 1026: Maybe leave a comment explaining why we're doing this?

sqlitecluster/SQLite.cpp
```diff
@@ -284,6 +285,20 @@ void SQLite::_sqliteLogCallback(void* pArg, int iErrCode, const char* zMsg) {
...
+
+        // Check if the tableOffset exists since not all conflicts are on tables
+        if (tableOffset) {
+            // Based on the SQLite log line, we should always have ';' after the table name,
+            // so let's use it to finish this loop.
+            int i = 0;
+            while (tableOffset[i] != ';') {
+                _conflictTable += tableOffset[i];
+                i++;
+            }
```
tylerkaraszewski, line 300: Instead of this loop, just call `strstr` again, which will check that we actually see the `;`.

Then you can set `_conflictTable` with constructor #8 here: https://en.cppreference.com/w/cpp/string/basic_string/basic_string by doing something like:
```
_conflictTable = string(tableOffset, semicolonOffset - tableOffset)
```

(might need a ` - 1` at the end?)

flodnv (changes requested):

BedrockServer.cpp
danieldoglas: Done

sqlitecluster/SQLite.cpp
```diff
@@ -282,8 +284,26 @@ void SQLite::_sqliteLogCallback(void* pArg, int iErrCode, const char* zMsg) {
...
+
+        // Check if the tableOffset exists since not all conflicts are on tables
+        if (tableOffset) {
+            // Based on the SQLite log line, we should always have ';' after the table name,
+            // so let's find it and use it to limit the size of the substring we need
+            const char* semicolonOffset = strstr(tableOffset, ";");
+            
+            // Let's add this check in case the SQLite log changes and we don't notice it
+            // since this would generate a runtime error.
+            if (semicolonOffset) {
```
danieldoglas, line 301: added this IF since the string initializer would generate a runtime error in case semicollongOffset was null: 
image

sqlitecluster/SQLite.cpp
danieldoglas: Thanks for the suggestion! Changed, and therés no need for the `-1`: 
image

sqlitecluster/SQLite.cpp
danieldoglas: Done

sqlitecluster/SQLite.cpp
```diff
@@ -282,8 +284,26 @@ void SQLite::_sqliteLogCallback(void* pArg, int iErrCode, const char* zMsg) {
     // This is sort of hacky to parse this from the logging info. If it works we could ask sqlite for a better interface to get this info.
     if (SStartsWith(zMsg, "cannot commit")) {
         // 17 is the length of "conflict at page" and the following space.
-        const char* offset = strstr(zMsg, "conflict at page") + 17;
-        _conflictPage = atol(offset);
+        const char* pageOffset = strstr(zMsg, "conflict at page") + 17;
```
tylerkaraszewski, line 287: Start at `offset` rather than `zMsg`

sqlitecluster/SQLite.cpp
tylerkaraszewski: I don't understand this comment. You need the `if`, yes, because otherwise `semicolonOffset - tableOffset` will be negative.

The error you posted looks like you were setting `msg` by hand, and that's the error you're seeing.

tylerkaraszewski (changes requested):
One more suggested change.

sqlitecluster/SQLite.cpp
danieldoglas: am I missing something here? This was the previous offset property, so we don't have an offset to use on `pageOffset`. 

`tableOffset` below is starting on `pageOffset` already, if that's what you mean

tylerkaraszewski: Oops, my mistake, I thought I was looking at the next line where we did this.

tylerkaraszewski (approved):

BedrockServer.cpp
```diff
@@ -778,6 +778,7 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
     }
 
     int64_t lastConflictPage = 0;
+    string lastConflictTable = "";
```
tylerkaraszewski, line 781: You don't need the ` = ""` here, it's the default.

tylerkaraszewski (approved):

tylerkaraszewski (approved):

BedrockServer.cpp
```diff
@@ -778,6 +778,7 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
     }
 
     int64_t lastConflictPage = 0;
+    string lastConflictTable;
```
flodnv, line 781: Why is this defined so far up?

flodnv (approved):

PR #1907, opened 2024-10-18
Correctly handle all 200 Spotnana responses

Gonals:
### Details
Spotnana is not sending `200 OK`, but `HTTP/1.1 200` (sometimes). We need to handle that correctly.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/431982

### Tests
Will be tested after deploy with real Spotnana flows
_________

### DIFF
--- a/libstuff/SHTTPSManager.cpp
+++ b/libstuff/SHTTPSManager.cpp
@@ -93,7 +93,7 @@ void SStandaloneHTTPSManager::postPoll(fd_map& fdm, SStandaloneHTTPSManager::Tra
         // content. Why this is the what constitutes a valid response is lost to time. Any well-formed response should
         // be valid here, and this should get cleaned up. However, this requires testing anything that might rely on
         // the existing behavior, which is an exercise for later.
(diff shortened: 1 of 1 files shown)


libstuff/SHTTPSManager.cpp
```diff
@@ -93,7 +93,7 @@ void SStandaloneHTTPSManager::postPoll(fd_map& fdm, SStandaloneHTTPSManager::Tra
         // content. Why this is the what constitutes a valid response is lost to time. Any well-formed response should
         // be valid here, and this should get cleaned up. However, this requires testing anything that might rely on
         // the existing behavior, which is an exercise for later.
-        if (SContains(transaction.fullResponse.methodLine, " 200 ") || transaction.fullResponse.content.size()) {
+        if (SContains(transaction.fullResponse.methodLine, " 200") || transaction.fullResponse.content.size()) {
```
justinpersaud, line 96: Can you update the comment above?

This doesn't necessarily check for `200 OK` now and instead just that there is a "200" in the response

justinpersaud (changes requested):

Gonals:
COmments addressed!

justinpersaud (approved):

PR #1973, opened 2024-11-21
Improving logs for crashes

danieldoglas:
### Details

This PR changes the way we're logging information when we crash to include more info related to the command that was being executed.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/446307

### Tests

- Do a change in Auth in any command of your liking and add an `abort();` call
- You'll need to comment [this line](https://github.com/Expensify/Bedrock/blob/285ab2bf40feadc67ae82a44e9c261e02cfc4e41/libstuff/SLog.cpp#L98) locally or else it will throw
- Looking at the logs, you should see a log like this:
```
CRASHING from BedrockServer::runCommand, command: 'GetOnyxUpdates' ~~ Connection: '<REDACTED>' accountID: '282' authEmail: 'dsilva+0111@expensifail.com' authPartnerID: '<REDACTED>' authToken: '<REDACTED>' requestID: 'VuyvmL' updateIDFrom: '<REDACTED>' updateIDTo: '<REDACTED>'
```
- Then, look into your crash file (check the latest one using `ls -la /tmp/bedrock_cras*` and then use `sudo cat /tmp/bedrock_crash_{timestamp}.log`)
- The last line of the crash should be the same as the log line above.

_________

### DIFF
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -127,6 +127,7 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return {
         {"command", request.methodLine},
         {"requestID", request.isSet("requestID") ? request["requestID"] : "xxxxxx"},
+        {"logParam", request.isSet("logParam") ? request["logParam"] : ""},
     };
  }
 
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -123,7 +123,7 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return true;
 }
 
- map<string,string> BedrockCommand::getCommandLogParams() {
+map<string,string> BedrockCommand::getCommandLogParams() {
     map<string,string> parameters({{"command", request.methodLine}});
     for(auto& pair : request.nameValueMap) {
         parameters[pair.first] = pair.second;
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -123,14 +123,6 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return true;
 }
 
-map<string,string> BedrockCommand::getCommandLogParams() {
-    map<string,string> parameters({{"command", request.methodLine}});
-    for (auto& pair : request.nameValueMap) {
-        parameters[pair.first] = pair.second;
-    }
-    return parameters;
- }
-
 void BedrockCommand::reset(BedrockCommand::STAGE stage) {
     if (stage == STAGE::PEEK && !shouldPrePeek()) {
         jsonContent.clear();
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -124,11 +124,11 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
 }
 
  map<string,string> BedrockCommand::getCommandLogParams() {
-    return {
-        {"command", request.methodLine},
-        {"requestID", request.isSet("requestID") ? request["requestID"] : "xxxxxx"},
-        {"logParam", request.isSet("logParam") ? request["logParam"] : ""},
-    };
+    map<string,string> parameters({{"command", request.methodLine}});
+    for(auto& pair : request.nameValueMap) {
+        parameters[pair.first] = pair.second;
+    }
+    return parameters;
  }
 
 void BedrockCommand::reset(BedrockCommand::STAGE stage) {
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -123,7 +123,7 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return true;
 }
 
- map<string,string> BedrockCommand::getLogParams() {
+ map<string,string> BedrockCommand::getCommandLogParams() {
     return {
         {"command", request.methodLine},
         {"requestID", request.isSet("requestID") ? request["requestID"] : "xxxxxx"},
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -127,7 +127,7 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return {
         {"command", request.methodLine},
         {"requestID", request.isSet("requestID") ? request["requestID"] : "xxxxxx"},
-        };
+    };
  }
 
 void BedrockCommand::reset(BedrockCommand::STAGE stage) {
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -125,7 +125,7 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
 
 map<string,string> BedrockCommand::getCommandLogParams() {
     map<string,string> parameters({{"command", request.methodLine}});
-    for(auto& pair : request.nameValueMap) {
+    for (auto& pair : request.nameValueMap) {
         parameters[pair.first] = pair.second;
     }
     return parameters;
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -123,6 +123,10 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return true;
 }
 
+ map<string,string> BedrockCommand::getLogParams() {
+    return {{"command", request.methodLine}};
+ }
+
 void BedrockCommand::reset(BedrockCommand::STAGE stage) {
     if (stage == STAGE::PEEK && !shouldPrePeek()) {
         jsonContent.clear();
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -124,7 +124,10 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
 }
 
  map<string,string> BedrockCommand::getLogParams() {
-    return {{"command", request.methodLine}};
+    return {
+        {"command", request.methodLine},
+        {"requestID", request.isSet("requestID") ? request["requestID"] : "xxxxxx"},
+        };
  }
 
 void BedrockCommand::reset(BedrockCommand::STAGE stage) {
--- a/BedrockCommand.h
+++ b/BedrockCommand.h
@@ -119,7 +119,7 @@ class BedrockCommand : public SQLiteCommand {
     bool areHttpsRequestsComplete() const;
 
     // Returns a map with all the parameters that can be used to log information about the command being run.
-    virtual map<string,string> getLogParams();
+    virtual map<string,string> getCommandLogParams();
 
     // If the `peek` portion of this command needs to make an HTTPS request, this is where we store it.
     template <typename T>
--- a/BedrockCommand.h
+++ b/BedrockCommand.h
@@ -118,6 +118,9 @@ class BedrockCommand : public SQLiteCommand {
     // Returns true if all of the httpsRequests for this command are complete (or if it has none).
     bool areHttpsRequestsComplete() const;
 
+    // Returns a map with all the parameters that can be used to log information about the command being run.
+    virtual map<string,string> getLogParams();
+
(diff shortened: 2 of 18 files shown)


BedrockServer.cpp
```diff
@@ -2009,7 +2011,6 @@ bool BedrockServer::shouldBackup() {
 }
 
 SData BedrockServer::_generateCrashMessage(const unique_ptr<BedrockCommand>& command) {
-    SHMMM("Generating CRASH_COMMAND command for " << command->request.methodLine);
```
flodnv, line 2012: Can we please not remove this? I added this not long ago to help debug weird scenarios 😅 

I know you added logs in the callers of this method, but I feel like we might as well leave this one

BedrockCommand.cpp
```diff
@@ -123,6 +123,13 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return true;
 }
 
+ map<string,string> BedrockCommand::getCommandLogParams() {
```
flodnv, line 126: I am not seeing how this returns all the parameters of the command 😕

danieldoglas: It doesn't return all parameters. I can make it return all parameters though, but I chose to list only accountID and email directly from Auth.

danieldoglas:
Hmm I wonder if it started failing because of this
https://github.com/Expensify/Bedrock/blob/3dafa7ccdaa10d31d808c0f669e74afe25567538/libstuff/SLog.cpp#L98-L100

Since we're now logging everything from the request if it crashes, then it might be returning error 500 while running the tests.

danieldoglas:
Yes, running tests locally it returns:

```
2024-11-22T23:02:04.844504+00:00 expensidev2004 bedrock10001: VY5Bzc (libstuff.cpp:156) SException [socket6] [info] Throwing exception with message: '500 Log param Connection not in the whitelist, either do not log that or add it to PARAMS_WHITELIST if it's not sensitive' from libstuff/SLog.cpp:99
2024-11-22T23:02:04.967090+00:00 expensidev2004 bedrock10004: NaOfJd (libstuff.cpp:156) SException [socket6] [info] Throwing exception with message: '500 Log param Connection not in the whitelist, either do not log that or add it to PARAMS_WHITELIST if it's not sensitive' from libstuff/SLog.cpp:99
```

danieldoglas:
Whitelisted the parameters used in the tests.

BedrockServer.cpp
```diff
@@ -711,7 +720,10 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
     // If a signal is caught on this thread, which should only happen for unrecoverable, yet synchronous
     // signals, like SIGSEGV, this function will be called.
     SSetSignalHandlerDieFunc([&](){
+        const string logMessage = addLogParams("CRASHING from BedrockServer::runCommand", command->getCommandLogParams());
```
flodnv, line 723:
```suggestion
        SALERT("CRASHING from BedrockServer::runCommand", command->getCommandLogParams());
```

This does not work?

libstuff/SSignal.cpp
```diff
@@ -246,16 +246,22 @@ void _SSignal_StackTrace(int signum, siginfo_t *info, void *ucontext) {
...
             SWARN("Calling DIE function.");
-            SSignalHandlerDieFunc();
-            SSignalHandlerDieFunc = [](){};
+            string logMessage = SSignalHandlerDieFunc();
+            SSignalHandlerDieFunc = [](){ return ""; };
             SWARN("DIE function returned.");
+            
+            // Finish writing the crash file with the request details if it exists
+            if (fd != -1 && !logMessage.empty()) {
+                logMessage = format("{}{}", logMessage, "\n");
```
flodnv, line 260: IMO this usage of `format` is confusing. Why not just do `logMessage += "\n"`?

libstuff/libstuff.h
```diff
@@ -36,7 +36,7 @@ extern void* SSIGNAL_NOTIFY_INTERRUPT;
 // Initialize libstuff on every thread before calling any of its functions
 void SInitialize(string threadName = "", const char* processName = 0);
 
-void SSetSignalHandlerDieFunc(function<void()>&& func);
+void SSetSignalHandlerDieFunc(function<string()>&& func);
```
flodnv, line 39: Maybe add some doc to (briefly) explain what this function is, but especially why we need the `func` to return `string`?

BedrockCommand.cpp
```diff
@@ -123,6 +123,14 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return true;
 }
 
+map<string,string> BedrockCommand::getCommandLogParams() {
+    map<string,string> parameters({{"command", request.methodLine}});
+    for(auto& pair : request.nameValueMap) {
```
flodnv, line 128:
```suggestion
    for (auto& pair : request.nameValueMap) {
```

BedrockServer.cpp
danieldoglas: It does, but I need that same message to be returned as a string in the `SSetSignalHandlerDieFunc` method. I'm setting it as the `logMessage` so we don't have to process the function addLogParams twice.

libstuff/libstuff.h
danieldoglas: Done

libstuff/SSignal.cpp
danieldoglas: Changed.

flodnv (dismissed):

BedrockServer.cpp
tylerkaraszewski: Why don't we just do the `SALERT` in `_SSignal_StackTrace` and not need it here then?

BedrockCommand.cpp
```diff
@@ -123,6 +123,14 @@ bool BedrockCommand::areHttpsRequestsComplete() const {
     return true;
 }
 
+map<string,string> BedrockCommand::getCommandLogParams() {
```
tylerkaraszewski, line 126: I don't get why we have this and the whitelist. If we have a global whitelist, what is this for? On the other hand, if we have a per-command function to select params, what's the global whitelist for, shouldn't each command be making that decision?

BedrockCommand.h
```diff
@@ -118,6 +118,9 @@ class BedrockCommand : public SQLiteCommand {
     // Returns true if all of the httpsRequests for this command are complete (or if it has none).
     bool areHttpsRequestsComplete() const;
 
+    // Returns a map with all the parameters that can be used to log information about the command being run.
+    map<string,string> getCommandLogParams();
```
tylerkaraszewski, line 122: Needs to be virtual if you want to override in auth.

tylerkaraszewski (changes requested):

BedrockCommand.h
danieldoglas: We won't overwrite it in Auth since we're logging everything from the request, that's why it's not virtual.

BedrockServer.cpp
danieldoglas: Because in 2 calls, we have SWARN, and in the other 2, we have SALERT. If they were all the same time, we could do that directly in `_SSignal_StackTrace`



https://github.com/Expensify/Bedrock/blob/eb0c80b31917c8d9d049c4ffd0502505e7b4db9f/BedrockServer.cpp#L487
https://github.com/Expensify/Bedrock/blob/eb0c80b31917c8d9d049c4ffd0502505e7b4db9f/BedrockServer.cpp#L724
https://github.com/Expensify/Bedrock/blob/eb0c80b31917c8d9d049c4ffd0502505e7b4db9f/BedrockServer.cpp#L587
https://github.com/Expensify/Bedrock/blob/eb0c80b31917c8d9d049c4ffd0502505e7b4db9f/BedrockServer.cpp#L678

BedrockCommand.cpp
danieldoglas: I'm not sure I understand your question. This method adds everything that currently exists in the request into the map. 

The whitelist that exists in `addLogParams` redacts everything that that specific function receives.

For the calls we specifically do for `getCommandLogParams`, we'll pass this array to `addLogParams` so it is correctly redacted, but a bunch of other places will call (indirectly, by calling SALERT/SINFO/etc with a second parameter) `addLogParams` and will need to have other things redacted.

If your point is that we could redact things directly here, I guess we can, but I don't see any gains on doing that.

BedrockServer.cpp
danieldoglas: Since I do think we want them to be different, I left them to use the appropriate log level where they were instead of inside `_SSignal_StackTrace`

flodnv: I think SALERT everywhere is fine

danieldoglas: Doesn't SALERT create issues on GH?

flodnv: SALERT and SWARN create issues if they happened more than 7 times in 30 minutes, unless they have the string `ENSURE_BUGBOT`, which ensures they get created regardless of occurrence count.

danieldoglas: @flodnv @tylerkaraszewski what do you think?

flodnv: I'm fine either way, don't block on me.

tylerkaraszewski: > We'll run the log line saying we're crashing after we've already sent the crash command. That doesn't make a lot of sense to me.

I don't think either order is obviously "better". I also don't think we need to long the same thing twice with ALERT and WARN levels. If we're crashing we can just ALERT once and that seems fine?

danieldoglas: OK, I'll change the PR to only use alert and we'll do the alert from `SSetSignalHandlerDieFunc`

BedrockCommand.cpp
tylerkaraszewski: If we are going to use a global whitelist to redact log parameters, I don't see why we need `getCommandLogParams` because the `commandLogParams` are just `command.nameValueMap`. Yes, this adds the command name, but we already have that everywhere we'd want this.

We can just change the two places we call this to do something like:
```
const string logMessage = addLogParams("CRASHING from BedrockServer::sync in command " + command->methodLine, command->nameValueMap);
```

Then we don't need `getCommandLogParams` to make an unnecessary copy of a map we already have.

BedrockCommand.h
tylerkaraszewski: Then I don't see why we need it. I thought we were adding it so that auth can return a different list of params.

Also:

Screenshot 2024-11-26 at 13 03 13

tylerkaraszewski (changes requested):

BedrockCommand.cpp
danieldoglas: `addLogParams` only receives a map as a second parameter and will only apply the redacting logic if it's part of the map, not if it's part of the text.

The second parameter for `addLogParams` doesn't accept the map we have on `command->request.nameValueMap` since it's a `map<string, SString, STableComp>`

I can either create a new function `addLogParams` that would receive that as a second parameter and iterate through it, or leave the `getCommandLogParams` that is creating a new map.

What do you think its better?

BedrockCommand.h
danieldoglas: It was at first, but after adding everything in the request we don't need auth to do that anymore. Let me fix the PR description.

BedrockCommand.cpp
danieldoglas: the problem with adding a second function with `addLogParams` that can receive an STable is that it will do exactly the same thing we do in `addLogParams`. But maybe that's not a real problem.

tylerkaraszewski (approved):

PR #1988, opened 2024-11-29
Remove log params that are not related to Bedrock

danieldoglas:
### Details

This function removes all whitelisted log params names that are not related to Bedrock, and adds a new function that will allow auth to include those param names.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/448720

### Tests
N/A
_________


libstuff/SLog.cpp
```diff
@@ -114,3 +73,7 @@ string addLogParams(string&& message, const STable& params) {
 
     return message;
 }
+
+void SWhitelistLogParams(set<string> params) {
```
iwiznia, line 77: NAB: add a comment on what this does/when to use

iwiznia (approved):

libstuff/SLog.cpp
danieldoglas: Done

iwiznia (approved):

PR #2008, opened 2024-12-09
Adding limit on how many times a crash command can run

danieldoglas:
### Details

We currently store specific identifying values when a crash happens. That means that if a specific user call a command in Bedrock (e.g. userID 1) and that command crashes the node, we will store that userID and won't let the cluster execute that command for that user again.

The problem is that if the issue is in the command, and not on specific data related to that user, UserID 2 or UserID 3 could still call the same command and progressively crash all nodes in the cluster.

This change will block commands from being executed if the same command already crashed the cluster more than one time, independent of the specific crash identifying values.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/451831

### Tests
need to work on tests.
_________


iwiznia (approved):

flodnv (approved):

danieldoglas:
@tylerkaraszewski all yours

tylerkaraszewski (approved):

pecanoro:

PR #2059, opened 2025-01-13
Update GH actions cache to 4.2.0 as 4.0 is deprecated

johnmlee101:
https://github.com/actions/cache/discussions/1510

Its deprecated so we need to bump this


pecanoro:

PR #2089, opened 2025-02-05
Throw instead of crashing when writing outside of transaction

puneetlath:
### Details

Right now, if you attempt to write in peek or prePeek then we crash. This PR changes that so that we throw an error instead of crashing. 

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/466913

### Tests
I updated the addComment command locally to attempt to do a write in prePeek. Before the change that generates a crash: 
Screenshot 2025-02-05 at 11 56 50 AM

After the change, no crash, the command just errors out. 
Screenshot 2025-02-05 at 11 55 05 AM

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -586,7 +586,7 @@ bool SQLite::writeUnmodified(const string& query) {
 
 bool SQLite::_writeIdempotent(const string& query, SQResult& result, bool alwaysKeepQueries) {
     if (!_insideTransaction) {
-        STHROW("Attempted to write outside of transaction");
+        STHROW("500 Attempted to write outside of transaction");
     }
 
     _queryCache.clear();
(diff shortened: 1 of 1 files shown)


puneetlath:
@tylerkaraszewski @flodnv do you know of a way to write an automated test for this? Or do you think what I have here is good enough?

sqlitecluster/SQLite.cpp
```diff
@@ -585,7 +585,10 @@ bool SQLite::writeUnmodified(const string& query) {
 }
 
 bool SQLite::_writeIdempotent(const string& query, SQResult& result, bool alwaysKeepQueries) {
-    SASSERT(_insideTransaction);
+    if (!_insideTransaction) {
+        STHROW("Attempted to write outside of transaction");
```
flodnv, line 589:
```suggestion
        STHROW("500 Attempted to write outside of transaction");
```

flodnv (changes requested):
Nice! I think the manual test is enough

flodnv (approved):

PR #2138, opened 2025-04-01
Seperate out request body

thienlnam:
### Details
Seperates out the request body from the msg so bugbots can be created even with different params

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/472994

### Tests
Compiled auth
_________

### DIFF
--- a/BedrockCore.cpp
+++ b/BedrockCore.cpp
@@ -372,7 +372,7 @@ void BedrockCore::_handleCommandException(unique_ptr<BedrockCommand>& command, c
 
     STable logParams;
     if (!e.body.empty()) {
-        logParams["request_body"] = e.body;
+        logParams = SParseJSONObject(e.body);
     }
 
     if (SContains(e.what(), "_ALERT_")) {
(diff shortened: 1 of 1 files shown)


BedrockCore.cpp
```diff
@@ -369,19 +369,22 @@ void BedrockCore::postProcessCommand(unique_ptr<BedrockCommand>& command, bool i
 
 void BedrockCore::_handleCommandException(unique_ptr<BedrockCommand>& command, const SException& e, SQLite* db, const BedrockServer* server) {
     string msg = "Error processing command '" + command->request.methodLine + "' (" + e.what() + "), ignoring.";
+
+    STable logParams;
     if (!e.body.empty()) {
-        msg = msg + " Request body: " + e.body;
+        logParams = SParseJSONObject(e.body);
```
flodnv, line 375: Are we sure `e.body` is always JSON, @tylerkaraszewski ?

flodnv (commented):
Nice and easy! 😄

BedrockCore.cpp
thienlnam: I was looking at SParseJSONObject and it seems like it will just return empty if it's invalid JSON so that should be fine but I guess this will clear out any string bodies

flodnv (approved):

pecanoro:
This broke everything, so reverting. Context here: https://expensify.slack.com/archives/C07J32337/p1743618812785699

PR #2148, opened 2025-04-09
Update MySQL Plugin to work

coleaeason:
### Details
@tylerkaraszewski Please review. This updates the MySQL plugin to work again. We broke it somewhere along the way with the removal of `onPortAccept()` and additional breaking changes in the MySQL protocol used in modern clients.

This has been tested using the CLI client, SequelAce, and Alteryx and it works with all three. It's still not a feature complete implementation but it supports probably 95% of what you'd want to do.

This also stops extending the DB plugin and instead makes this a standalone plugin, this allows us to use this in conjunction with our internal DB plugin version, or any plugin that implements a `Query` command.

### Fixed Issues
Broken plugin

### Tests

_________


tylerkaraszewski (approved):
Seems fine to me.

PR #2229, opened 2025-07-09
Protect pointer delete

aldo-expensify:
### Details

My PR here seems to have made it much more likely to make this race condition fail: https://github.com/Expensify/Auth/pull/16131#issuecomment-3050749255

~I think this is the source of our random segmentation faults we get in github when running Auth tests.~ (don't think this is the case anymore.. this seems to be a new segmentation fault that we can have thanks to the changes I'm proposing in Auth)

The fix is trying to make sure there can't be two threads calling `delete _db` before we make `_db = nullptr`

I'm not sure if there are other areas where we assign `_db` that should be protected, but at least this seems to be enough to fix the problem.

### Fixed Issues
Fixes https://github.com/Expensify/Auth/pull/16131#issuecomment-3050749255

### Tests

To make it easier to fail, you can check out https://github.com/Expensify/Auth/pull/16131

1. Run tests with `./authtest -only BatchTopUpCredits` many times and confirm that the test passes without crashing.

Without this bedrock PR, I get a lot this crash:

```
expensidev-2404->Auth/test (aldo_wait-for-tables-in-cluster-to-be-ready) > ./authtest -only BatchTopUpCredits
Aborted (core dumped)
```

specially if you checkout the mentioned Auth PR.

_________

### DIFF
--- a/test/lib/BedrockTester.cpp
+++ b/test/lib/BedrockTester.cpp
@@ -534,6 +534,7 @@ vector<SData> BedrockTester::executeWaitMultipleData(vector<SData> requests, int
 
 SQLite& BedrockTester::getSQLiteDB()
 {
+    lock_guard<decltype(_dbMutex)> lock(_dbMutex);
     if (!_db) {
         // Assumes wal2 mode.
         _db = new SQLite(_args["-db"], 1000000, 3000000, -1, 0, ENABLE_HCTREE);
@@ -542,10 +543,7 @@ SQLite& BedrockTester::getSQLiteDB()
 }
 
 void BedrockTester::freeDB() {
-    if (_db == nullptr) {
-        return;
-    }
-    lock_guard<decltype(_testersMutex)> lock(_testersMutex);
+    lock_guard<decltype(_dbMutex)> lock(_dbMutex);
     delete _db;
     _db = nullptr;
 }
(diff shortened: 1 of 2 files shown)


aldo-expensify:
I'm starting to wonder if this can really happen without my changes here: https://github.com/Expensify/Auth/pull/16131

My changes call `tester->readDB` which initializes `_db`... but maybe in `main` nothing is initializing `_db` before we start running those threads

tylerkaraszewski (changes requested):
This is a great catch. I duplicated it with your test case and I see why it is hard to reproduce otherwise. The auth change here:
https://github.com/Expensify/Auth/pull/16131/files

Triggers creating the _db object earlier, such that this is much more likely to happen.

Without that change, which doesn't even directly look related, I couldn't reproduce.

Suggestions:

1. Use a local mutex for this, a member of the `BedrockTester` object, rather than a global mutex across all bedrock testers.
2. Use the same protection in `getSQLiteDB` where we assign the DB. This will prevent potential leaks of DB objects.
3. I don't think we need to bother with the early return, just lock our local mutex at the top of `freeDB` and `getSQLiteDB`

aldo-expensify:
Addressed comments!

PR #2247, opened 2025-08-04
Log an alert when we fail to create a job for any reason

flodnv:
### Fixed Issues
In https://github.com/Expensify/Web-Expensify/pull/47929, we saw that we had been failing to create jobs for a long while. This PR intends to make such cases more obvious in the future by logging an alert.

### Tests
Run this command:
```
CreateJob
asdf
```

Check logs:
```
2025-08-04T12:03:42.897613+00:00 expensidev-2404 bedrock: P9m2ok (Jobs.cpp:324) peek [socket0] [alrt] {Jobs} Failed creating job ~~ what: '402 Missing name'
```

### DIFF
--- a/libstuff/SLog.cpp
+++ b/libstuff/SLog.cpp
@@ -42,25 +42,21 @@ void SLogStackTrace(int level) {
 
 // If the param name is not in this whitelist, we will log <REDACTED> in addLogParams.
 static set<string> PARAMS_WHITELIST = {
-    "approver",
-    "approvers",
     "command",
     "Connection",
     "Content-Length",
     "count",
-    "employeeEmail",
-    "employees",
     "indexName",
     "isUnique",
     "logParam",
     "message",
     "peer",
-    "policyID",
     "reason",
     "requestID",
     "status",
     "topic",
     "userID",
+    "what",
 };
 
 string addLogParams(string&& message, const STable& params) {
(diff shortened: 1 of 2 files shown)


libstuff/SLog.cpp
```diff
@@ -42,25 +42,21 @@ void SLogStackTrace(int level) {
 
 // If the param name is not in this whitelist, we will log <REDACTED> in addLogParams.
 static set<string> PARAMS_WHITELIST = {
-    "approver",
```
flodnv, line 45: Counterpart PR: https://github.com/Expensify/Auth/pull/16525

plugins/Jobs.cpp
```diff
@@ -206,118 +206,123 @@ bool BedrockJobsCommand::peek(SQLite& db) {
...
+                            // Return early, no need to pass to leader, there are no more jobs to create.
+                            SINFO("Job already existed and unique flag was passed, reusing existing job " << result[0][0] << ", mocked? " << (mockRequest ? "true" : "false"));
+                            jsonContent["jobID"] = result[0][0];
+                            return true;
+                        }
                     }
                 }
             }
+        } catch (const exception& e) {
+            SALERT("Failed creating job", {{"what", e.what()}});
```
iwiznia, line 324: For non fire and forget jobs this means we will create 2 issues, one for the log in PHP and one for the log in auth?

iwiznia (approved):

plugins/Jobs.cpp
flodnv: Yes

iwiznia: Do we want that or maybe we rather only log this on fire&forget commands?

flodnv: I think this is fine.

danieldoglas (approved):

danieldoglas:
https://github.com/Expensify/Auth/pull/16525 will be deployed in a few minutes.

PR #2250, opened 2025-08-08
Create SQLite user-defined function: deburr

roryabraham:
### Details
Creating a user-defined SQLite function called `DEBURR` to match [lodash deburr](https://lodash.com/docs/4.17.15#deburr)

### Fixed Issues
related to https://github.com/Expensify/Expensify/issues/492122

### Tests
Added automated tests in both Bedrock and Auth. Added benchmarks, which I used to measure a few optimizations which made it much faster.
_________


benchmarks/BenchmarkBase.h
```diff
@@ -0,0 +1,130 @@
...
+    template<typename InputType, typename Func>
+    uint64_t runBench(const std::string& name, 
+                      const std::vector<InputType>& inputs, 
+                      int iterations, 
+                      Func func) {
+        setup();
+
+        // Warm-up: run a few times to get CPU caches ready
+        volatile size_t guard = 0;
+        for (int i = 0; i < 100; ++i) {
```
luacmartins: Why 100?

benchmarks/main.cpp
```diff
@@ -0,0 +1,284 @@
...
+                 << right << setw(15) << "N/A"
+                 << setw(15) << fixed << setprecision(2) << currentResult.throughputMBps
+                 << setw(12) << "NEW"
+                 << setw(15) << "✨ NEW" << "\n";
+        } else {
+            const auto& baselineResult = it->second;
+            double changePercent = ((currentResult.throughputMBps - baselineResult.throughputMBps) 
+                                   / baselineResult.throughputMBps) * 100.0;
+            
+            // Use colors for significant changes (>1% difference)
```
luacmartins: 1% seems kinda low, no? I'd imagine something like 5-10% being a better indicator of performance change

luacmartins (commented):
LGTM but I think we should get an additional reviewer who's more familiar with Bedrock. Additionally, are there manual test steps?

benchmarks/BenchmarkBase.h
roryabraham: Chosen randomly. Also idk how important the CPU warm-up really is. I think it's commonplace in benchmarking libraries, but also this code was written primarily by Claude.

benchmarks/main.cpp
roryabraham: Yeah that's good feedback, I'll bump it up

inimaga (approved):
Code changes look good to me. But going to loop in @tylerkaraszewski as I can't be the Bedrock expert you're after 😅

luacmartins (approved):

benchmarks/BenchmarkBase.h
```diff
@@ -0,0 +1,130 @@
...
+ *   void myBenchmark() {
+ *       std::vector<std::string> inputs = {"test1", "test2"};
+ *       runBench("MyTest", inputs, 10000, [](const std::string& s) {
+ *           return s.size(); // Your actual work here
+ *       });
+ *   }
+ */
+class BenchmarkBase {
+public:
+    BenchmarkBase(const std::string& name) : testName(name) {}
```
tylerkaraszewski: we omit `std::` and use `using namespace std` as a stylistic choice.

benchmarks/BenchmarkBase.h
```diff
@@ -0,0 +1,130 @@
...
+                      int iterations, 
+                      Func func) {
+        setup();
+
+        // Warm-up: run a few times to get CPU caches ready
+        volatile size_t guard = 0;
+        for (int i = 0; i < 100; ++i) {
+            for (const auto& input : inputs) {
+                auto result = func(input);
+                guard += sizeof(result); // Use the result somehow
```
tylerkaraszewski, line 78: What is this supposed to do? Why do we even bother to keep the result of `func(input)` here?

benchmarks/BenchmarkBase.h
```diff
@@ -0,0 +1,130 @@
...
+        
+        for (int it = 0; it < iterations; ++it) {
+            for (const auto& input : inputs) {
+                auto result = func(input);
+                guard += sizeof(result); // Prevent optimization
+                totalBytes += getInputSize(input);
+            }
+        }
+        
+        const uint64_t elapsedUs = STimeNow() - start;
```
tylerkaraszewski: I prefer std::chrono, but you don't need to change it if you like this.

benchmarks/BenchmarkBase.h
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+private:
+    /**
+     * Get the size in bytes of different input types.
+     */
+    template<typename T>
+    size_t getInputSize(const T& input) {
+        if constexpr (std::is_same_v<T, std::string>) {
+            return input.size();
```
tylerkaraszewski, line 133: So if this is just supposed to work on strings for now, I would just *always* use the above line and never use `sizeof`. `sizeof` will be wrong for all standard container types. If we need this for integral types, I would add a specialization with `is_integral` rather than having everything that's not a string use `sizeof`.

benchmarks/main.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+int main(int argc, char* argv[]) {
+    set<string> include;
+    set<string> exclude;
+    string baselineRef;
+    
+    // Parse arguments
+    for (int i = 1; i < argc; ++i) {
+        string arg = argv[i];
+        if (arg == "--baseline" && i + 1 < argc) {
```
tylerkaraszewski: Can we either use `SParseCommandLine` like bedrock does or `getopt` like everything else does?

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,316 @@
...
+ * 1. Go through each character in the text
+ * 2. If it's a regular letter (a-z, A-Z), keep it as-is
+ * 3. If it's an accented character (é, ñ, etc.), replace it with the basic version
+ * 4. If it's an accent mark by itself, delete it
+ * 5. Everything else (numbers, punctuation, emoji) stays the same
+ *
+ * Examples: "café" → "cafe", "naïve" → "naive", "Zürich" → "Zurich"
+ */
+std::string SDeburr::deburr(const std::string& input) {
+    const unsigned char* in = reinterpret_cast<const unsigned char*>(input.c_str());
```
tylerkaraszewski: I don't think we need reinterpret_cast here.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,316 @@
...
+    const size_t len = input.size();
+    string result;
+    result.reserve(len);
+    size_t i = 0;
+    while (i < len) {
+        // Speed optimization: copy regular ASCII text in chunks
+        if (in[i] < 0x80) {
+            size_t asciiStart = i++;
```
tylerkaraszewski: the `++` on this line serves to make this more confusing.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,316 @@
...
+            while (i < len && in[i] < 0x80) {
+                ++i;
+            }
+            result.append(reinterpret_cast<const char*>(in + asciiStart), i - asciiStart);
+            continue;
+        }
+
+        // Handle special characters (accented letters, etc.)
+        size_t start = i;
+        uint32_t cp = decodeUTF8Codepoint(in, len, i);
```
tylerkaraszewski: decodeUTF8Codepoint should either take a string and and index, or it should just take a pointer to the character in question and a pointer to the end of the string.

tylerkaraszewski (changes requested):
So as I was reviewing this, I thought the whole thing was sort of overly complex and hard to read. 

I implemented a very basic replacement function, that I think is much clearer to read and maintain. It lacks some of the optimizations for speed we have in the current code, but I don't think that will actually matter. I also think that we probably don't actually need a benchmarking framework for this.

This is what I wrote:

```
// This is only a sample, and needs the rest of:
// Latin-1 Supplement: https://en.wikipedia.org/wiki/Latin-1_Supplement#Character_table
// and Latin Extended-A: https://en.wikipedia.org/wiki/Latin_Extended-A
static const map<string, string> UNICODE_REPLACEMENTS{
    {"Ć", "C"},
    {"ć", "c"},
};

static string unicodeDeburr(const string& input) {
    string output;
    const char* current = input.data();
    while (current < input.data() + input.size()) {
        // Determine how many bytes there are in this character. See "Description" here: https://en.wikipedia.org/wiki/UTF-8
        size_t byteLength = 1;
        if ((current[0] & 0b11000000) == 0b11000000) {
            byteLength++;
            if ((current[0] & 0b11100000) == 0b11100000) {
                byteLength++;
                if ((current[0] & 0b11110000) == 0b11110000) {
                    byteLength++;
                }
            }
        }

        // basic ASCII characters can be copied with no additional checking.
        if (byteLength == 1) {
            output += current[0];
            current += byteLength;
            continue;

        }

        // Combining Diacritical Marks allow for adding accent marks to other characters essentailly by following a "base" character
        // with an accent mark. Lodash removes these (https://lodash.com/docs/4.17.15#deburr) so we will as well.
        // See the description here: https://en.wikipedia.org/wiki/Combining_Diacritical_Marks
        if (byteLength == 2) {
            bool isInCombiningDiacriticalMarks =
                (current[0] == 0xCC && current[1] >= 0x80 && current[1] <= 0xBF) ||
                (current[0] == 0xCD && current[1] >= 0x80 && current[1] <= 0xAF);
            if (isInCombiningDiacriticalMarks) {
                // If this *is* in the Combining Diacritical Marks block, we can just drop it and be left with the base character.
                current += byteLength;
                continue;
            }
        }

        // We are now left with any multi-byte unicode character that's not a Combining Diacritical Mark.
        // Create a string to represent it.
        string lookup(current, byteLength);

        // Look up and see if we need to replace this string. If so, do that, otherwise keep the unicode string we already made.
        auto replacement = UNICODE_REPLACEMENTS.find(lookup);
        if (replacement == UNICODE_REPLACEMENTS.end()) {
            output += lookup;
        } else {
            output += replacement->second;
        }

        // We have consumed these bytes and can move on.
        current += byteLength;
    }

    return output;
}

    // UTF-8: 0x65 (e), 0xCC 0x81 (combining acute accent)
    const string combiningDiacriticalAccentedLowerCaseE = "\x65\xCC\x81";

    string input = "A String with Ć and ć in it. And ś, which is not currently replaced. Also, " + combiningDiacriticalAccentedLowerCaseE
                    + ", which should be automatically.";
    string output = unicodeDeburr(input);
    cout << input << endl;
    cout << output << endl;

    return 0;
}
```

When run, this outputs:
```
expensidev-2404->Bedrock/test (main) > ./test
A String with Ć and ć in it. And ś, which is not currently replaced. Also, é, which should be automatically.
A String with C and c in it. And ś, which is not currently replaced. Also, e, which should be automatically.
```

If we fill out `UNICODE_REPLACEMENTS` more completely, I feel this would be much easier to read and maintain.

We could also add a help function that could build `UNICODE_REPLACEMENTS` from a set of pairs like:
```
{"c": "ćĉċč"}
```

Which might make it easier to read still.

benchmarks/BenchmarkBase.h
roryabraham: Well Claude's claim was that by using it somehow you ensure that the compiler doesn't just remove the unused code.

roryabraham:
Still looking at how to make this more readable without compromising performance too much. Here's a comparison of benchmarks between the current implementation and Tyler's:

|Benchmark|Current|Tyler's simplified|
|---|---|---|
|ShortASCII|8ms|44ms|
|Latin1|12ms|39ms|
|MixedLong|5ms|19ms|
|ShortASCIIHigh|220ms|506ms|
|Latin1High|650ms|1623ms|
|MixedLongHigh|578ms|1929ms|

luacmartins:
That's quite significant, but also concerning that it can take that long to process those strings. @roryabraham what does the test data look like? Are you running the benchmark in this PR?

tylerkaraszewski:
These are the times I got using -O2 and Rory's code in his benchmark to use as a baseline:
```
✅ SDeburrBench::benchShortASCII (20ms)
✅ SDeburrBench::benchLatin1 (18ms)
✅ SDeburrBench::benchMixedLong (7ms)
✅ SDeburrBench::benchShortASCIIHigh (166ms)
✅ SDeburrBench::benchLatin1High (507ms)
✅ SDeburrBench::benchMixedLongHigh (447ms)
```

These are what I got using my code posted above (with only two characters to replace):
```
✅ SDeburrBench::benchShortASCII (13ms)
✅ SDeburrBench::benchLatin1 (18ms)
✅ SDeburrBench::benchMixedLong (6ms)
✅ SDeburrBench::benchShortASCIIHigh (159ms)
✅ SDeburrBench::benchLatin1High (486ms)
✅ SDeburrBench::benchMixedLongHigh (431ms)
```

These are basically the same, indicating that if mine is slower, it's mostly due to map lookup. Let me flesh out the map.

tylerkaraszewski:
After adding latin-1 supplement but not latin extended-A:

```
✅ SDeburrBench::benchShortASCII (15ms)
✅ SDeburrBench::benchLatin1 (18ms)
✅ SDeburrBench::benchMixedLong (6ms)
✅ SDeburrBench::benchShortASCIIHigh (159ms)
✅ SDeburrBench::benchLatin1High (481ms)
✅ SDeburrBench::benchMixedLongHigh (422ms)
```

Still basically the same.

tylerkaraszewski:
After adding extended-A
```
✅ SDeburrBench::benchShortASCII (5ms)
✅ SDeburrBench::benchLatin1 (8ms)
✅ SDeburrBench::benchMixedLong (4ms)
✅ SDeburrBench::benchShortASCIIHigh (61ms)
✅ SDeburrBench::benchLatin1High (152ms)
✅ SDeburrBench::benchMixedLongHigh (224ms)
```

Not sure why it's faster.

tylerkaraszewski:
So I made an alternate bracnh here:
https://github.com/Expensify/Bedrock/pull/2251

It's just the code I posted above in place of @roryabraham's, which is commented out, and with the rest of `Latin-1 Supplement` and `Latin Extended-A` added.

I compiled with `-O2` and ran @roryabraham's benchmark:
```
./benchmarks/bench --baseline 418f2ba037aa8839b4c07b70147d4ea3e06552c9
```

That commit is the latest in this branch to use as the baseline. This is the result:
```
=== Benchmark Comparison Report ===

Benchmark                       Baseline MB/s   Current MB/s    Change %         Status
-------------------------------------------------------------------------------------
SDeburrBench::Latin1                   230.16         243.73        +5.9%   ✅ IMPROVED
SDeburrBench::Latin1High               227.97         230.21        +1.0%  ≈ UNCHANGED
SDeburrBench::MixedLong                212.39         219.12        +3.2%  ≈ UNCHANGED
SDeburrBench::MixedLongHigh            228.17         234.11        +2.6%  ≈ UNCHANGED
SDeburrBench::ShortASCII               581.64         565.38        -2.8%  ≈ UNCHANGED
SDeburrBench::ShortASCIIHigh           550.84         558.28        +1.4%  ≈ UNCHANGED

Summary: 1 improved, 0 regressed, 5 unchanged
```

This shows these to be roughly equal.

roryabraham:
With moderate optimization enabled `BEDROCK_OPTIM_COMPILE_FLAG='-O2'`, I get these benchmarks for the current implementation:

```
vagrant@expensidev-2404:/vagrant/Bedrock$ ./benchmarks/bench
--------------
[SDeburrBench] ShortASCII: inputs=6, iters=20000, bytes=1060000, time_us=3435, throughput_MBps=294.293
✅ SDeburrBench::benchShortASCII (3ms)
[SDeburrBench] Latin1: inputs=9, iters=10000, bytes=780000, time_us=3759, throughput_MBps=197.889
✅ SDeburrBench::benchLatin1 (3ms)
[SDeburrBench] MixedLong: inputs=1, iters=5000, bytes=565000, time_us=2683, throughput_MBps=200.83
✅ SDeburrBench::benchMixedLong (2ms)
[SDeburrBench] ShortASCIIHigh: inputs=6, iters=500000, bytes=26500000, time_us=50812, throughput_MBps=497.37
✅ SDeburrBench::benchShortASCIIHigh (50ms)
[SDeburrBench] Latin1High: inputs=9, iters=500000, bytes=39000000, time_us=87117, throughput_MBps=426.935
✅ SDeburrBench::benchLatin1High (87ms)
[SDeburrBench] MixedLongHigh: inputs=1, iters=500000, bytes=56500000, time_us=93996, throughput_MBps=573.244
✅ SDeburrBench::benchMixedLongHigh (94ms)

Slowest Test Classes: 
243ms: SDeburrBench : 100% of total test time
Total test time across threads: 243ms
```

clearly, it makes a big difference

Then just switching to an `unordered_map` instead of the flat arrays, we get these results:

```
vagrant@expensidev-2404:/vagrant/Bedrock$ ./benchmarks/bench
--------------
[SDeburrBench] ShortASCII: inputs=6, iters=20000, bytes=1060000, time_us=2851, throughput_MBps=354.576
✅ SDeburrBench::benchShortASCII (2ms)
[SDeburrBench] Latin1: inputs=9, iters=10000, bytes=780000, time_us=8321, throughput_MBps=89.3962
✅ SDeburrBench::benchLatin1 (8ms)
[SDeburrBench] MixedLong: inputs=1, iters=5000, bytes=565000, time_us=3206, throughput_MBps=168.068
✅ SDeburrBench::benchMixedLong (3ms)
[SDeburrBench] ShortASCIIHigh: inputs=6, iters=500000, bytes=26500000, time_us=48003, throughput_MBps=526.475
✅ SDeburrBench::benchShortASCIIHigh (48ms)
[SDeburrBench] Latin1High: inputs=9, iters=500000, bytes=39000000, time_us=130028, throughput_MBps=286.041
✅ SDeburrBench::benchLatin1High (130ms)
[SDeburrBench] MixedLongHigh: inputs=1, iters=500000, bytes=56500000, time_us=151968, throughput_MBps=354.565
✅ SDeburrBench::benchMixedLongHigh (152ms)

Slowest Test Classes: 
346ms: SDeburrBench : 100% of total test time
Total test time across threads: 346ms
```

the numbers aren't as alarming, but it's still a ~30% increase

roryabraham:
I'm much happier with the state of the code now. I think it's much more readable than it was. It's also nice and fast (compiled with `-O2`)

```
vagrant@expensidev-2404:/vagrant/Bedrock$ ./benchmarks/bench
--------------
[SDeburrBench] ShortASCII: inputs=6, iters=20000, bytes=1060000, time_us=5082, throughput_MBps=198.917
✅ SDeburrBench::benchShortASCII (5ms)
[SDeburrBench] Latin1: inputs=9, iters=10000, bytes=780000, time_us=3928, throughput_MBps=189.375
✅ SDeburrBench::benchLatin1 (4ms)
[SDeburrBench] MixedLong: inputs=1, iters=5000, bytes=565000, time_us=2294, throughput_MBps=234.885
✅ SDeburrBench::benchMixedLong (2ms)
[SDeburrBench] ShortASCIIHigh: inputs=6, iters=500000, bytes=26500000, time_us=62314, throughput_MBps=405.565
✅ SDeburrBench::benchShortASCIIHigh (62ms)
[SDeburrBench] Latin1High: inputs=9, iters=500000, bytes=39000000, time_us=65513, throughput_MBps=567.724
✅ SDeburrBench::benchLatin1High (65ms)
[SDeburrBench] MixedLongHigh: inputs=1, iters=500000, bytes=56500000, time_us=78149, throughput_MBps=689.485
✅ SDeburrBench::benchMixedLongHigh (78ms)

Slowest Test Classes: 
218ms: SDeburrBench : 100% of total test time
Total test time across threads: 218ms
```

I still need to do a bit more cleanup of the benchmark program though.

benchmarks/BenchmarkBase.h
roryabraham: updated

benchmarks/BenchmarkBase.h
roryabraham: updated

benchmarks/BenchmarkBase.h
roryabraham: Good catch, updated.

benchmarks/main.cpp
roryabraham: updated

libstuff/SDeburr.cpp
roryabraham: Updated! I agree in general that if it matters whether it's `++i` or `i++`, then it's confusing and should just be on its own line

libstuff/SDeburr.cpp
roryabraham: Resolving as this function has been removed and is inlined in `SDeburr::deburr`

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <cstring>
+#include <string>
+
+#include <libstuff/libstuff.h>
+#include <libstuff/sqlite3.h>
+
+using namespace std;
+
+const char* SDeburr::unicodeToAscii(uint32_t codepoint) {
+    if (codepoint >= 0x00C0 && codepoint <= 0x017F) {
```
tylerkaraszewski, line 16: Can you comment what these indicate? Possibly with a link to the unicode spec or something.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+        return SDeburr::UNICODE_TO_ASCII_MAP[codepoint - 0x00C0];
+    }
+
+    if (codepoint == 0x1E9E) {
```
tylerkaraszewski: This is also probably worth a comment.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+        return "SS"; // ẞ
+    }
+
+    return nullptr;
+}
+
+string SDeburr::deburr(const string& input) {
+    const unsigned char* input_bytes = reinterpret_cast<const unsigned char*>(input.c_str());
```
tylerkaraszewski: Shouldn't need the cast here.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+    const size_t len = input.size();
```
tylerkaraszewski: Seems unnecessary as a separate variable.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+        // Count the leading ones in the current byte
+        // 0 leading ones -> ASCII character
+        // 1 leading one -> continuation byte
+        // 2-4 leading ones -> 2, 3, or 4-byte sequence
+        // 5+ leading ones -> invalid UTF-8
+        const int numLeadingOnes = countl_one(input_bytes[i]);
+
+        // Most common case: regular ASCII text
+        if (numLeadingOnes == 0) {
+            result.append(reinterpret_cast<const char*>(input_bytes + i), 1);
```
tylerkaraszewski: I think this is clearer as just result.append(1, input_bytes[i]);

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+            i++;
+            continue;
+        }
+
+        // Check for invalid UTF-8. If invalid, skip the current byte and continue.
+        // Cases we check for:
+        // - A continuation byte (1 leading one). This should be handled only as part of a multi-byte sequence.
+        // - More than 4 leading ones
+        // - Not enough bytes to form a valid sequence
+        if (numLeadingOnes == 1 || numLeadingOnes > 4 || (i + numLeadingOnes) > len) {
```
tylerkaraszewski: Should this warn?

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+            i++;
+        }
+
+        // Accent marks by themselves (like ´ ` ^) → skip them
+        if (codepoint >= 0x0300 && codepoint <= 0x036F) {
+            continue;
+        }
+
+        // Map the unicode codepoint to ascii
+        const char* mapped = unicodeToAscii(codepoint);
```
tylerkaraszewski, line 98: I think it would make more sense for this to return `char` instead of a `char *`. It can still return 0 as the not found value.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,137 @@
...
+        if (mapped == nullptr) {
+            // No conversion needed, keep the original character
+            result.append(reinterpret_cast<const char*>(input_bytes + start), i - start);
```
tylerkaraszewski: Shouldn't need the cast.

libstuff/SDeburr.h
```diff
@@ -0,0 +1,102 @@
...
+     * Fast lookup table for converting accented characters.
+     *
+     * It's a flat array that works like a dictionary where the character's number is used as an array index,
+     * so we can look up it's ASCII equivalent very quickly.
+     */
+    inline static constexpr auto UNICODE_TO_ASCII_MAP = []() constexpr {
+        array<const char*, 0x0180 - 0x00C0> map = {};
+
+        // Latin-1 Supplement mappings (0xC0..0xFF)
+        map[0x00C0 - 0x00C0] = "A"; map[0x00C1 - 0x00C0] = "A"; map[0x00C2 - 0x00C0] = "A"; map[0x00C3 - 0x00C0] = "A"; map[0x00C4 - 0x00C0] = "A"; map[0x00C5 - 0x00C0] = "A";
```
tylerkaraszewski: I think this would be clearer as something like:
```
// À through Å
for (char32_t codePoint = 0x00C0; codePoint <= 0x00C5; ++codePoint) {
    map[codePoint - 0x00C0] = "A";
}
```

Etc through the mapping.

tylerkaraszewski (commented):
Only reviewing the deburring code, not the benchmarking code.

libstuff/SDeburr.cpp
roryabraham: I ended up removing this because I think it was essentially an LLM hallucination. Lodash doesn't handle this character.

libstuff/SDeburr.cpp
roryabraham: added!

libstuff/SDeburr.cpp
roryabraham: yeah, you mentioned this but I get compiler errors if I don't cast here. `input.c_str()` and `input.data()` both return `const char*`, not `const unsigned char*`. The `countl_one` C++ stdlib utility expects a `const unsigned char*`, so I couldn't figure out how to avoid the cast. A bit of research suggested that changing a `char*` to an `unsigned char*` is one of the primary use-cases for `reinterpret_cast` (`static_cast` didn't work).

lmk if you see something I don't here

libstuff/SDeburr.cpp
roryabraham: removed

libstuff/SDeburr.cpp
roryabraham: updated

libstuff/SDeburr.cpp
roryabraham: Well it returns `char*` because some unicode characters are represented as more than one ascii characters (`œ` becomes `oe`, for example).

I could make it return a C++ string though, which I think would simplify things.

benchmarks/main.cpp
roryabraham: done

libstuff/SDeburr.cpp
roryabraham: consolidating in the thread below

libstuff/SDeburr.h
roryabraham: Hmmm I tried implementing this but reverted it. I think it hurts readability, but more importantly falls apart once you get to https://en.wikipedia.org/wiki/Latin_Extended-A because it starts alternating between uppercase and lowercase.

image

libstuff/SDeburr.cpp
roryabraham: Hmmm the table is constexpr so you can't use a non-transient `std::string` in there. We could convert it to a string by just using the `string()` constructor, but that seems like unnecessary overhead vs just appending the `char*` to the output string directly.

tylerkaraszewski: I didn't realize until afterward that some of these map to multiple characters, so I guess it's fine to leave as-is.

libstuff/SDeburr.h
tylerkaraszewski: Ok, here's a suggestion that vastly improves readability at the cost of a single 192 bytes of memory:

```
// This value is picked as the last character past Latin_Extended-A.
array<const char*, 0x0180> map = {};
```

Then you drop the subtraction from the array.
```
map[0x00E0] = "a"; map[0x00E1] = "a"; map[0x00E2] = "a"; map[0x00E3] = "a"; map[0x00E4] = "a"; map[0x00E5] = "a";
```

You can shorten this up further with:
```
map[0x00E0] = map[0x00E1] = map[0x00E2] = map[0x00E3] = map[0x00E4] = map[0x00E5] = "a";
```

You can add another comment above it that says something like:
```
// Array indices are simply the unicode code points that map to the given character
```

You could also do it like this if it's clearer with a helper function:
```
void mapCodePoints(array<const char*, 0x0180>& map, const char* value, list<char32_t> codePoints) {
    for (auto codePoint : codePoints) {
        map[codePoint] = value;
    }
}
```

And then the initializations could look like:
```
mapCodePoints(map, "a", {0x00E0, 0x00E1, 0x00E2, 0x00E3, 0x00E4, 0x00E5});
```

Edit: see comment below about `constexpr` as well.

tylerkaraszewski: Also there's no real reason for this to be inline constexpr. It can be a normal static array and it will be fine.

You can do something like:

```
// In .h
static const array<const char*, 0x0180> UNICODE_TO_ASCII_MAP;
static decltype(UNICODE_TO_ASCII_MAP) initUnicodeToASCIIMap();
```

And:
```
// in .cpp
decltype(SDeburr::UNICODE_TO_ASCII_MAPs) SDeburr::UNICODE_TO_ASCII_MAPs = SDeburr::initUnicodeToASCIIMap();

decltype(SDeburr::UNICODE_TO_ASCII_MAPs) SDeburr::initUnicodeToASCIIMap() {
    array<const char*, 0x0180> map = {};
    
    // Any ititialization you want here, regardless of constexpr restrictions.
    
    return map;
}
```

I think it ends up easier to read and you don't really get any advantage here from `constexpr` anyway.

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,134 @@
...
+        if (sqlite3_value_type(argv[0]) == SQLITE_NULL) {
+            sqlite3_result_null(ctx);
+            return;
+        }
+        const unsigned char* text = sqlite3_value_text(argv[0]);
+        if (!text) {
+            sqlite3_result_null(ctx);
+            return;
+        }
+        string out = SDeburr::deburr(reinterpret_cast<const char*>(text));
```
tylerkaraszewski: If `SDeburr::deburr` just takes a `const unsigned char*` instead of a string, that simplifies a few things (including avoiding constructing a string with a copy of `text`)

libstuff/SDeburr.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (codepoint >= 0x00C0 && codepoint <= 0x017F) {
+        return SDeburr::UNICODE_TO_ASCII_MAP[codepoint - 0x00C0];
+    }
+    return nullptr;
+}
+
+string SDeburr::deburr(const string& input) {
+    const unsigned char* inputBytes = reinterpret_cast<const unsigned char*>(input.c_str());
+    string result;
+    result.reserve(input.size());
```
tylerkaraszewski: So I'm going to suggest changing the previous few lines to:

```
string SDeburr::deburr(const unsigned char* inputBytes) {
    string result;
    size_t inputLength = strlen(reinterpret_cast<const char*>(inputBytes));
    result.reserve(inputLength);
```

the main point of this is to let you pass `text` directly in `SDeburr::registerSQLite` without needing to duplicate it into a string first, as you never use that constructed string again in `deburr` except to get it's length.

This maybe cleans up the number or `reinterpret_cast`s you need, but there are still two if you do it this way. The one above for `strlen`, and one later for `result.append`.

You could also use `string SDeburr::deburr(const char* inputBytes)` and do the casts to `unsigned` instead, but I think this is a "6 of one, half a dozen of the other" situation.

tylerkaraszewski (changes requested):

libstuff/SDeburr.h
roryabraham: Good suggestions!

I had resisted wasting the 192 bytes (a previous version of the code even used pointer arithmetic to get a negative pointer) so you could index the array correctly without a subtraction, but honestly 192 bytes shouldn't matter. Much better now!

I implemented a version of the `mapCodePoints` solution, I think that makes things nice and readable. I also decided to just order all the code points numerically because that was the easiest way to cross-check it against the wikipedia page.

Regarding the constexpr, I didn't really find too much in the way of limitations (just use a `std::vector` instead of a `std::list` in `mapCodePoints`) so I left it as-is. I like how it's declared and initialized all in the same place. I did change `auto` to an explicit type, which I think gives the same clarity that your proposed alternative does.

libstuff/SDeburr.cpp
roryabraham: consolidating to the other thread

libstuff/SDeburr.cpp
roryabraham: ok, I've updated `SDeburr::deburr` to take a `const unsigned char*`. I also added a string-based overload for convenience. If we ever want to use this code outside of SQLite (which, to be fair I have no idea if we will), then it's nice to not have to do the `reinterpret_cast<const unsigned char*>(myString.c_str());`.

libstuff/SDeburr.cpp
roryabraham: I found something quite curious that I can't explain here. I added an `SWARN` in this block, and after doing that found that the benchmarks degraded by about 25%, even though the log is never printed to `/var/log/syslog`. I double-and-triple checked this - it's very reliable to reproduce. I also tried adding a regular `cout` in there and found that it wasn't printing anything.

So yeah, that's weird.

roryabraham: The fact that logging macros can impact performance, even if they're not executed, suggests to me that they'd be better off as normal functions or perhaps inline function (though my limited research on reddit threads suggests it's better to just let the compiler decide when functions should be inlined or not).

libstuff/SDeburr.h
```diff
@@ -0,0 +1,134 @@
...
+        mapCodePoints("T", {0x0162, 0x0164, 0x0166});
+        mapCodePoints("t", {0x0163, 0x0165, 0x0167});
+        mapCodePoints("U", {0x0168, 0x016A, 0x016C, 0x016E, 0x0170, 0x0172});
+        mapCodePoints("u", {0x0169, 0x016B, 0x016D, 0x016F, 0x0171, 0x0173});
+        mapCodePoints("W", {0x0174});
+        mapCodePoints("w", {0x0175});
+        mapCodePoints("Y", {0x0176, 0x0178});
+        mapCodePoints("y", {0x0177});
+        mapCodePoints("Z", {0x0179, 0x017B, 0x017D});
+        mapCodePoints("z", {0x017A, 0x017C, 0x017E});
```
tylerkaraszewski, line 128: I like this version a lot better!

tylerkaraszewski (dismissed):

inimaga (dismissed):

luacmartins (dismissed):

tylerkaraszewski:
@roryabraham - you might need to merge main

luacmartins:
It seems like we have unverified comments 🤔

roryabraham:
Fuck cursor for turning off commit signing

roryabraham:
Seems `HTTPSTest::testWaitForHTTPSRequests` is flaky, created https://github.com/Expensify/Expensify/issues/537766

PR #2263, opened 2025-08-29
Add support for querying connection id in MySQL plugin

chuckdries:
### Details

Implements a shim for `SELECT connection_id()`, which is the only missing feature that I couldn't figure out how to easily work around in my Beekeeper Studio fork

This PR is a subset of https://github.com/Expensify/Bedrock/pull/2262

### Fixed Issues
Fixes GH_LINK

### Tests
1. Check out my [Expensidev PR](https://github.com/Expensify/Expensidev/pull/966) and do `vagrant reload`
2. Check out and build this PR
3. Build and run [my fork of Beekeeper Studio](https://github.com/beekeeper-studio/beekeeper-studio/pull/3478)
4. Create a new connection in Beekeeper, select Bedrock, then in the connection details, use expensify.com.dev as the host and port 3307 for the Bedrock db and 3308 for the Auth db
_________

### DIFF
--- a/plugins/MySQL.cpp
+++ b/plugins/MySQL.cpp
@@ -387,7 +387,7 @@ void BedrockPlugin_MySQL::onPortRecv(STCPManager::Socket* s, SData& request) {
                 s->send(MySQLPacket::serializeQueryResponse(packet.sequenceID, result));
             // Add SHOW KEYS() support
 
-            } else if (SREMatch("^SELECT\\s+CONNECTION_ID\\(\\s*\\)(?:\\s+AS\\s+(\\w+))?\\s*;?$", SToUpper(query), false, false, nullptr)) {
+            } else if (SREMatch("^SELECT\\s+CONNECTION_ID\\(\\s*\\)(?:\\s+AS\\s+(\\w+))?\\s*;?$", SToUpper(query), false, false, &matches)) {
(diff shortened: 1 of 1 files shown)


plugins/MySQL.cpp
```diff
@@ -387,6 +387,23 @@ void BedrockPlugin_MySQL::onPortRecv(STCPManager::Socket* s, SData& request) {
...
+                // Return connection ID - handles SELECT connection_id(); and SELECT connection_id() AS alias;
+                SINFO("Responding with connection ID");
+                SQResultRow row;
+                row.push_back("1"); // Return a simple connection ID
+                vector<SQResultRow> rows = {row};
+                
+                // Extract the alias if present, otherwise use default column name
+                vector<string> matches;
+                string columnName = "connection_id()";
+                if (SREMatch("^SELECT\\s+CONNECTION_ID\\(\\s*\\)(?:\\s+AS\\s+(\\w+))?\\s*;?$", SToUpper(query), false, false, &matches) && matches.size() > 1) {
```
coleaeason: instead of doing this regex twice, i think you could just use the matches from the outer scope of the `if` on L279 in your regex on L390, then just check if matches is empty and set the column name.

coleaeason (changes requested):

coleaeason (approved):

PR #2273, opened 2025-09-05
Add log paramater types to the whitelist

deetergp:
### Fixed Issues
Part of https://github.com/Expensify/Expensify/issues/484579

### Tests

_________


flodnv (changes requested):
Please add Auth plugin specific params to the Auth whitelist.

deetergp:
> Please add Auth plugin specific params to the Auth whitelist.

Oh, nice! I traced the error message back to Bedrock and figured I had to add it there, but makes sense that we'd have a separate whitelist in Auth 👍

PR #2289, opened 2025-10-01
Pass 10k instructions param to sqlite3_progress_handler

flodnv:
### Details

image

### Fixed Issues
Fixes https://expensify.slack.com/archives/C094TGUTZ/p1759246098087409?thread_ts=1759244930.670939&cid=C094TGUTZ

### Tests

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -230,7 +230,7 @@ void SQLite::commonConstructorInitialization(bool hctree) {
 
     // We saw queries where the progress counter never exceeds 551,000, so we're setting it to a lower number
     // based on Richard Hipp's recommendation.
-    sqlite3_progress_handler(_db, 10'000, _progressHandlerCallback, this);
+    sqlite3_progress_handler(_db, 100'000, _progressHandlerCallback, this);
 
     // Setting a wal hook prevents auto-checkpointing.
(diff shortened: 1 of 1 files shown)


tylerkaraszewski (approved):

sqlitecluster/SQLite.cpp
```diff
@@ -228,9 +228,9 @@ void SQLite::commonConstructorInitialization(bool hctree) {
     // Register application-defined deburr function.
     SDeburr::registerSQLite(_db);
 
-    // I tested and found that we could set about 10,000,000 and the number of steps to run and get a callback once a
-    // second. This is set to be a bit more granular than that, which is probably adequate.
-    sqlite3_progress_handler(_db, 1'000'000, _progressHandlerCallback, this);
+    // We saw queries where the progress counter never exceeds 551,000, so we're setting it to a lower number
+    // based on Richard Hipp's recommendation.
+    sqlite3_progress_handler(_db, 10'000, _progressHandlerCallback, this);
```
danieldoglas: I'd decrease this to 100k just so we increase the number of times we're executing the callback by 10x instead of 100.
```suggestion
    sqlite3_progress_handler(_db, 100'000, _progressHandlerCallback, this);
```

flodnv: Yeah, I wasn't sure... we can go with this for now, since we only encountered this problem at 500k.

PR #2372, opened 2025-11-25
Fix nodes getting stuck shutting down and slow replication.

tylerkaraszewski:
### Details
So this does two main things.

First, it limits the amount of time we'll spend waiting for the replication queue to finish before falling out of `FOLLOWING` to 10 seconds. This is sort of a floor, a single slow replicated transaction can go past this limit, but the next one will not start.

We intentionally try and complete the replication queue when we stop following. If leader crashes, we want to make sure that at least *some other node* has all the commits, so we don't want to abandon these immediately. However, this issue is the opposite case, where a follower is well behind, it currently keeps attempting to apply the entire queue until it times out. The 10 second timeout fix should still handle the case where followers are in sync, but leader dies, without losing leader's last commits. At the same time, this allows slow nodes to shut down without getting stuck. They will sync whatever they were missing when they come back up.

The second thing this change does is largely disable the extra logging for slow HC-Tree commits. ~~This will still allow exactly 1 of these log lines per run, which should be enough to keep supplying the sqlite team with diagnostic info.~~ Edit: I turned them off completely in a subsequent commit.

These can take a *crazy long time*:
Screenshot 2025-11-25 at 16 27 15

These cause replication backups, and I think are the primary problem causing https://github.com/Expensify/Expensify/issues/558782

To be clear, we do these because commits themselves are slow, but the slow commits take 1-2 seconds, and then the diagnostic queries take 30+ seconds. 

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/560525
Fixes https://github.com/Expensify/Expensify/issues/558782

### Tests

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -826,6 +826,7 @@ int SQLite::commit(const string& description, const string& commandName, functio
     result = SQuery(_db, "COMMIT");
 
     // In HCTree mode, we log extra info for slow commits.
+    /* Currently disabled because the diagnostic query takes 30+s to run.
     if (_hctree) {
         uint64_t afterCommit = STimeNow();
         // log for any commit over 1 second.
@@ -840,6 +841,7 @@ int SQLite::commit(const string& description, const string& commandName, functio
             }
         }
     }
+    */
 
     _lastConflictPage = _conflictPage;
     _lastConflictLocation = _conflictLocation;
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -802,6 +802,7 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash) {
     return true;
 }
 
+static atomic<bool> __SLOW_COMMIT_LOG{false};
 int SQLite::commit(const string& description, const string& commandName, function<void()>* preCheckpointCallback) {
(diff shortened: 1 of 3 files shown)


tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

sqlitecluster/SQLite.cpp
```diff
@@ -826,6 +826,7 @@ int SQLite::commit(const string& description, const string& commandName, functio
     result = SQuery(_db, "COMMIT");
 
     // In HCTree mode, we log extra info for slow commits.
+    /* This is disabled because the diagnostic logging is unreasonably expensive (30+ seconds for the query).
```
flodnv, line 829: Should we delete this code then?

flodnv (commented):
> The second thing this change does is largely disable the extra logging for slow HC-Tree commits. This will still allow exactly 1 of these log lines per run, which should be enough to keep supplying the sqlite team with diagnostic info.

How is this PR doing that?

sqlitecluster/SQLite.cpp
tylerkaraszewski: I sort of want to keep it until the issue is resolved, there's an ongoing discussion with sqlite about it.

tylerkaraszewski:
@flodnv updated

flodnv (approved):

PR #2415, opened 2025-12-17
Only attempt actual send if socket connection is established

fukawi2:
Add a state check before attempting to send the HTTP CONNECT message to Squid.

Before:
- `send()` immediately called `S_sendconsume()` on `preSendBuffer`, **regardless of socket state**
- This caused `ENOTCONN` errors when socket was still in CONNECTING state

After:
  - `send()` now checks if `state >= Socket::State::CONNECTED` before calling `S_sendconsume()`
  - If still CONNECTING, returns true (success) and defers actual send
  - The poll loop will automatically retry once `postPoll()` transitions the socket to CONNECTED

Connect/Retry steps:

  1. Socket created → CONNECTING state
  2. `send()` called → CONNECT message buffered in preSendBuffer
  3. State check fails → returns true without attempting send
  4. Poll loop detects connection complete → state transitions to CONNECTED
  5. Poll loop calls `send()` again → state check passes → CONNECT message sent to Squid
  6. Normal proxy negotiation proceeds

This eliminates the race condition where send() was called before the TCP handshake to Squid completed, which caused sporadic ENOTCONN failures under load.

### Fixed Issues
https://github.com/Expensify/Expensify/issues/578735

### Tests

TODO

### DIFF
--- a/libstuff/SHTTPSProxySocket.cpp
+++ b/libstuff/SHTTPSProxySocket.cpp
@@ -37,13 +37,23 @@ bool SHTTPSProxySocket::send(size_t* bytesSentCount) {
     size_t oldSize = sendBuffer.size();
     size_t oldPreSendSize = preSendBuffer.size();
     if (oldPreSendSize) {
-        result = S_sendconsume(s, preSendBuffer);
-        size_t bytesSent = oldPreSendSize - preSendBuffer.size();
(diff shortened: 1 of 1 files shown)


fukawi2:
[Justin has put this PR up instead](https://github.com/Expensify/Bedrock/pull/2416).

PR #2499, opened 2026-02-06
Add JSON logging to Fluentd TCP socket for Bedrock over a MPSC ring buffer

rushatgabhane:
### Details
cc @nathanmetcalf 

adds the ability to log directly to Fluentd's TCP socket in native JSON format. Since logs are pre formatted as JSON, fluentd can forward them without regex parsing : )

Recconnects on failure.

Supports three log destinations via `-logDestination`                                                         
  - rsyslog - existing behavior                                                                                
  - fluentd - JSON logging to Fluentd only                                                                               
  - both - dual logging

  Fluentd options:                                                                                                       
  - fluentdHost (default: 127.0.0.1)                                                                                    
  - fluentdPort (default: 24224)                                                                                        
  - fluentdTag (default: bedrock)                                                                                       

**Log format**
```json
{
  "timestamp": "1772059866346128",
  "source": "bedrock", // hardcoded
  "request_id": "4FOxTC",
  "log_param": "rushat+1@expensify.com",
  "type_tag": "[info]",
  "thread_name": "socket1_cmd",
  "file": "SQLite.cpp",
  "line": "1085",
  "function": "logLastTransactionTiming",
  "message": "Transaction rollback with 1 read queries...",
  ... additional params passed by the app like command
}
```

**Field descriptions:**
- `request_id` - "xxxxxx" when unknown
- `log_param` - "rushat+1@exfy.com", "we@dont.know" when unknown
- `type_tag` - `[dbug]`, `[info]`, `[hmmm]`, `[warn]`, `[alrt]`, `[eror]`
- Additional params are included and redacted if not whitelisted

#### **Old** 
Rsyslog: bottleneck at 50k logs/sec per server.

#### This PR
Ring bufffer: measured support for 300k logs/sec per server.

### Ring buffer

The buffer is heap allocated. 
10M messages buffer. At 50K logs/sec, it'll take 200 seconds for it to become full and it'll fallback to syslog.

Please review these functions 
- `push(T&&)`
- `pop()` 

`isReady` flag prevents reading partially written slots

### Fluentd Logger                                                    
           
Please review these functions                                                                                                                            
- log(string&&) - pushes to ring buffer                                                                                        
- senderLoop() - Background thread writes to Fluentd TCP socket 
- sendAll() - Handles partial TCP writes
- Destructor - Drains buffer before shutdown  

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/594121

### Tests
1. Run Bedrock with `-logDestination both` by adding it in expensidev/config
2. Install fluentd on your VM and configure it to use TCP - https://docs.fluentd.org/input/tcp                                            
3. Verify logs go to both rsyslog and Fluentd                                       
4. Test reconnection by restarting Fluentd while Bedrock is running                                                    
5. Verify that you see json logs in fluentd

image

image

### Shutdown flushes the logs

https://github.com/user-attachments/assets/00faf182-5912-471c-befd-3c77a5037c6c

_________


rushatgabhane:
@MelvinBot review this PR please

rushatgabhane:
@MelvinBot review this PR please

libstuff/libstuff.cpp
```diff
@@ -280,6 +280,87 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+    STable record;
+    record["timestamp"] = to_string(time(nullptr));
+    record["priority"] = to_string(priority);
+    record["thread_name"] = SThreadLogName;
+    record["thread_prefix"] = SThreadLogPrefix;
+    record["process"] = SProcessName;
+    record["message"] = message;
+
+    for (const auto& [key, value] : params) {
+        record[key] = value;
```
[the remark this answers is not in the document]
rushatgabhane: resolved

libstuff/libstuff.cpp
```diff
@@ -280,6 +280,87 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+
+    // Reconnect if needed
+    if (SFluentdSocketFD == -1) {
+        if (!SFluentdConnect(lock)) {
+            return;
+        }
+    }
+
+    // Try to send the log over TCP. Close the socket on failure. It'll try to reconnect on next attempt
+    if (send(SFluentdSocketFD, json.c_str(), json.size(), MSG_NOSIGNAL) == -1) {
```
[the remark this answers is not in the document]
rushatgabhane: handled partial writes using `SendAll()`

libstuff/libstuff.cpp
```diff
@@ -280,6 +280,87 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+    STable record;
+    record["timestamp"] = to_string(time(nullptr));
+    record["priority"] = to_string(priority);
+    record["thread_name"] = SThreadLogName;
+    record["thread_prefix"] = SThreadLogPrefix;
+    record["process"] = SProcessName;
+    record["message"] = message;
+
+    for (const auto& [key, value] : params) {
+        record[key] = value;
```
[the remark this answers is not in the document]
rushatgabhane: resolved

libstuff/libstuff.cpp
```diff
@@ -280,6 +280,87 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+
+void SFluentdInitialize(const string& host, int port, const string& tag)
+{
+    lock_guard<mutex> lock(SFluentdSocketMutex);
+    SFluentdHost = host;
+    SFluentdPort = port;
+    SFluentdTag = tag;
+    SFluentdConnect(lock);
+    SFluentdConfigured.store(true);
+}
```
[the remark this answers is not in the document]
rushatgabhane: older diff

libstuff/libstuff.cpp
```diff
@@ -280,6 +280,87 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+    SFluentdSocketFD = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+    if (SFluentdSocketFD == -1) {
+        return false;
+    }
+
+    struct sockaddr_in addr;
+    memset(&addr, 0, sizeof(addr));
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(SFluentdPort);
+    inet_pton(AF_INET, SFluentdHost.c_str(), &addr.sin_addr);
```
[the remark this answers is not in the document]
rushatgabhane: invalid. we gonna use fluentd in localhost only

rushatgabhane:
@codex review

libstuff/SFluentdLogger.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+    int fd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+    if (fd == -1) {
+        return -1;
+    }
+
+    struct sockaddr_in addr = {};
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(port);
+    inet_pton(AF_INET, host.data(), &addr.sin_addr);
```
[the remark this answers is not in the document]
rushatgabhane: non issue because fluentd will be local

libstuff/SFluentdLogger.cpp
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        if (fd == -1) {
+            fd = openSocket();
+        }
+
+        if (!sendAll(fd, entry.value())) {
+            close(fd);
+            fd = -1;
+            syslog(LOG_WARNING, "%s", entry.value().data());
```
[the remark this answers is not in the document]
rushatgabhane: fixing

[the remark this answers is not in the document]
rushatgabhane: fixed

rushatgabhane:
@codex full review

rushatgabhane:
@codex fix style please

rushatgabhane:
@codex fix style please

libstuff/SFluentdLogger.cpp
```diff
@@ -0,0 +1,94 @@
...
+    if (fd == -1) {
+        return -1;
+    }
+
+    struct sockaddr_in addr = {};
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(port);
+    inet_pton(AF_INET, host.data(), &addr.sin_addr);
+
+    if (connect(fd, (struct sockaddr*) &addr, sizeof(addr)) == -1) {
```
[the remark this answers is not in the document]
rushatgabhane: i don't know if this is a real concern. OS' system call timeout should be sufficient no?

destructor waits because we wanna flush the buffer before exiting the thread. or we might lose logs

libstuff/SFluentdLogger.h
```diff
@@ -0,0 +1,35 @@
...
+#include <string>
+#include <thread>
+
+#include <libstuff/SRingBuffer.h>
+
+using namespace std;
+
+struct FluentdLogRecord
+{
+    int priority;
```
rushatgabhane: we need this separately even tho the info in the json because we need the priorirty when logging to fallback - syslog(priority, json)

libstuff/libstuff.cpp
```diff
@@ -280,6 +281,44 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
     }
 }
 
+void SSyslogNoop(int priority, const char* format, ...)
+{
+}
+
+static unique_ptr<SFluentdLogger> fluentdLogger;
```
tylerkaraszewski: Can we put this in `SFluentdLogger::fluentdLogger`

BedrockServer.cpp
```diff
@@ -1299,6 +1299,19 @@ BedrockServer::BedrockServer(const SData& args_)
...
+    string logDestination = args.isSet("-logDestination") ? args["-logDestination"] : "rsyslog";
+    if (logDestination == "fluentd" || logDestination == "both") {
+        string host = args.isSet("-fluentdHost") ? args["-fluentdHost"] : "127.0.0.1";
+        int port = args.isSet("-fluentdPort") ? SToInt(args["-fluentdPort"]) : 24224;
+        string tag = args.isSet("-fluentdTag") ? args["-fluentdTag"] : "bedrock";
+        SFluentdInitialize(host, port, tag);
+    }
+    if (logDestination == "fluentd") {
+        SSyslogFunc = &SSyslogNoop;
+    }
```
tylerkaraszewski: I think this block should be in `main.cpp`. Probably the one above it should be, as well, so I see how it ended up here, but it doesn't seem like the correct place for it.

libstuff/libstuff.cpp
```diff
@@ -280,6 +281,44 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+static string fluentdTag;
+
+void SFluentdInitialize(const string& host, int port, const string& tag)
```
tylerkaraszewski: port should be `unsigned short` or `in_port_t` but it doesn't really matter that much.

libstuff/libstuff.cpp
```diff
@@ -280,6 +281,44 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+
+void SFluentdLog(int priority, const string& message, const STable& params)
+{
+    if (!fluentdLogger) {
+        return;
+    }
+
+    STable record;
+    record["_timestamp"] = to_string(time(nullptr));
+    record["_priority"] = to_string(priority);
```
tylerkaraszewski: Should this be `INFO`, `WARN` etc?

libstuff/libstuff.cpp
```diff
@@ -280,6 +281,44 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+}
+
+void SFluentdLog(int priority, const string& message, const STable& params)
+{
+    if (!fluentdLogger) {
+        return;
+    }
+
+    STable record;
+    record["_timestamp"] = to_string(time(nullptr));
```
tylerkaraszewski: This is only to the second, seems not granular enough, right?

libstuff/libstuff.cpp
```diff
@@ -280,6 +281,44 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+static unique_ptr<SFluentdLogger> fluentdLogger;
+static string fluentdTag;
+
+void SFluentdInitialize(const string& host, int port, const string& tag)
+{
+    fluentdTag = tag;
+    fluentdLogger = make_unique<SFluentdLogger>(host, port);
+}
+
+void SFluentdLog(int priority, const string& message, const STable& params)
```
tylerkaraszewski: I wonder if this could take params as an r-value reference and avoid copying it on every logline.

libstuff/SRingBuffer.h
```diff
@@ -0,0 +1,71 @@
...
+        head.store(currentHead + 1, memory_order_release);
+
+        return bufferData;
+    }
+
+private:
+    array<BufferElement, C> buffer;
+
+    // Single consumer reads from here
+    atomic<size_t> head{0};
```
tylerkaraszewski: So there is no canonical way to determine if `head` versus `tail` is the reader or writer here so these names aren't wrong, but they're counterintuitive to me, which might not be true for everyone. I'm going to suggest renaming them `readIndex` and `writeIndex`

libstuff/libstuff.h
```diff
@@ -252,12 +252,27 @@ void SLogStackTrace(int level = LOG_WARNING);
 // This method will allow plugins to whitelist log params they need to log.
 void SWhitelistLogParams(const set<string>& params);
 
+// Check if a log param is in the whitelist.
+bool SIsLogParamWhitelisted(const string& key);
```
tylerkaraszewski, line 257: Do we need this considering it's a one-liner called in one place?

libstuff/SFluentdLogger.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    if (fd != -1) {
+        close(fd);
+    }
+}
+
+bool SFluentdLogger::log(int priority, string&& json)
+{
+    FluentdLogRecord record{priority, move(json)};
+    return buffer->push(move(record));
```
tylerkaraszewski: Can combine these in one line.

libstuff/SFluentdLogger.cpp
```diff
@@ -0,0 +1,94 @@
...
+    int fd = -1;
+
+    while (true) {
+        auto entry = buffer->pop();
+
+        if (!entry.has_value()) {
+            if (!running.load()) {
+                break;
+            }
+            this_thread::sleep_for(chrono::milliseconds(1));
```
tylerkaraszewski: I think we can avoid this entirely by using atomic::wait() and atomic::notify_one(). See comments in `SRingBuffer.h` about that.

We probably need to one more method to detect when the queue is empty when we're shutting down in order to not get stuck in `pop` at shutdown after the last message.

libstuff/SRingBuffer.h
```diff
@@ -0,0 +1,71 @@
...
+        buffer[index].data = move(data);
+        buffer[index].isReady.store(true, memory_order_release);
+
+        return true;
+    }
+
+    optional<T> pop()
+    {
+        size_t currentHead = head.load();
+        size_t index = currentHead % C;
```
tylerkaraszewski: I think if we add:
```
buffer[index].isReady.wait(false, memory_order_acquire);
```

Below this line, we can avoid the case where we return nullopt and need to poll.

libstuff/SRingBuffer.h
```diff
@@ -0,0 +1,71 @@
...
+                return false;
+            }
+            if (tail.compare_exchange_weak(currentTail, currentTail + 1, memory_order_acq_rel, memory_order_relaxed)) {
+                break;
+            }
+        }
+
+        size_t index = currentTail % C;
+        buffer[index].data = move(data);
+        buffer[index].isReady.store(true, memory_order_release);
```
tylerkaraszewski: I think we can call:
```
buffer[index].isReady.notify_one();
```

Below this line to notify the writer that it has work to do.

tylerkaraszewski (changes requested):

libstuff/SFluentdLogger.cpp
rushatgabhane: good idea!

libstuff/SRingBuffer.h
rushatgabhane: fair enough, will change it! i was looking at it like a queue so writing to tail made sense to me

libstuff/libstuff.cpp
rushatgabhane: good catch. fixed it to transfer ownership instead of copying

libstuff/libstuff.h
rushatgabhane: yes we need it because `PARAMS_WHITELIST` is static. And I think we shouldn't expose `PARAMS_WHITELIST` to other classes because it could be modified unintentionally 

let me know if we should `extern` it or something

libstuff/libstuff.cpp
rushatgabhane: changed it to use STimeNow(). it uses micro secs

libstuff/libstuff.cpp
rushatgabhane: done

libstuff/SFluentdLogger.cpp
rushatgabhane: done

libstuff/libstuff.h
tylerkaraszewski: Hmm, I was going to say "well, let's make it `static const` then so that's not an issue, but then I see we allow plugins to modify this and it's not thread-safe at all, which maybe we should address.

@MelvinBot can you create an issue addressing thread-safety of `SWhitelistLogParams()`?

rushatgabhane: Sadly Melvin bot doesn't work for Bedrock

tylerkaraszewski: boo melvin.

tylerkaraszewski: Made an issue here: https://github.com/Expensify/Expensify/issues/602617

rushatgabhane:
still working on https://github.com/Expensify/Bedrock/pull/2499#discussion_r2829959147

will take another hour

libstuff/SFluentdLogger.cpp
rushatgabhane: Made some changes to avoid the polling loop in the sender thread. Used your suggestion `atomic::wait() and atomic::notify_one()`.

> We probably need to one more method to detect when the queue is empty when we're shutting down in order to not get stuck in pop at shutdown after the last message

Yeahh added a new `State::SHUTDOWN`, and it calls `notify_one()` to wake up the and flush the buffer

rushatgabhane:
@codex review

libstuff/SRingBuffer.h
```diff
@@ -0,0 +1,101 @@
...
+        atomic<State> state{State::Empty};
+    };
+
+    bool push(T&& data)
+    {
+        size_t currentWriteIndex = writeIndex.load(memory_order_relaxed);
+
+        while (true) {
+            if (currentWriteIndex - readIndex.load(memory_order_acquire) >= C) {
+                return false;
```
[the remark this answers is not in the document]
rushatgabhane: this is acceptable. will run at static destruction at program shutdown.  

_please correct me if im wrong_

rushatgabhane:
@tylerkaraszewski review comments addressed. Looking forward to your feedback : )

im out of time for the day so will have to check sat/sun

libstuff/SFluentdLogger.cpp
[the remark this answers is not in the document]
tylerkaraszewski: Let's just try it as-is and see if we need to improve it in the future.

libstuff/SRingBuffer.h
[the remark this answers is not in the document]
tylerkaraszewski: This will run at static destruction at program shutdown. Nothing should still be logging then anyway.

tylerkaraszewski (dismissed):
I think I'm OK with this.

flodnv (commented):
Started reviewing here: https://expensify.slack.com/archives/C05CBC62HGW/p1772042190630239?thread_ts=1771879554.162819&cid=C05CBC62HGW

rushatgabhane:
Addressed _some_ of Florent's feedback:

1. Renamed thread_prefix to request_id 
2. Fixed message to only contain the actual log message (removed [info] prefix and SWHEREAMI metadata)
3. Added type_tag field with the log level ([info], [warn]. This should be named `level` / `priority` imho but I'm matching what we have in Web-E
4. Added filename field 
5. Removed process field since it's redundant with tag
6. email - added email to logs in the field `logParam`. It's named that way when [it is received from php](https://github.com/Expensify/Auth/blob/879c8ec1bc07ab7e91922e94b8643e8794c1bb11/auth/AuthCommand.cpp#L86). So didn't name it as email

rushatgabhane:
improved log format as per @flodnv's feedback - 


image

image




**Log format**
```json
{
  "timestamp": "1772059866346128",
  "source": "bedrock",
  "request_id": "4FOxTC",
  "log_param": "rushat+1@expensify.com",
  "type_tag": "[info]",
  "thread_name": "socket1_cmd",
  "file": "SQLite.cpp",
  "line": "1085",
  "function": "logLastTransactionTiming",
  "message": "Transaction rollback with 1 read queries...",
  ... additional params passed by the app like command
}
```

**Field descriptions:**
- `request_id` - "xxxxxx" when unknown
- `log_param` - "rushat+1@exfy.com", "we@dont.know" when unknown
- `type_tag` - `[dbug]`, `[info]`, `[hmmm]`, `[warn]`, `[alrt]`, `[eror]`
- Additional params are included and redacted if not whitelisted

rushatgabhane:
this is syslog

image

rushatgabhane:
@codex review

rushatgabhane:
@flodnv all yours. latest screens here - https://github.com/Expensify/Bedrock/pull/2499#issuecomment-3962631109

libstuff/libstuff.cpp
```diff
@@ -280,6 +281,47 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+    record["log_param"] = SThreadLogParam;
+    record["type_tag"] = typeTag;
+    record["thread_name"] = SThreadLogName;
+    record["file"] = file;
+    record["line"] = to_string(line);
+    record["function"] = function;
+    record["message"] = move(message);
+
+    for (auto& [key, value] : params) {
+        record[key] = SIsLogParamWhitelisted(key) ? move(value) : "<REDACTED>";
```
flodnv, line 313: This isn't the same format as web where there's a `params` key containing an array(?) of the params: https://github.com/Expensify/Web-Expensify/pull/49231/changes#diff-12e5a719fc2a0fecd83693b14884bcf35cc1bca5d374cd7d5c14632b85efbae0R799

libstuff/libstuff.cpp
```diff
@@ -280,6 +281,47 @@ void SSyslogSocketDirect(int priority, const char* format, ...)
...
+    }
+
+    // Forward protocol: [tag, time (seconds), record]
+    string json = "[\"" + SFluentdLogger::tag + "\"," + to_string(time(nullptr)) + "," + SComposeJSONObject(record, true) + "]\n";
+
+    if (!SFluentdLogger::instance->log(priority, move(json))) {
+        // Fallback to syslog if fluentdLogger fails to log
```
flodnv, line 320: NAB: this is quite strange to me.

main.cpp
```diff
@@ -297,6 +297,18 @@ int main(int argc, char* argv[])
         SLogLevel(LOG_WARNING);
     }
 
+    // Log destination: rsyslog, fluentd, or both. Default is rsyslog.
+    string logDestination = args.isSet("-logDestination") ? args["-logDestination"] : "rsyslog";
+    if (logDestination == "fluentd" || logDestination == "both") {
+        string host = args.isSet("-fluentdHost") ? args["-fluentdHost"] : "127.0.0.1";
+        in_port_t port = args.isSet("-fluentdPort") ? SToInt(args["-fluentdPort"]) : 24224;
+        string tag = args.isSet("-fluentdTag") ? args["-fluentdTag"] : "bedrock";
+        SFluentdInitialize(host, port, tag);
```
flodnv, line 306: Should we fallback to rsyslog if we fail to initialize?

flodnv (commented):
Thanks!

main.cpp
rushatgabhane: no, i don't think so. the config is wrong if that is happening

libstuff/libstuff.cpp
rushatgabhane: Web-E puts the params directly into the root too when it is an array - https://github.com/Expensify/Web-Expensify/pull/49231/changes#diff-12e5a719fc2a0fecd83693b14884bcf35cc1bca5d374cd7d5c14632b85efbae0R796


And puts it in `params` key for strings.
 
Do you want us to put all params under a params key instead of merging to root?

```cpp
if (!params.empty()) {                                                                                                                                                                   
    STable sanitizedParams;                                                                                                                                                              
    for (auto& [key, value] : params) {                                                                                                                                                  
        sanitizedParams[key] = SIsLogParamWhitelisted(key) ? move(value) : "<REDACTED>";                                                                                                 
    }                                                                                                                                                                                    
    record["params"] = SComposeJSONObject(sanitizedParams);                                                                                                                              
}                                                                                                                                                                                        

```

rushatgabhane: so the log will look something like this

```
{                                                                                                                                                                                        
  "timestamp": "123456",                                                                                                                                                                 
  "message": "hello",                                                                                                                                                                    
  "params": {                                                                                                                                                                            
    "accountID": "123",                                                                                                                                                                  
    "email": "test@example.com"                                                                                                                                                          
  }                                                                                                                                                                                      
}                                                                                                                                                                                        
```

rushatgabhane:
@flodnv wdyt? https://github.com/Expensify/Bedrock/pull/2499#discussion_r2862334657

libstuff/libstuff.cpp
flodnv: Oh hm, it may be better to always have them in `params` for consistency; what do you think @nathanmetcalf ?

nathanmetcalf: Web-E uses the root of the JSON array for params. EG: 
```
{                                                                                                                                                                                        
  "timestamp": "123456",                                                                                                                                                                 
  "message": "hello",                                                                                                                                                                    
  "accountID": "123",                                                                                                                                                                  
  "email": "test@example.com"                                                                                                                                                          
  "foo": "foo"
  "bar":"bar"
} 
```

We should follow this standard here too - That is consistant with Web-E

main.cpp
nathanmetcalf: I Agree, we shouldn't fall back to rsyslog. I can imagine there is a situation in the future where we might fall back to another instance of fluentd - for a HA setup.  I wouln't over complicated it for now.

libstuff/libstuff.cpp
rushatgabhane: cool, that is what we're doing in Bedrock then :)

flodnv (approved):

tylerkaraszewski (approved):

PR #2504, opened 2026-02-09
Track max transaction elapsed time across multiple transactions

danieldoglas:
## Summary
- `_totalTransactionElapsed` was only storing the last transaction's time when a command ran multiple transactions (e.g., when peek makes httpsRequests and the transaction is closed and reopened). This meant the slowest transaction could be silently lost in timing logs.
- Changed rollback and commit to use `max()` so the highest transaction time is preserved across all transactions within a single command.
- Added `resetMaxTransactionElapsed()` and call it at db assignment boundaries (pool handle initialization and sync thread command loop) so stale timing doesn't carry over between commands on the same db handle.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


danieldoglas:
NVM this, I forgot that we log the line every time we call rollback, so we are keepting track of all transactions, not only the last one.

PR #2555, opened 2026-03-19
Add blocking queue rate limit

stitesExpensify:
### Details
This PR blocks users from adding to the blocking queue if they already have too many commands in the queue. This prevents a single user from being able to DOS us (intentionally or on accident). 

Once the queue is empty for 30 seconds, we reset the block list and allow those users to continue using the product per usual

### Fixed Issues
https://github.com/Expensify/Expensify/issues/568969

### Tests
Added automated tests
_________


BedrockServer.cpp
```diff
@@ -1241,7 +1313,7 @@ BedrockServer::BedrockServer(const SData& args_)
     _isCommandPortLikelyBlocked(false),
     _syncLoopShouldBeRunning(true), _syncNode(nullptr), _clusterMessenger(nullptr), _shutdownState(RUNNING),
     _multiWriteEnabled(args.test("-enableMultiWrite")), _enableConflictPageLocks(args.test("-enableConflictPageLocks")), _shouldBackup(false), _detach(args.isSet("-bootstrap")),
-    _controlPort(nullptr), _commandPortPublic(nullptr), _commandPortPrivate(nullptr), _maxConflictRetries(3),
+    _controlPort(nullptr), _commandPortPublic(nullptr), _commandPortPrivate(nullptr), _maxConflictRetries(3), _maxBlockingQueuePerUser(0),
```
stitesExpensify: I have this set to 0 right now, we can update it to whatever we want using `SetBlockingRateLimit` after deployment of this and the auth PR

stitesExpensify:
@codex review again

stitesExpensify:
This is my first bedrock PR ever, let me know if anything needs to be changed! I realize that a lot of the comments are a mixture of "what is happening" and "why is it happening", but it made it a lot more clear to me. Happy to remove them if they are unnecessary to people more familiar with the codebase

BedrockCommand.h
```diff
@@ -293,6 +293,10 @@ class BedrockCommand : public SQLiteCommand {
     };
     CrashMap crashIdentifyingValues;
 
+    // An identifier for rate-limiting commands in the blocking queue.
+    // Typically populated with an accountID by Auth. Empty means skip rate limiting.
```
tylerkaraszewski: I would refrain from putting comments about Auth in Bedrock, make this more generic.

BedrockServer.h
```diff
@@ -418,13 +418,23 @@ class BedrockServer : public SQLiteServer {
     // particular command for it count as a match likely to cause a crash.
     map<string, set<STable>> _crashCommands;
 
+    // Per-user blocking queue rate limiting.
+    shared_timed_mutex _blockingRateLimitMutex;
+    map<string, int> _blockingQueueUserCounts;
+    set<string> _blockedUsers;
+    atomic<int> _maxBlockingQueuePerUser{0};
```
tylerkaraszewski: Much like `BedrockCommand` uses `blockingIdentifier` here (as there's no requirement that this is a "user"), these varaibes names should be named something else.

BedrockServer.h
```diff
@@ -418,13 +418,23 @@ class BedrockServer : public SQLiteServer {
...
+
     // Returns whether or not the command was a status or control command. If it was, it will have already been handled
     // and responded to upon return
     bool _handleIfStatusOrControlCommand(unique_ptr<BedrockCommand>& command);
 
     // Check a command against the list of crash commands, and return whether we think the command would crash.
     bool _wouldCrash(const unique_ptr<BedrockCommand>& command);
 
+    // Decrement the blocking queue rate limit count for a command that is leaving the blocking queue.
+    void _decrementBlockingQueueCount(const unique_ptr<BedrockCommand>& command);
```
tylerkaraszewski: This should not be here, it should be in `BedrockBlockingCommandQueue`, which should decrement this atomically with dequeue.

In fact, basically all of the changes to `BedrockServer.h` should be in `BedrockBlockingCommandQueue`, which also eliminates the second mutex requirement (i.e., we wouldn't need `_blockingRateLimitMutex` because we can use `_queueMutex` which already exists).

tylerkaraszewski (changes requested):

BedrockServer.h
stitesExpensify: Updated! Moved pretty much everything to `BedrockBlockingCommandQueue`. Ready for another review

BedrockBlockingCommandQueue.h
```diff
@@ -10,4 +10,30 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
     static void startTiming(unique_ptr<BedrockCommand>& command);
     static void stopTiming(unique_ptr<BedrockCommand>& command);
+
+    // Check per-identifier rate limit before pushing a command to the blocking queue.
+    // Returns true if the command was rejected (503 set on response), caller should reply and return.
+    // Returns false if the command was pushed onto the queue (command is moved, caller should not use it).
+    bool checkRateLimitAndPush(unique_ptr<BedrockCommand>& command, bool isBlocking);
+
+    // Decrement the rate limit count for a command leaving the blocking queue.
+    void decrementCount(const unique_ptr<BedrockCommand>& command);
```
tylerkaraszewski: This shouldn't be a thing, this should just happen when a command is removed from the queue.

BedrockBlockingCommandQueue.h
```diff
@@ -10,4 +10,30 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+
+    // Clear all rate limiting state.
+    void resetRateLimitState();
+
+    // Populate the given table with rate limiting status info for the Status command.
+    void populateRateLimitStatus(STable& content);
+
+    // Set the max commands per identifier threshold. Returns the previous value.
+    int setMaxPerIdentifier(int value);
```
tylerkaraszewski: use `size_t` for this.

BedrockServer.cpp
```diff
@@ -907,6 +917,10 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
                 core.prePeekCommand(command, isBlocking);
 
                 if (command->complete) {
+                    // Command completed in prePeek and is leaving the blocking queue, so decrement its rate limit count.
+                    if (isBlocking) {
+                        _blockingCommandQueue.decrementCount(command);
```
tylerkaraszewski: We shouldn't need any of this, removing a command from the queue should decrement its count.

BedrockBlockingCommandQueue.cpp
```diff
@@ -14,3 +14,116 @@ BedrockBlockingCommandQueue::BedrockBlockingCommandQueue() :
...
+    lock_guard<decltype(_queueMutex)> lock(_queueMutex);
+
+    int maxPerIdentifier = _maxPerIdentifier.load();
+    if (maxPerIdentifier > 0 && !command->blockingIdentifier.empty()) {
+        // Clear blocks if the blocking queue has been empty for 30 seconds.
+        uint64_t emptyTime = _emptyTime.load();
+        if (emptyTime > 0 && STimeNow() - emptyTime >= 30'000'000) {
+            _blockedIdentifiers.clear();
+            _identifierCounts.clear();
+            _emptyTime.store(0);
```
tylerkaraszewski: `_emptyTime` should just get set whenever the queue size drops to zero, whether we call `clear()` on it or the last item is removed.

BedrockBlockingCommandQueue.cpp
```diff
@@ -14,3 +14,116 @@ BedrockBlockingCommandQueue::BedrockBlockingCommandQueue() :
...
+            SALERT("Blocking queue rate limit: rejecting '" << command->request.methodLine
+                   << "' for identifier '" << command->blockingIdentifier << "'");
+            command->response.methodLine = "503 Blocking queue rate limited";
+            command->complete = true;
+            return true;
+        }
+
+        // Only count on first entry from worker threads, not when the blocking commit
+        // thread re-queues a command that conflicted again.
+        if (!isBlocking) {
```
tylerkaraszewski: I don't understand at all what this is supposed to do. If a command is in the blocking queue, it counts, if it's not, then it doesn't.

tylerkaraszewski (changes requested):

stitesExpensify:
TYVM for your reviews @tylerkaraszewski. Ready for another one 😄

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+                    }
+                }
+            });
+        }
+
+        for (thread& t : threads) {
+            t.join();
+        }
+
+        cout << "[BlockingQueueRateLimitTest] 200s: " << count200.load() << ", 503s: " << count503.load() << endl;
```
tylerkaraszewski: These lines are fine for development but should be removed before merging. The expectation is that the test suite will pass the vast majority of the time, and so what the user wants to see is a stream of ✅ without a bunch of diagnostic info for tests that actually passed, which is especially true when some other test failed. Only print extra diagnostic info in the case that the test is failing.

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+            SData resetConflict("SetConflictParams");
+            resetConflict["MaxConflictRetries"] = "3";
+            node.executeWaitVerifyContent(resetConflict, "200", true);
+        }
+
+        // Verify the previously blocked user can send commands again.
+        SData cmd("idcollision");
+        cmd["writeConsistency"] = "ASYNC";
+        cmd["blockingIdentifier"] = "testuser";
+        cmd["value"] = "after-clear";
```
tylerkaraszewski: Looks like this line doesn't do anything?

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+        list<thread> threads;
+
+        for (int i : {0, 1, 2}) {
+            threads.emplace_back([this, i, &count503, &count200]() {
+                BedrockTester& node = tester->getTester(i);
+
+                vector<SData> requests;
+                for (int j = 0; j < 200; j++) {
+                    SData cmd("idcollision");
+                    cmd["writeConsistency"] = "ASYNC";
```
tylerkaraszewski: Commmands are ASYNC by default and have been for a long time, you shouldn't need to specify this in new code.

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+        atomic<int> count503(0);
+        atomic<int> count200(0);
+        list<thread> threads;
+
+        for (int i : {0, 1, 2}) {
+            threads.emplace_back([this, i, &count503, &count200]() {
+                BedrockTester& node = tester->getTester(i);
+
+                vector<SData> requests;
+                for (int j = 0; j < 200; j++) {
```
tylerkaraszewski, line 122: How long does this test take to run? Would setting this to a smaller number be plenty adequate? If the test runs in 1 second, it doesn't really matter, but if it takes 15 seconds, perhaps we could set this to 50 or something?

BedrockServer.cpp
```diff
@@ -1084,7 +1086,12 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
         if (maxRetries || _shutdownState.load() != RUNNING) {
             if (command->processCount > maxRetries) {
                 SINFO("Max retries (" << maxRetries << ") hit in worker, sending '" << command->request.methodLine << "' to blocking queue with size " << _blockingCommandQueue.size());
-                _blockingCommandQueue.push(move(command));
+                if (isBlocking) {
```
tylerkaraszewski: What is this check? If we're already in the blocking commit queue and we get a conflict, run in the blocking queue again? 

1. How would that happen?
2. Why would we skip the rate limit in that case?

BedrockBlockingCommandQueue.h
```diff
@@ -10,4 +10,30 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+    bool checkRateLimitAndPush(unique_ptr<BedrockCommand>& command);
+
+    // Clear the queue and all rate limiting state.
+    void clear();
+
+    // Clear all rate limiting state (blocked identifiers, counts, empty time).
+    void resetRateLimitState();
+
+    // Populate the given table with rate limiting status info for the Status command.
+    void populateRateLimitStatus(STable& content);
```
tylerkaraszewski: Instead of taking a table by reference here, just return a new table. You can just call the method "getState()" and the caller can do what they want with the data.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+}
+
+void BedrockBlockingCommandQueue::clearBlocks()
 {
+    lock_guard<decltype(_queueMutex)> lock(_queueMutex);
+    size_t cleared = _blockedIdentifiers.size();
+    _blockedIdentifiers.clear();
+    _identifierCounts.clear();
+    _emptyTime.store(0);
+    SINFO("Manually cleared " << cleared << " blocked identifiers.");
```
tylerkaraszewski: Having this function as well as `resetRateLimitState` that do basically the same thing is confusing. And now there are 3 different `clear` functions.

I would just have two:

```
clear(); // clear everything, including the content of the queue
```

```
clearRateLimits() // resets the rate limit counters without emptying the queue. Returns the number of unblocked identifiers.
```

The caller for `clearRateLimits` can decide if it cares about the number of unblocked identifiers or not.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+            countsTable[p.first] = to_string(p.second);
+        }
+        content["blockingQueueIdentifierCounts"] = SComposeJSONObject(countsTable);
+    }
+}
+
+size_t BedrockBlockingCommandQueue::setMaxPerIdentifier(size_t value)
+{
+    size_t previous = _maxPerIdentifier.load();
+    _maxPerIdentifier.store(value);
```
tylerkaraszewski: Use `std::atomic<T>::exchange` instead of separate load and store. This prevents the value changing between when you loaded it and when you stored the new value.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+    _startFunction(command);
+    _lookupByTimeout.insert(make_pair(timeout, make_pair(priority, scheduledTime)));
+    _queue[priority].emplace(scheduledTime, ItemTimeoutPair(move(command), timeout));
+    _queueCondition.notify_one();
+    return false;
+}
+
+void BedrockBlockingCommandQueue::clear()
+{
+    lock_guard<decltype(_queueMutex)> lock(_queueMutex);
```
tylerkaraszewski: I would be somewhat inclined to change the underlying type of `_queueMutex` to `recursive_mutex` so that below here you can just call:
```
clearRateLimits(); // Note rename from a different comment in this review.
BedrockCommandQueue:: clear();
```

Instead of repeating the implementation of the parent class.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+            _blockedIdentifiers.insert(command->blockingIdentifier);
+            SALERT("Blocking queue rate limit: flagging identifier '" << command->blockingIdentifier
+                   << "' with " << count << " commands in blocking queue (threshold: " << maxPerIdentifier << ")");
+        }
+    }
+
+    // Reset empty time since a command is entering the queue.
+    _emptyTime.store(0);
+
+    // Inline the push to avoid double-locking _queueMutex.
```
tylerkaraszewski: I think this is another thing we could simplify if we change `_queueMutex` to `recursive_mutex` where we can just call the parent class's function.

BedrockBlockingCommandQueue.h
```diff
@@ -10,4 +10,30 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
     // Functions to start and stop timing on the commands when they're inserted/removed from the queue.
     static void startTiming(unique_ptr<BedrockCommand>& command);
     static void stopTiming(unique_ptr<BedrockCommand>& command);
+
+    // Check per-identifier rate limit before pushing a command to the blocking queue.
+    // Returns true if the command was rejected (503 set on response), caller should reply and return.
+    // Returns false if the command was pushed onto the queue (command is moved, caller should not use it).
+    bool checkRateLimitAndPush(unique_ptr<BedrockCommand>& command);
```
tylerkaraszewski: I would be inclined to change the interface of this to simply override the parent class:
```
void push(unique_ptr<BedrockCommand>&& command);
```
(note that takes an r-value reference with the double `&&`)

Then we'd return `void`. The implementation could just `STHROW("503 whatever")`. We only call this in one place, we could catch this there and set the response there.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+        uint64_t emptyTime = _emptyTime.load();
+        if (emptyTime > 0 && STimeNow() - emptyTime >= 30'000'000) {
+            _blockedIdentifiers.clear();
+            _identifierCounts.clear();
+            _emptyTime.store(0);
+        }
+
+        if (_blockedIdentifiers.count(command->blockingIdentifier)) {
+            SALERT("Blocking queue rate limit: rejecting '" << command->request.methodLine
+                   << "' for identifier '" << command->blockingIdentifier << "'");
```
tylerkaraszewski: This alert will create bugbot issues every time this happens. Is that overkill?

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+            command->response.methodLine = "503 Blocking queue rate limited";
```
tylerkaraszewski: See other comment, but what if we throw here instead?

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+        count++;
+        if (count >= maxPerIdentifier) {
+            _blockedIdentifiers.insert(command->blockingIdentifier);
+            SALERT("Blocking queue rate limit: flagging identifier '" << command->blockingIdentifier
+                   << "' with " << count << " commands in blocking queue (threshold: " << maxPerIdentifier << ")");
+        }
+    }
+
+    // Reset empty time since a command is entering the queue.
+    _emptyTime.store(0);
```
tylerkaraszewski: This is entirely wrong, right? The queue is now *not* empty.

BedrockBlockingCommandQueue.h
```diff
@@ -10,4 +10,30 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+
+    // Populate the given table with rate limiting status info for the Status command.
+    void populateRateLimitStatus(STable& content);
+
+    // Set the max commands per identifier threshold. Returns the previous value.
+    size_t setMaxPerIdentifier(size_t value);
+
+    // Clear all blocked identifiers and counts.
+    void clearBlocks();
+
```
tylerkaraszewski, line 40: Don't we need to override `T get(uint64_t waitUS = 0, bool loggingEnabled = false);` from `SScheduledPriorityQueue` such that it decrements the count for the item removed from the queue? As it is, where do we ever reduce the count of items in `_identifierCounts`?

Also, that seems like the right place to set `_emptyTime`.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,131 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+    // Track when the queue becomes empty for auto-clearing blocks.
+    size_t queueSize = 0;
+    for (const auto& q : _queue) {
+        queueSize += q.second.size();
+    }
+    if (queueSize == 0 && _emptyTime.load() == 0) {
+        _emptyTime.store(STimeNow());
+    }
+}
+    )
```
tylerkaraszewski: Instead of doing the above, I think it's clearer to override `get()`.

tylerkaraszewski (changes requested):
The comments in this review were left sort of out-of-order. Read through them once and see if any later ones clarify earlier ones.

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
stitesExpensify: It takes less than 1.2s. I tried with smaller amounts and it wasn't consistent.

BedrockBlockingCommandQueue.h
stitesExpensify: This is now handled in dequeue

BedrockBlockingCommandQueue.cpp
stitesExpensify: Yep you're right, changed this to `0` (it was `STimeNow` before which was entirely wrong)

stitesExpensify:
Okay I think this is ready for another review! The build is failing, but it appears to be due to `ExpensifyBackupManager` (it is failing on main too)

BedrockServer.cpp
stitesExpensify: You were correct, that would not happen. This has been removed

BedrockBlockingCommandQueue.cpp
stitesExpensify: This comment is pretty out of date at this point, but I ended up overriding `dequeue` instead

stitesExpensify:
JK, my build is failing locally, but the actual test is failing here. I'll get that fixed now

stitesExpensify:
Okay we just needed more threads. The test is still not in the "slowest tests" list, so we're good to go for another review!

tylerkaraszewski:
@stitesExpensify - is this ready to review again?

stitesExpensify:
yep!

tylerkaraszewski:
The intent is to review this today. Sorry I haven't gotten to it yet, I have multiple several hundred-several thousand line reviews requested.

tylerkaraszewski (commented):
This comment was intended as personal notes.

BedrockBlockingCommandQueue.h
```diff
@@ -10,4 +10,37 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+    // Enforce per-identifier rate limits before enqueuing. Hides BedrockCommandQueue::push().
+    // Throws SException("503 ...") if the identifier is rate limited; caller should catch and reply.
+    //
+    // Rate limit state auto-resets when the queue has been continuously empty for 30 seconds — this
+    // is a global reset, not per-identifier. A brief drain (even one dequeue that empties the queue)
+    // starts the timer and will unblock all identifiers once 30 seconds elapse. This is intentional:
+    // the primary threat is a sustained burst from one identifier, so a quiet period is a safe signal
+    // to restore normal operation.
+    void push(unique_ptr<BedrockCommand>&& command);
```
tylerkaraszewski, line 22: this should be marked `override` and `BedrockCommandQueue.h`'s `push()` for this signature should be `virtual`.

`_dequeue` seems to have the correct model for this.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,114 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+    size_t cleared = _blockedIdentifiers.size();
+    _identifierCounts.clear();
+    _blockedIdentifiers.clear();
+    _emptyTime.store(0);
+    return cleared;
+}
+
+STable BedrockBlockingCommandQueue::getState()
+{
+    lock_guard<decltype(_queueMutex)> lock(_queueMutex);
```
tylerkaraszewski, line 104: We should run this check from `push()` here:
```
        if (emptyTime > 0 && STimeNow() - emptyTime >= 30'000'000) {
            _blockedIdentifiers.clear();
            _identifierCounts.clear();
        }
```

As is, if the queue empties and we don't push anything in it for a while, this will return out-of-date info.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,114 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+        }
+    }
+
+    // Track when the queue becomes empty for auto-clearing blocks.
+    // We iterate _queue directly rather than calling size() because size() acquires _queueMutex,
+    // which would deadlock here since _dequeue() is called by get() with that lock already held.
+    size_t queueSize = 0;
+    for (const auto& q : _queue) {
+        queueSize += q.second.size();
+    }
```
tylerkaraszewski, line 78: You can just call `_queue.empty()` in the `if` block instead (which requires the lock be held for safety, but as per the comment, that's already true).

BedrockServer.cpp
```diff
@@ -1084,7 +1088,14 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
         if (maxRetries || _shutdownState.load() != RUNNING) {
             if (command->processCount > maxRetries) {
                 SINFO("Max retries (" << maxRetries << ") hit in worker, sending '" << command->request.methodLine << "' to blocking queue with size " << _blockingCommandQueue.size());
-                _blockingCommandQueue.push(move(command));
+                try {
+                    _blockingCommandQueue.push(move(command));
+                } catch (const SException& e) {
```
tylerkaraszewski, line 1093: Add to the documentation for `push` that any exceptions are thrown *before* the call to `move()` happens and as such, your original unique_ptr is still valid on exception.

tylerkaraszewski (changes requested):

stitesExpensify:
Updated!

BedrockServer.cpp
```diff
@@ -1224,6 +1239,8 @@ void BedrockServer::_resetServer()
     _upgradeCompleted = false;
     _shutdownTime = chrono::time_point<chrono::steady_clock>{};
```
cead22:
```suggestion
```

BedrockServer.cpp
```diff
@@ -303,6 +303,9 @@ void BedrockServer::sync()
             for (auto& cmd : commands) {
                 _commandQueue.push(move(cmd));
             }
+            // Clear rate limit state since we're transitioning away from leading. Any blocks accumulated
```
cead22, line 313:
```suggestion

            // Clear rate limit state since we're transitioning away from leading. Any blocks accumulated
```

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,114 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+            _identifierCounts.clear();
+        }
+
+        if (_blockedIdentifiers.count(command->blockingIdentifier)) {
+            SINFO("Blocking queue rate limit: rejecting '" << command->request.methodLine
+                  << "' for identifier '" << command->blockingIdentifier << "'");
+            STHROW("503 Blocking queue rate limited");
+        }
+
+        size_t& count = _identifierCounts[command->blockingIdentifier];
```
cead22: Is this a reference so the next line updates `_identifierCounts[command->blockingIdentifier]` ?

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+            SData resetLimit("SetBlockingRateLimit");
+            resetLimit["MaxPerIdentifier"] = "0";
+            node.executeWaitVerifyContent(resetLimit, "200", true);
+        }
+    }
+
+    void testRateLimiting()
+    {
+        // Set MaxConflictRetries=1 on all nodes so commands enter the blocking queue after 1 conflict.
```
cead22: Do we have to send this command to all nodes? It's not enough to send it to leader? Same for SetBlockingRateLimit below

BedrockBlockingCommandQueue.cpp
stitesExpensify: yep!

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
stitesExpensify: Yes we do, it doesn't get replicated to the other servers

tylerkaraszewski: Why does it matter what it is on followers? I thought based on this comment we'd switch leader somewhere during this test but that doesn't seem to be the case, and the assertion at the end is:
```
// Verify the leader shows blocked users.
```

tylerkaraszewski (dismissed):
I haven't re-reviewed everything in detail since my last reviews. Tentatively approved though I agree with Carlos on one other comment remaining.

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
stitesExpensify: Oh duh. Yeah it doesn't matter because all of the writing is on the leader regardless. I'm thinking in production we'll want to do it on all of them though in case leader changes at any point

stitesExpensify: Updated this.

tylerkaraszewski (dismissed):

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,114 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+        uint64_t emptyTime = _emptyTime.load();
+        if (emptyTime > 0 && STimeNow() - emptyTime >= 30'000'000) {
+            _blockedIdentifiers.clear();
+            _identifierCounts.clear();
+        }
+
+        if (_blockedIdentifiers.count(command->blockingIdentifier)) {
+            SINFO("Blocking queue rate limit: rejecting '" << command->request.methodLine
+                  << "' for identifier '" << command->blockingIdentifier << "'");
+            STHROW("503 Blocking queue rate limited");
```
flodnv: Instead of deploying the block directly, can we start by deploying just logs and check whether it makes sense? I'm wary of blocking commands that we shouldn't be / needn't be blocking.

cead22:
Conflicts

BedrockBlockingCommandQueue.cpp
stitesExpensify: To clarify, are you saying we should start by not blocking at all, and just create an issue or something so that we can decide which commands to block and make sure we don't accidently break things?

flodnv: > To clarify, are you saying we should start by not blocking at all

Correct, as I feel that the potential for unintentional disruption to paying end users is quite high. 

> and just create an issue or something so that we can decide which commands to block and make sure we don't accidently break things?

We can just keep the `info` log, and use the existing issue.

stitesExpensify:
Ouch these are super bad conflicts too :( Working on it now

BedrockBlockingCommandQueue.cpp
stitesExpensify: I can, we will have to disable the tests though. I'm fine with doing it that way just to be safe!

stitesExpensify: Updated

stitesExpensify:
Okay this is ready for review again! The merge conflicts were due to a pretty big structure change, so things had to be moved around. I disabled the actual throw and left the log so that we can test before actually enabling

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,139 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+    }
+
+    // A command is entering the queue, so it is no longer empty. Clear the empty timestamp so
+    // the 30-second auto-reset window doesn't fire until the queue drains again.
+    _emptyTime.store(0);
+
+    try {
+        // Base class acquires its own (non-recursive) `_queueMutex`.
+        BedrockCommandQueue::push(move(command));
+    } catch (...) {
```
tylerkaraszewski, line 57: I think there's an edge case where we need to reset `_emptyTime` in this catch block as well.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,139 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+        // Roll back the increment so a base-class enqueue failure can't permanently inflate counts
+        // and falsely block this identifier.
+        if (checking) {
```
tylerkaraszewski: Could consider making a function out of the inside of this block since it's repeated below in _dequeue.

BedrockBlockingCommandQueue.h
```diff
@@ -10,4 +10,41 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+    // and records when the queue becomes empty.
+    unique_ptr<BedrockCommand> _dequeue() override;
+
+private:
+    // Guards rate limit state (`_identifierCounts`, `_blockedIdentifiers`). Separate from the base
+    // class `_queueMutex` because the base mutex is non-recursive and is held while `_dequeue` runs.
+    mutex _rateLimitMutex;
+
+    map<string, size_t> _identifierCounts;
+    set<string> _blockedIdentifiers;
```
tylerkaraszewski: I don't think we need this. It's redundant with `_identifierCounts`.

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,139 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+
+STable BedrockBlockingCommandQueue::getState()
+{
+    lock_guard<decltype(_rateLimitMutex)> lock(_rateLimitMutex);
+
+    uint64_t emptyTime = _emptyTime.load();
+    if (emptyTime > 0 && STimeNow() - emptyTime >= 30'000'000) {
+        _blockedIdentifiers.clear();
+        _identifierCounts.clear();
+    }
```
tylerkaraszewski: Copy `_identifierCounts` here, and then release the lock, and compute the rest of the response based on the copy.

tylerkaraszewski (changes requested):

stitesExpensify:
UPdated with suggestions, but tests are failing now (we just don't see it here because we disabled the tests since we're only logging)

BedrockCommand.h
```diff
@@ -281,6 +281,9 @@ class BedrockCommand : public SQLiteCommand {
     };
     CrashMap crashIdentifyingValues;
 
+    // An identifier for rate-limiting commands in the blocking queue. Empty means skip rate limiting.
+    string blockingIdentifier;
```
flodnv:
```suggestion
    string blockingQueueRateLimitIdentifier;
```

BedrockServer.cpp
```diff
@@ -2000,6 +2024,18 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
         }
     } else if (SIEquals(command->request.methodLine, "SetConflictPageLocks")) {
         _enableConflictPageLocks = command->request.test("enable");
+    } else if (SIEquals(command->request.methodLine, "SetBlockingRateLimit")) {
```
flodnv:
```suggestion
    } else if (SIEquals(command->request.methodLine, "SetBlockingQueueRateLimit")) {
```

BedrockServer.cpp
```diff
@@ -2000,6 +2024,18 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
...
+        if (command->request.isSet("MaxPerIdentifier")) {
```
flodnv:
```suggestion
        if (command->request.isSet("MaxRequestsPerIdentifier")) {
```

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,140 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+    BedrockCommandQueue(function<void(unique_ptr<BedrockCommand>&)>(startTiming),
+                        function<void(unique_ptr<BedrockCommand>&)>(stopTiming))
 {
 }
+
+void BedrockBlockingCommandQueue::push(unique_ptr<BedrockCommand>&& command)
+{
+    const string identifier = command->blockingIdentifier;
+    const size_t maxPerIdentifier = _maxPerIdentifier.load();
+    const bool checking = maxPerIdentifier > 0 && !identifier.empty();
```
flodnv:
```suggestion
    const bool shouldCheck = maxPerIdentifier > 0 && !identifier.empty();
```

BedrockBlockingCommandQueue.cpp
flodnv: Hm, this is still throwing? 😄

flodnv (commented):
Thanks for this, I'm eager to see it in action!

BedrockBlockingCommandQueue.cpp
stitesExpensify: heh bad timing. I just re-added the throw so that I could make sure it passed CI, and was about to remove it again right when you reviewed :)

stitesExpensify:
Okay we should be good to go here!

flodnv (dismissed):
All yours @tylerkaraszewski

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,141 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+            _decrementIdentifierCount(identifier);
+        }
+        throw;
+    }
+}
+
+unique_ptr<BedrockCommand> BedrockBlockingCommandQueue::_dequeue()
+{
+    // Called by `BedrockCommandQueue::get()` with the base `_queueMutex` held. Inspect `_queue` directly here;
+    // calling any base method that reacquires `_queueMutex` would deadlock.
```
cead22:
```suggestion

/**
 * Dequeues command and inspects _queue to update rate limit counts and _emptyTime
 * Called by `BedrockCommandQueue::get()` with the base `_queueMutex` held. Calling any base method that reacquires `_queueMutex` would deadlock.
 */
unique_ptr<BedrockCommand> BedrockBlockingCommandQueue::_dequeue()
{
```

cead22 (dismissed):

tylerkaraszewski (approved):

BedrockBlockingCommandQueue.cpp
```diff
@@ -11,6 +11,143 @@ void BedrockBlockingCommandQueue::stopTiming(unique_ptr<BedrockCommand>& command
...
+
+        if (count > maxPerIdentifier) {
+            SINFO("Blocking queue rate limit: rejecting '" << command->request.methodLine
+                  << "' for identifier '" << identifier << "'");
+            // TODO: enable enforcement after monitoring confirms thresholds are correct in production.
+            // STHROW("503 Blocking queue rate limited");
+        }
+
+        count++;
+        if (count > maxPerIdentifier) {
```
flodnv, line 44: Wait what, this `if (count > maxPerIdentifier) {` is the same as line 36 😅

stitesExpensify: lol oops, looks like copy/pasta to me, or maybe a relic of when there were 2 different identifiers. PR to de-duplicate here https://github.com/Expensify/Bedrock/pull/2615

stitesExpensify: I think I need another change there, please hold

stitesExpensify: Okay it's ready now, there was another error in the same spot

PR #2575, opened 2026-04-08
Use UTF-8 safe `char8_t` type when iterating over encode string input

s77rt:
### Details

When encoding a string, we iterate over each byte as `char` and `char` sign is platform dependent, on arm it's unsigned and on x64 it's signed. When this is signed, non-ASCII characters e.g. `ā` or `11000100 10000001` bytes in UTF-8 are seen as `-60` and `-127` and after [bit-shifting](https://github.com/Expensify/Bedrock/blob/7131d845bcedca41b4a5be7f35a4ed056b10af7e/libstuff/libstuff.cpp#L1465) we get `-4` and `-8`. These are used as indexes to access `hexChars` and since they are negative and out-of-bound this causes UB. 

This PR changes that we treat the bytes as `char8_t` which is guaranteed to be unsigned.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/617781

### Tests

- `LibStuff::testEncodeDecodeURIComponent`


s77rt:
@carlosmiceli Can you please review when you get a chance

carlosmiceli:
Looking now.

carlosmiceli:
@s77rt can you try merging main and see if that makes tests pass?

s77rt:
I tried I think there is nothing new since then. Can you rerun the workflow instead

carlosmiceli:
I did before, just did again, but still no luck :(

s77rt:
> Error: Unable to download artifact(s): Artifact not found for name: bedrock-binaries

Hmm can't be related to code.

carlosmiceli:
@MelvinBot could you diagnose what could be the issue  with Bedrock tests failing?

s77rt:
Merged main maybe that would help?

carlosmiceli (approved):

PR #2578, opened 2026-04-14
Add DB compression

tylerkaraszewski:
Implement release 1 of https://docs.google.com/document/d/1b05o9uNkH-_INm6ihBRHj_B7Khc9orvMa7aMMQJPRe8/edit?tab=t.0

Release 1 does not actually turn on any compression yet.

#### Notes
How journal snippets were selected:
```
sudo readdb.sh -j -csv "SELECT id, query FROM journal ORDER BY id DESC LIMIT 100" > recent_journal.csv
```

Sample data for zstd training created with:
```
#!/usr/bin/env python3
import csv
import os
import sys

csv.field_size_limit(sys.maxsize)

os.makedirs("sample_data", exist_ok=True)

with open("recent_journal.csv", "r", encoding="utf-8", errors="replace") as f:
    reader = csv.reader(f)
    next(reader)  # skip header
    for row in reader:
        row_id, query = row[0], row[1]
        with open(f"sample_data/{row_id}.txt", "w") as out:
            out.write(query)
```

Create the dictionary with:
```
find sample_data -name '*.txt' | xargs zstd --train -o journal.dict --dictID=1
```

Insert the dictionary into the DB on dev:
```
sqlite3 /var/auth/main.db <<'EOF'
CREATE TABLE IF NOT EXISTS zstdDictionaries (
    dictionaryID INTEGER PRIMARY KEY,
    description TEXT,
    dictionary BLOB
);
INSERT INTO zstdDictionaries (dictionaryID, description, dictionary)
    VALUES (1, 'journal dictionary', readfile('test/sample_data/journal.dict'));
EOF
```

Currently synchronization sends compressed data to nodes that are catching up, but live replication is not compressed.

### Fixed Issues
https://github.com/Expensify/Expensify/issues/601964

### Tests

Automated tests added.

Manually set this up on dev and verified it is compressing entries:
Screenshot 2026-04-20 at 14 28 09

Also confirmed it can be decompressed:
Screenshot 2026-04-20 at 14 34 21

_________


tylerkaraszewski:
@codex review

sqlitecluster/SQLiteNode.cpp
```diff
@@ -2164,7 +2165,7 @@ void SQLiteNode::_recvSynchronize(SQLitePeer* peer, const SData& message)
         if (!_db.beginTransaction()) {
             STHROW("failed to begin transaction");
         }
-        if (!_db.writeUnmodified(commit.content)) {
+        if (!_db.writeUnmodified(BedrockPlugin_Zstd::decompress(commit.content))) {
```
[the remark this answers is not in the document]
tylerkaraszewski: This wont happen in practice, these will be done offline.

.github/workflows/bedrock.yml
```diff
@@ -62,6 +62,8 @@ jobs:
     - name: Mark failure if debugging
       if: runner.debug == '1'
       run: exit 1
+    - name: Install zstd
+      run: apt-get update -y && apt-get install -y libzstd-dev
```
flodnv, line 66: I think this needs to be added to the VM?

plugins/Compression.cpp
```diff
@@ -0,0 +1,313 @@
...
+void BedrockPlugin_Compression::initializeFromDB(SQLite& db)
+{
+    loadDictionariesFromDB(db);
+}
+
+void BedrockPlugin_Compression::upgradeDatabase(SQLite& db)
+{
+    bool created;
+    SASSERT(db.verifyTable("zstdDictionaries",
+        "CREATE TABLE zstdDictionaries ("
```
flodnv, line 47: I think we need to create this table before deploying?

plugins/Compression.cpp
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        _dictionaries[id] = {cdict, ddict};
+        SINFO("Loaded zstd dictionary " << id << " (" << dictData.size() << " bytes)");
+    }
+
+    SINFO("Loaded " << _dictionaries.size() << " zstd dictionaries from DB.");
+}
+
+// SQLite UDF implementations
```
flodnv:
```suggestion
// SQLite User Defined Functions implementations
```

sqlitecluster/SQLite.cpp
```diff
@@ -262,7 +267,8 @@ SQLite::SQLite(const string& filename, int cacheSize, int maxJournalSize,
     _hctree(validateDBFormat(_filename, hctree)),
     _db(initializeDB(_filename, mmapSizeGB, _hctree)),
     _journalNames(initializeJournal(_db, minJournalTables)),
-    _sharedData(initializeSharedData(_db, _filename, _journalNames, _hctree)),
+    // Note that it's significant that _sharedData is initialized after _hctree, _db, and _journalNames. Re-ordereing these initialzations will break this.
```
flodnv: Can you update the comment to explain why?

plugins/Compression.cpp
aldo-expensify: `upgradeDatabase` is only called when the node starts leading, right?

aldo-expensify (dismissed):
besides this that may need to be fixed, looks good to me!

https://github.com/Expensify/Bedrock/pull/2578/changes#r3114222652

great work

.github/workflows/bedrock.yml
tylerkaraszewski: It's already in the VM for EBM

plugins/Compression.cpp
tylerkaraszewski: We should probably create it offline like normal, though I expect this would work just fine even if we didn't.

plugins/Compression.cpp
```diff
@@ -0,0 +1,313 @@
...
+{
+    SQResult result;
+    if (!db.read("SELECT dictionaryID, dictionary FROM zstdDictionaries;", result)) {
+        SWARN("Failed to read zstdDictionaries table.");
+        return;
+    }
+
+    for (size_t i = 0; i < result.size(); i++) {
+        size_t id = SToUInt64(result[i][0]);
+        const string dictData = result[i][1];
```
iwiznia:
```suggestion
        size_t id = SToUInt64(result[i]["dictionaryID"]);
        const string dictData = result[i]["dictionary"];
```

iwiznia (dismissed):
Let's add some instrumentation to know how long it takes to compress/decompress the data?

plugins/Compression.cpp
flodnv: This CREATE TABLE statement will get replicated and followers will just create the table via replication?

tylerkaraszewski:
Updated.

plugins/Compression.cpp
```diff
@@ -0,0 +1,313 @@
...
+    for (size_t i = 0; i < result.size(); i++) {
+        size_t id = SToUInt64(result[i]["dictionaryID"]);
+        const string dictData = result[i]["dictionary"];
+
+        if (dictData.empty()) {
+            SWARN("Empty dictionary data for dictionaryID " << id << ", skipping.");
+            continue;
+        }
+
+        int compressionLevel = 3;
```
cead22: Should we make this a class const?

plugins/Compression.cpp
```diff
@@ -0,0 +1,313 @@
...
+        ZSTD_CDict* cdict = ZSTD_createCDict(dictData.data(), dictData.size(), compressionLevel);
+        ZSTD_DDict* ddict = ZSTD_createDDict(dictData.data(), dictData.size());
+
+        if (!cdict || !ddict) {
+            SWARN("Failed to compile dictionary " << id);
```
cead22: Should we throw here (and where we warn above) instead of continuing? I think the node won't be able to work correctly if this fails

plugins/Compression.cpp
flodnv: Ping?

flodnv:
> Let's add some instrumentation to know how long it takes to compress/decompress the data?

That would be good, what do you think @tylerkaraszewski ?

test/clustertest/tests/CompressionTest.cpp
```diff
@@ -0,0 +1,298 @@
...
+    {
+        SData status("Status");
+        string response = node.executeWaitVerifyContent(status);
+        STable json = SParseJSONObject(response);
+        return SToUInt64(json["commitCount"]);
+    }
+
+    // Run a SELECT query against the journal tables via the bedrock server's DB plugin.
+    // All reads go through the server so that UDFs and dictionaries are available.
+    string queryJournal(BedrockTester& node, const string& selectExpr, uint64_t commitID)
```
cead22:
```suggestion
    string queryJournal(BedrockTester& node, const string& selectExpression, uint64_t commitID)
```

test/clustertest/tests/CompressionTest.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+        uint64_t commitAfter = getCommitCount(leader);
+        ASSERT_EQUAL(commitBefore + 1, commitAfter);
+
+        // Wait for both followers to replicate.
+        follower1.waitForStatusTerm("commitCount", to_string(commitAfter));
+        follower2.waitForStatusTerm("commitCount", to_string(commitAfter));
+
+        // Verify the journal entry is compressed on all three nodes.
+        string originalQuery = "INSERT INTO test VALUES(9999, 'Verifying test 3');";
```
cead22: Define this above and use it in `command["Query"] = ...`

cead22 (dismissed):

plugins/Compression.cpp
tylerkaraszewski: Yes, they will, the reason we don't normally do this is because we normally are creating new indexes rather than tables, and they can be huge and take a long time to create, and the slow timing of the replication breaks stuff.

iwiznia (dismissed):

flodnv (dismissed):

tylerkaraszewski:
@flodnv @iwiznia @cead22 - addressed remaining review items if you guys wouldn't mind re-approving.

flodnv (approved):

aldo-expensify (approved):

iwiznia (approved):

PR #2580, opened 2026-04-23
Check that followers actually have the new table

tylerkaraszewski:
### Details

When investigating this comment:
https://github.com/Expensify/Bedrock/pull/2578#discussion_r3118831602

I noticed that `UpgradeDBTest` verifies that the table exists with a write command. However, that will always run on leader, which doesn't actually mean the new table gets replicated to followers. This changes the test to write one row to leader, and then check that it exists on followers. If it succeeds, it means the table itself was replicated.

### Fixed Issues
Part of https://github.com/Expensify/Expensify/issues/601964

### Tests
Is test
_________


heekinho (approved):
LGTM!

tylerkaraszewski:
@heekinho style failed, please re-review.

heekinho (approved):

flodnv (approved):

PR #2589, opened 2026-05-04
Make PARAMS_WHITELIST thread-safe via atomic snapshot

tylerkaraszewski:
### Details

`PARAMS_WHITELIST` in `libstuff/SLog.cpp` was a bare `static set<string>` with no synchronization between `SWhitelistLogParams` (writer, currently called once during Auth plugin init) and the readers — `addLogParams` and `SIsLogParamWhitelisted` — which fire from every parameterized log call site. Today's startup ordering happens to make this safe by construction (writes complete before any worker thread starts, and `std::thread` creation establishes a happens-before edge), but the API doesn't enforce that, so any future runtime caller of `SWhitelistLogParams` would silently introduce a data race on a `std::set`.

This swaps the storage for `atomic<shared_ptr<const set<string>>>`. Readers do a single atomic load and dereference an immutable snapshot — no lock acquisition on the read path, which matters because `addLogParams` is one of the hottest reads in the binary (every `SINFO`/`SWARN`/etc. with params hits the whitelist N+1 times: once for syslog, once per key for the Fluentd path). The writer builds a new snapshot via copy-on-write and swaps it in with `compare_exchange_weak`, which keeps the function correct even under concurrent writers (the structural fragility the issue calls out).

I considered the `shared_timed_mutex` pattern that `BedrockServer::_blacklistedParallelCommands` uses, but that pattern fits incremental runtime mutation (its writers add commands via the control port) — whereas `PARAMS_WHITELIST` is read-mostly, write-once-at-startup, and the snapshot semantics are natural.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/602617

### Tests
No new tests added.
_________

### DIFF
--- a/libstuff/SLog.cpp
+++ b/libstuff/SLog.cpp
@@ -1,5 +1,6 @@
 #include "libstuff.h"
 #include <execinfo.h> // for backtrace*
+#include <memory>
 
 // Global logging state shared between all threads
 atomic<int> _g_SLogMask(LOG_INFO);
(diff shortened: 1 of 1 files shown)


libstuff/SLog.cpp
```diff
@@ -102,10 +108,17 @@ string addLogParams(string&& message, const STable& params)
 
 void SWhitelistLogParams(const set<string>& params)
 {
-    PARAMS_WHITELIST.insert(params.begin(), params.end());
+    auto current = PARAMS_WHITELIST.load();
+    while (true) {
+        auto next = make_shared<set<string>>(*current);
+        next->insert(params.begin(), params.end());
+        if (PARAMS_WHITELIST.compare_exchange_strong(current, next)) {
```
tylerkaraszewski, line 115: What this does, is if `PARAMS_WHITELIST` is still `current`, it replaces it with `next` and returns true, and if `PARAMS_WHITELIST` has changed from `current`, it does nothing and returns false.

This allows two threads to make updates simultaneously without writing over one another's changes by using the containing loop here, and never needs to lock.

libstuff/SLog.cpp
```diff
@@ -102,10 +108,17 @@ string addLogParams(string&& message, const STable& params)
 
 void SWhitelistLogParams(const set<string>& params)
 {
-    PARAMS_WHITELIST.insert(params.begin(), params.end());
+    auto current = PARAMS_WHITELIST.load();
+    while (true) {
+        auto next = make_shared<set<string>>(*current);
```
aldo-expensify, line 113: I think I understand the idea of this, but I have one question.. don't we need `auto current = PARAMS_WHITELIST.load();` inside the `while (true)` so it is updated if `PARAMS_WHITELIST.compare_exchange_strong(current, next)` returns `false`?

wouldn't the pointer have changed in such case and we need to get the new one?

tylerkaraszewski: Here is a shortened version of the docs for `compare_exchange_strong`:
> Atomically compares `*this` with that of `expected`. If those are bitwise-equal, replaces the former with `desired`. Otherwise, loads the actual value stored in `*this` into `expected`.

So `compare_exchange_strong` resets `current` on failure.

Good question though, it's not super obvious.

aldo-expensify: ohh, I missed that 👍

aldo-expensify (approved):

PR #2592, opened 2026-05-08
Fix race condition and mutex timeout in `_upgradeDB`

tylerkaraszewski:
### Details
This PR does a few things, some more related to the DB crash in the issue than others. I'll go through them in the order the files appear in the GH PR.

1. We reset `peer->subscribed` when we stop leading. This is maybe redundant but it's weird because otherwise this only ever gets reset if a peer disconnects or goes searching. That might cover every actual case but it's clearer to just do it when we stop leading.
2. Rename `committingCommand` to `commitInProgress`, as this is for any commit, not just commands (notably, DB upgrades).
3. There's a bunch of cleanup and refactoring around removing `_upgradeInProgress`. This is sort of redundant with `_upgradeCompleted` and so now we only use the latter. 
4. We simplified the check for when we can try and run an upgrade (we haven't already done it, we're leading, and we don't have a `commitInProgress`.
5. We wrap `_upgradeDB` in a try/catch to handle the actual crash we saw.
6. we replace `isUpgradeComplete` with `dbReadyToHandleRequests`. I'll talk about this for the rest of the PR here.

So, the old code has two different sorts of race conditions between where we do a DB upgrade and where we run commands.

We did this check:
```
if(LEADING) {
    if (hasQuorum()) {
        _upgradeInProgress = true;
```

So, first, let's talk about what it means to be be LEADING but not `hasQuorum()`:

LEADING really means "A quorum number nodes have told me that I can go LEADING".

And then `hasQuorum()` means "enough of those nodes that told me I could go LEADING are actually following".

There's a time period between these two where nodes that are attempting to follow are actually SUBSCRIBING, and this lasts at least one network round trip.

We do set _upgradeInProgress before that, but there's still a race condtion here where there's a window in which we are LEADING and `_upgradeInProgress` is false, but the upgrade hasn't happened.

So, the check in `runCommand` for whether we can start a command used to be:
```
while (_upgradeInProgress || (getState() != SQLiteNodeState::LEADING && getState() != SQLiteNodeState::FOLLOWING)) {
```

So you see there that it's possible to start a command thinking "upgrade is done" when we're leading but actually, the upgrade hasn't started yet. The logs for this crash show that happened for at least one command. It's possible some other command started in that window as well, and took over 5 seconds, and cause out lock to fail and resulting in what we saw:

```
2026-05-07T15:33:32.790976+00:00 db2.sjc bedrock: xxxxxx we@dont.know (SQLite.cpp:451) beginTransaction [sync] [warn] Failed to acquire commit lock in sync thread exclusive transaction!
2026-05-07T15:33:32.791032+00:00 db2.sjc bedrock: xxxxxx we@dont.know (libstuff.cpp:170) SException [sync] [info] Throwing exception with message: '512 Internal Lock Timeout' from sqlitecluster/SQLite.cpp:452
```

I don't actually have a concrete answer for what prevented us getting that lock in `_upgradeDB` for 5 seconds, but commands being able to start when they're supposed to be blocked seems like a good candidate.

So `dbReadyToHandleRequests` replaces this old check. This checks:

1. Are we FOLLOWING? We can serve requests.
2. Are we LEADING or STANDINGDOWN? We can serve requests if `_upgradeCompleted` is `true`, which is never `true` until an upgrade is true, so there's no race condition where you can check it before it's set and assume it's done.
3. Are we in some other state? We can't serve requests.

And then we augment this particularly in `runCommand` to exclude the STANDINGDOWN case, which just lets us shutdown more gracefully.

The main fix is the try/catch around `_upgradeDB` allowing a mutex that's blocked for 5 seconds to be retried in that case.

The secondary fix is to get rid of the race condition that allows commands to sneak in before `_upgradeDB` even runs, which hopefully (bot not provably) prevents the `Internal Lock Timeout` exception from ever happening.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/634788

### Tests
None
_________

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -141,7 +141,7 @@ void BedrockServer::sync()
     // Now we jump into our main command processing loop.
     uint64_t nextActivity = STimeNow();
     unique_ptr<BedrockCommand> command(nullptr);
-    bool committingCommand = false;
+    bool commitInProgress = false;
     bool upgradeInProgress = false;
 
     // Timer for S_poll performance logging. Created outside the loop because it's cumulative.
@@ -285,7 +285,7 @@ void BedrockServer::sync()
             _notifyDoneSync.postPoll(fdm);
         }
 
-        if (command && committingCommand) {
+        if (command && commitInProgress) {
             void (*onPrepareHandler)(SQLite& db, int64_t tableID) = nullptr;
             bool enabled = command->shouldEnableOnPrepareNotification(db, &onPrepareHandler);
             if (enabled) {
@@ -321,10 +321,10 @@ void BedrockServer::sync()
             // If we bailed out while doing a upgradeDB, clear state
             if (upgradeInProgress) {
                 upgradeInProgress = false;
-                if (committingCommand) {
+                if (commitInProgress) {
                     const string commandName = command ? command->getMethodName() : "No Command";
                     db.rollback(commandName);
-                    committingCommand = false; //NOLINT
+                    commitInProgress = false; //NOLINT
                 }
             }
 
@@ -362,20 +362,10 @@ void BedrockServer::sync()
             continue;
         }
 
-        // This first time we start LEADING, we will try and upgrade the DB.
-        // We then set _upgradeCompleted and will not attempt to upgrade again, there is no more work to do.
-        if (
-            // If we've aleady set that we've completed an upgrade, we can skip this block.
-            !_upgradeCompleted &&
-
-            // Run this block id we *JUST* switched
-            ((preUpdateState != SQLiteNodeState::LEADING && getState() == SQLiteNodeState::LEADING)
-             ||
-            (getState() == SQLiteNodeState::LEADING && upgradeInProgress && !committingCommand))
-        ) {
-            // Store this before we start writing to the DB, which can take a while depending on what changes were made
-            // (for instance, adding an index).
-            upgradeInProgress = true;
+        // If _upgradeCompleted is set, I.e., we're run at least one upgrade, we can skip this block.
+        // Otherwise, if we're leading, let's start the commit.
+        // Also, if a commit is in progress, skip it (it's the commit we started on the last loop iteration).
+        if (!_upgradeCompleted && getState() == SQLiteNodeState::LEADING && !commitInProgress) {
             if (!_syncNode->hasQuorum()) {
                 // Because we are going to commit the upgrade as a QUORUM commit (i.e., wait until over half of nodes
                 // approve it), we wait until `hasQuorum()` is true before we even attempt to start it.
@@ -392,7 +382,7 @@ void BedrockServer::sync()
                 continue;
             }
             if (_upgradeDB(db)) {
-                committingCommand = true;
+                commitInProgress = true;
                 _syncNode->startCommit(SQLiteNode::QUORUM);
                 _lastQuorumCommandTime = STimeNow();
 
@@ -405,7 +395,6 @@ void BedrockServer::sync()
             } else {
                 // If we're not doing an upgrade, we don't need to keep suppressing multi-write, and we're done with
                 // the upgradeInProgress flag.
-                upgradeInProgress = false;
                 _upgradeCompleted = true;
                 SINFO("UpgradeDB skipped, done.");
             }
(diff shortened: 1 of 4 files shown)


BedrockServer.cpp
```diff
@@ -364,27 +356,38 @@ void BedrockServer::sync()
...
-            if (_upgradeDB(db)) {
-                committingCommand = true;
+            bool dbHasChanges = false;
+            try {
+                dbHasChanges = _upgradeDB(db);
+            } catch (const SException& e) {
+                // It's possible that _upgradeDB throws at least "512 Internal Lock Timeout".
+                // Because upgrading the DB is critical and needs to be completed before we can use the
+                // DB for anything, we attempt a retry here. A lock timeout should be retryable.
+                SWARN("Got exception: " << e.method << " attempting to upgrade DB, will retry.");
```
flodnv: Can we log using log params, and log `e.what` as well?

BedrockServer.cpp
```diff
@@ -425,50 +428,46 @@ void BedrockServer::sync()
...
             if (command->shouldPostProcess() && command->response.methodLine == "200 OK") {
                 // PostProcess if the command should run postProcess, and there have been no errors thrown thus far.
                 core.postProcessCommand(command, false);
             }
 
             if (_syncNode->commitSucceeded()) {
-                if (command) {
-                    SINFO("[performance] Sync thread finished committing command " << command->request.methodLine);
-                    _conflictManager.recordTables(command->request.methodLine, db.getTablesUsed());
+                SINFO("[performance] Sync thread finished committing command " << command->request.methodLine);
```
flodnv: NAB: I think very few commands run on the sync thread, so could we remove `[performance]` from this log line so that they get into VL?

BedrockServer.h
```diff
@@ -215,8 +215,15 @@ class BedrockServer : public SQLiteServer {
     // Returns if we are detached and the sync thread has exited.
     bool isDetached();
 
-    // Returns if all plugins have completed calling UpgradeDB
-    bool isUpgradeComplete();
+    // If a node is LEADING (or STANDINGDOWN) and has, at some point, completed a DB upgrade,
+    // OR if the node is FOLLOWING, it is ready to handle requests.
+    bool dbReadyToHandleRequests();
+
+    // Legacy name for dbReadyToHandleRequests, retained for backwards compatibility.
```
flodnv, line 222: backwards compatiblity with...?

BedrockServer.cpp
```diff
@@ -1691,9 +1692,22 @@ bool BedrockServer::isDetached()
     return _detach && !_syncLoopShouldBeRunning && _pluginsDetached;
 }
 
-bool BedrockServer::isUpgradeComplete()
+bool BedrockServer::dbReadyToHandleRequests()
 {
-    return _upgradeCompleted;
+    auto state = getState();
+    if (state == SQLiteNodeState::FOLLOWING) {
+        // We are as upgraded as we can be, because someone else is controlling what the state of "upgraded" is,
```
flodnv: This comment is sort of strange to read in isolation of the rest of the changes and explanations in this PR. It raises the question "what does 'being upgraded' have to do with dbReadyToHandleRequests?"

flodnv (commented):
Thanks!

BedrockServer.h
tylerkaraszewski: Auth

tylerkaraszewski:
@flodnv - updated.

cead22 (dismissed):

flodnv (dismissed):

BedrockServer.cpp
```diff
@@ -1691,9 +1692,22 @@ bool BedrockServer::isDetached()
     return _detach && !_syncLoopShouldBeRunning && _pluginsDetached;
 }
 
-bool BedrockServer::isUpgradeComplete()
+bool BedrockServer::dbReadyToHandleRequests()
 {
-    return _upgradeCompleted;
+    auto state = getState();
+    if (state == SQLiteNodeState::FOLLOWING) {
+        // If we are following someone else is in charge of upgrades, we can serve reqeusts.
```
aldo-expensify:
```suggestion
        // If we are following someone else is in charge of upgrades, we can serve requests.
```

NAB

aldo-expensify (dismissed):

BedrockServer.cpp
tylerkaraszewski: I'm gonna merge regardless so we don't need to get everyone else to re-approve.

aldo-expensify:
The HCTree tests continue to fail, could the failure be related to the changes here?

BedrockServer.cpp
tylerkaraszewski: I had to update a test so I'll do this as well.

tylerkaraszewski:
@aldo-expensify @cead22 @flodnv Updated, I needed to adjust a test to not flake out here.

flodnv (approved):

PR #2599, opened 2026-05-15
Enable sending compressed journal entries over the network.

tylerkaraszewski:
### Details
HOLD for https://github.com/Expensify/Bedrock/pull/2609

When compressing journal entries on leader, we currently send the original plaintext to followers. This allows us to send compressed queries to followers, saving network bandwidth.

### Fixed Issues
https://github.com/Expensify/Expensify/issues/601964

### Tests
Existing compression tests handle this case.
_________

### DIFF
--- a/plugins/Compression.cpp
+++ b/plugins/Compression.cpp
@@ -260,6 +260,44 @@ void BedrockPlugin_Compression::registerSQLite(sqlite3* db)
                                nullptr, ::sqliteDecompress, nullptr, nullptr, nullptr);
 }
 
+string BedrockPlugin_Compression::compress(const string& input, size_t dictID)
+{
+    if (dictID == 0 || input.empty()) {
+        return input;
+    }
+
+    ZSTD_CDict* cdict = getCompressionDictionary(dictID);
+    if (!cdict) {
+        SWARN("compress(): no dictionary found for ID " << dictID);
+        return input;
+    }
+
+    ZSTD_CCtx* cctx = ZSTD_createCCtx();
+    if (!cctx) {
+        SWARN("compress(): failed to create compression context");
+        return input;
+    }
+
+    // Match the SQL UDF's frame parameters so on-disk and on-wire bytes are interchangeable.
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_dictIDFlag, 1);
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_checksumFlag, 0);
+
+    size_t dstCap = ZSTD_compressBound(input.size());
+    string output(dstCap, '\0');
+
+    size_t compressedSize = ZSTD_compress_usingCDict(cctx, output.data(), dstCap,
+                                                     input.data(), input.size(), cdict);
+    ZSTD_freeCCtx(cctx);
+
+    if (ZSTD_isError(compressedSize)) {
+        SWARN("compress(): " << ZSTD_getErrorName(compressedSize));
+        return input;
+    }
+
+    output.resize(compressedSize);
+    return output;
+}
+
 string BedrockPlugin_Compression::decompress(const string& input)
 {
     if (input.empty()) {
--- a/plugins/Compression.cpp
+++ b/plugins/Compression.cpp
@@ -268,14 +268,12 @@ string BedrockPlugin_Compression::compress(const string& input, size_t dictID)
 
     ZSTD_CDict* cdict = getCompressionDictionary(dictID);
     if (!cdict) {
-        SWARN("compress(): no dictionary found for ID " << dictID);
-        return input;
+        SERROR("compress(): no dictionary found for ID " << dictID << " — configured dictionary is missing from the database");
     }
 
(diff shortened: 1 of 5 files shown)


tylerkaraszewski:
@codex review

plugins/Compression.cpp
```diff
@@ -260,6 +260,41 @@ void BedrockPlugin_Compression::registerSQLite(sqlite3* db)
...
+string BedrockPlugin_Compression::compress(const string& input, size_t dictID)
+{
+    if (dictID == 0 || input.empty()) {
+        return input;
+    }
+
+    ZSTD_CDict* cdict = getCompressionDictionary(dictID);
+    if (!cdict) {
+        SERROR("compress(): no dictionary found for ID " << dictID << " — configured dictionary is missing from the database");
+    }
```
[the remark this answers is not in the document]
tylerkaraszewski: If we turn on compression and set a dictionary, but that dictionary is not in the DB, this is a configuration error and the error is reasonable.

sqlitecluster/SQLiteNode.cpp
```diff
@@ -2338,7 +2338,8 @@ void SQLiteNode::_handleBeginTransaction(SQLite& db, SQLitePeer* peer, const SDa
     }
 
     // Inside transaction; get ready to back out on error
-    if (!db.writeUnmodified(message.content)) {
+    //  decompress() is a no-op for non-zstd-framed input, so this works whether or not the leader is compressing.
+    if (!db.writeUnmodified(BedrockPlugin_Compression::decompress(message.content))) {
```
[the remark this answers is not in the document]
tylerkaraszewski: If the transaction payload is corrupted in transit nothing has changed, corrupt SQL will crash, corrupt zstd frames will crash.

tylerkaraszewski:
This is off hold, please review.

plugins/Compression.cpp
```diff
@@ -260,6 +260,41 @@ void BedrockPlugin_Compression::registerSQLite(sqlite3* db)
...
 
+string BedrockPlugin_Compression::compress(const string& input, size_t dictID)
+{
+    if (dictID == 0 || input.empty()) {
+        return input;
+    }
+
+    ZSTD_CDict* cdict = getCompressionDictionary(dictID);
+    if (!cdict) {
+        SERROR("compress(): no dictionary found for ID " << dictID << " — configured dictionary is missing from the database");
```
flodnv, line 271: I feel like these 3 calls to `SERROR` should `STHROW` instead of crash... no?

plugins/Compression.h
```diff
@@ -30,6 +30,10 @@ class BedrockPlugin_Compression : public BedrockPlugin {
     // Register the compress(data, dictID) and decompress(data) SQLite UDFs.
     static void registerSQLite(sqlite3* db);
 
+    // Returns input compressed with the given dictionary ID. If dictID is 0 or input is empty,
+    // returns input unchanged. Produces a byte-identical zstd frame to the compress() SQL UDF.
```
flodnv, line 34:
```suggestion
    // returns input unchanged. Produces a byte-identical zstd frame to the compress() SQLite UDF.
```

sqlitecluster/SQLite.cpp
```diff
@@ -895,22 +895,21 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash)
         *transactionhash = _uncommittedHash;
     }
 
-    // Wrap the bound query value with compress() for zstd compression. When journalZstdDictionaryID is 0
-    // (the default), compress() returns data unchanged.
-    string query = "INSERT INTO " + _journalName + " VALUES (:commitID, compress(:query, :dictID), :hash)";
+    if (!_uncommittedQuery.empty()) {
+        _uncommittedQuery = BedrockPlugin_Compression::compress(_uncommittedQuery, journalZstdDictionaryID.load());
+    } else {
+        SINFO("Will commmit blank query");
```
flodnv, line 901: Update the log or add a comment or something to explain why we're doing this?

fukawi2:
Removing myself as I cannot provide a meaningful review.

plugins/Compression.cpp
coleaeason: I would expect it to crash, the server is unable to function in this state, it can't compress data to send or store

sqlitecluster/SQLite.cpp
coleaeason: Should this be a hmm?

coleaeason (approved):

plugins/Compression.cpp
tylerkaraszewski: Yeah, my take is the same as @coleaeason - if you forgot to add the dictionary you should crash and then go add the dictionary to the DB.

sqlitecluster/SQLite.cpp
tylerkaraszewski: It's been there forever and never really mattered so I'm fine leaving it.

PR #2600, opened 2026-05-16
Add bound parameters to SQuery

tylerkaraszewski:
### Details

This is a cool change. It will let you do this:

```
db.write(
    "INSERT INTO reportActions (created, reportID, accountID, action, message) "
    "VALUES (:created, :reportID, :accountID, :action, :message);",
    {
        {":created",   SQLite::Parameter::text(SCURRENT_TIMESTAMP_MS())},
        {":reportID",  SQLite::Parameter::i(reportID)},
        {":accountID", SQLite::Parameter::i(requestingAccountID)},
        {":action",    SQLite::Parameter::text(ReportAction::ACTION_BILLABLE_DELEGATE)},
        {":message",   SQLite::Parameter::text(delegateActionMessage.serialize())},
    }
);
```

Which every other language that interacts with DBs has had for forever, but we've never had in bedrock. There was a difficulty in doing this around replicating queries, but sqlite provides the API we need to replicate these queries after the replacement happens, so we do that.

This also means that we don't need `SQ()` for queries in this format. 

I didn't do this just because it would be cool to add. I was implementing sending journal entries already compressed to followers, which requires dong the compression first, and then inserting the compressed (and thus binary) data into the journal, and delivering it to followers over the network. To be clear, we can compress the data by doing something like `INSERT INTO journal VALUES(compress(query));` and the journal *stores* the data compressed but then we don't have access to the compressed data to send to followers. Instead, we will do something like:

```
compressedData = compress(query);
db.write("INSERT INTO journal VALUES(:compressedData);", {{":compressedData", compressedData}});
follower.send(compressedData);
```

This avoids us having to do the compression twice, a second query, or an ascii-encoding of the compressed data. Note that sending compressed journal queries to followers will be a separate PR that depends on this one, this one just does named query parameters.

The PR is fairly large but the most significant and least obvious part is the refactor of `SQuery`. Half the change is tests and test library updates, and the new `SQliteParameter` files are fairly straightforward.

We removed a deprecated API in this PR, thus we need to wait for the following to merge before we can merge this:
~Hold for: https://github.com/Expensify/FuzzyBot/pull/350~
~Hold for: https://github.com/Expensify/ExpensifyBackupManager/pull/157~

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/601964

### Tests
New tests added.

Auth, WAL2:
```
[ TEST RESULTS ] Passed: 8295, Failed: 0
```

Auth, HC-Tree:
```
[ TEST RESULTS ] Passed: 8295, Failed: 0
```

_________

### DIFF
--- a/libstuff/SQResult.cpp
+++ b/libstuff/SQResult.cpp
@@ -302,3 +302,8 @@ void SQResult::emplace_back(SQResultRow&& row)
 {
     rows.emplace_back(move(row));
 }
+
+void SQResult::resize(size_t newSize)
+{
+    rows.resize(newSize);
+}
--- a/libstuff/SQResult.cpp
+++ b/libstuff/SQResult.cpp
@@ -303,7 +303,7 @@ void SQResult::emplace_back(SQResultRow&& row)
     rows.emplace_back(move(row));
 }
 
-void SQResult::resize(size_t newSize)
+void SQResult::resize(const size_t newSize)
 {
     rows.resize(newSize);
 }
--- a/libstuff/SQResult.h
+++ b/libstuff/SQResult.h
@@ -54,7 +54,7 @@ public:
     // Mutators
     void clear();
     void emplace_back(SQResultRow&& row);
-    void resize(size_t newSize);
+    void resize(const size_t newSize);
 
     // Operators
     const SQResultRow& operator[](size_t rowNum) const;
--- a/libstuff/SQResult.h
+++ b/libstuff/SQResult.h
@@ -54,6 +54,7 @@ public:
     // Mutators
     void clear();
     void emplace_back(SQResultRow&& row);
+    void resize(size_t newSize);
 
     // Operators
     const SQResultRow& operator[](size_t rowNum) const;
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2816,6 +2816,12 @@ int SQuery(sqlite3* db, const string& sql, SQResult& result, int64_t warnThresho
     size_t stepTimeUS = 0;
     size_t longestStepTimeUS = 0;
 
+    // When bound parameters are supplied, accumulate the SQL with placeholders substituted by
+    // their actual values so the slow/failed query logs below show real data instead of
+    // ":name" placeholders. Stays empty when no params were supplied, in which case the logs
+    // fall back to the raw template (no values to inline).
+    string boundExpandedSql;
+
     SDEBUG(sql.substr(0, MAX_LOG_QUERY_SIZE));
 
     // Retry the whole query (prepare + bind + step) on SQLITE_BUSY. Prepare and step both can return
@@ -2827,6 +2833,7 @@ int SQuery(sqlite3* db, const string& sql, SQResult& result, int64_t warnThresho
         if (expandedSql) {
             expandedSql->clear();
         }
+        boundExpandedSql.clear();
         error = SQLITE_OK;
         const char* statementRemainder = sql.c_str();
         while (*statementRemainder != 0 && error == SQLITE_OK) {
@@ -2991,15 +2998,24 @@ int SQuery(sqlite3* db, const string& sql, SQResult& result, int64_t warnThresho
                 }
             }
 
-            if (error == SQLITE_OK && expandedSql) {
-                char* expanded = sqlite3_expanded_sql(preparedStatement);
-                if (expanded) {
+            // Expand all query parameters so that we have complete queries for replication and logging.
+            char* expanded = sqlite3_expanded_sql(preparedStatement);
+            if (!expanded) {
+                // Error out if expansion failed.
+                SWARN("sqlite3_expanded_sql failed for: " << sql.substr(0, MAX_LOG_QUERY_SIZE));
+                error = SQLITE_NOMEM;
+            } else {
+                // Expansion suceeded.
+                // If the query succeeded and we were passed a string to store the expanded version in, do that now.
+                if (expandedSql && error == SQLITE_OK) {
                     *expandedSql += expanded;
-                    sqlite3_free(expanded);
-                } else {
-                    SWARN("sqlite3_expanded_sql returned NULL for: " << sql.substr(0, MAX_LOG_QUERY_SIZE));
-                    error = SQLITE_NOMEM;
                 }
+
+                // We store boundExpandedSql regardless, as we will use it for logging even if queries failed.
+                boundExpandedSql += expanded;
+
+                // Done.
+                sqlite3_free(expanded);
             }
 
             sqlite3_finalize(preparedStatement);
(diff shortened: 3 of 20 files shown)


tylerkaraszewski:
@codex review

libstuff/libstuff.cpp
```diff
@@ -2816,70 +2816,123 @@ int SQuery(sqlite3* db, const string& sql, SQResult& result, int64_t warnThresho
...
-        const char* statementRemainder = sql.c_str();
-        do
-        {
-            numLoops++;
-            sqlite3_stmt* preparedStatement = nullptr;
-            size_t beforePrepare = 0;
-            if (isSyncThread) {
-                beforePrepare = STimeNow();
-            }
+            // This will just drop through to the general error handling below.
```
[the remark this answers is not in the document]
tylerkaraszewski: This shouldn't be able to happen, and even if it does, failing on the first instead of third try in this case wouldn't really matter.

libstuff/libstuff.cpp
```diff
@@ -2858,7 +2865,57 @@ int SQuery(sqlite3* db, const string& sql, SQResult& result, int64_t warnThresho
             } else if (!preparedStatement) {
                 // If we get a null statement (from parsing a blank string) we can skip, this isn't an error.
                 error = SQLITE_OK;
-                break;
+                continue;
+            }
+
+            // Bind parameters once per prepared statement. The bind state survives sqlite3_reset, so we never
+            // need to rebind on a SQLITE_BUSY retry. Map keys must include the prefix character (`:`, `@`, or
```
iwiznia, line 2879: Why do we have 3 prefixes?

iwiznia: Oh I think this follows what sqlite does?

iwiznia (dismissed):
Nice, love this!

Will this break our logging of slow/failed queries by not logging the actual values?

Should we require all new code to use bound params via AI review? They seem better in almost every way, no?

test/tests/LibStuffTest.cpp
```diff
@@ -968,6 +969,63 @@ struct LibStuff : tpunit::TestFixture
         ASSERT_EQUAL(result[0]["value"], "value1");
     }
 
+    void testSQueryBoundParameters()
```
coleaeason, line 973: I realize this would be testing the downstream sqlite behavior but if we are going to replace SQ() with this syntax it would be nice to have a dedicated injection test to make sure it prevents injections in the expected ways

coleaeason (changes requested):
I hope you find more cool things to add while working on unrelated items :D, i've been wanting this for a long time!

libstuff/libstuff.cpp
tylerkaraszewski: Yeah, because sqlite supports all of them.

tylerkaraszewski:
> Will this break our logging of slow/failed queries by not logging the actual values?

It doesn't need to, I may need to adjust a logline for that. I will look.

libstuff/libstuff.cpp
```diff
@@ -2816,13 +2816,20 @@ int SQuery(sqlite3* db, const string& sql, SQResult& result, int64_t warnThresho
     size_t stepTimeUS = 0;
     size_t longestStepTimeUS = 0;
 
+    SDEBUG(sql.substr(0, 20000));
```
heekinho:
```suggestion
    SDEBUG(sql.substr(0, MAX_LOG_QUERY_SIZE));
```

heekinho:
Wow! I love this too! This is a nice improvement!

tylerkaraszewski:
All review feedback thus far addressed, including adding one more test for you @coleaeason .

coleaeason (dismissed):

tylerkaraszewski:
@coleaeason - sorry, changed one more thing (and style).

libstuff/SQResult.h
```diff
@@ -54,6 +54,7 @@ class SQResult {
     // Mutators
     void clear();
     void emplace_back(SQResultRow&& row);
+    void resize(size_t newSize);
```
aldo-expensify:
```suggestion
    void resize(const size_t newSize);
```

NAB

libstuff/SQResult.cpp
```diff
@@ -302,3 +302,8 @@ void SQResult::emplace_back(SQResultRow&& row)
 {
     rows.emplace_back(move(row));
 }
+
+void SQResult::resize(size_t newSize)
```
aldo-expensify:
```suggestion
void SQResult::resize(const size_t newSize)
```

NAB

aldo-expensify (dismissed):
Looks great!!

> Should we require all new code to use bound params via AI review? They seem better in almost every way, no?

+1

<hr>

@tylerkaraszewski is it possible to repeat the same parameter name in a query? for example `:accountID` in:

https://github.com/Expensify/Auth/blob/07640d98cac5c377fd6656b806ab00ccb0bb827a/auth/lib/Report.cpp#L15541-L15562


Looking at the code https://github.com/Expensify/Bedrock/pull/2600/changes#diff-d0f3e9340d9d47b780f0bf98f9eb8e1edb6f041652946c35fcae859e2ea82ebcR2885-R2919 makes me think that this wouldn't be supported because there would be multiple indexes for the same name, no?

tylerkaraszewski:
> is it possible to repeat the same parameter name in a query? for example :accountID in: ...

This should work fine, so long as you want both instances of it to have the same value. When doing the substitution into the query, it can substitute multiple times.

tylerkaraszewski:
Applied your suggestion and addressed your question @aldo-expensify

test/tests/LibStuffTest.cpp
```diff
@@ -968,6 +970,133 @@ struct LibStuff : tpunit::TestFixture
...
+
+        // The single-value read overload also takes bound parameters.
+        ASSERT_EQUAL("hello", db.read("SELECT t FROM params WHERE i = :i;",
+                                      {{":i", SQLite::Parameter::i(42)}}));
+
+        // Empty params with a parameter-free statement still works (existing zero-arg behavior unchanged).
+        ASSERT_TRUE(db.read("SELECT COUNT(*) FROM params;", result));
+        ASSERT_EQUAL("1", result[0][0]);
+
+        // Supplying a parameter name that isn't referenced by the SQL is allowed — extra entries are silently
```
aldo-expensify, line 1026: NAB: Should we throw instead to make it more visible if a developer is making this by mistake?

heekinho (approved):

test/tests/LibStuffTest.cpp
tylerkaraszewski: I don't know if one is better than another.

test/tests/LibStuffTest.cpp
```diff
@@ -968,6 +970,133 @@ struct LibStuff : tpunit::TestFixture
         ASSERT_EQUAL(result[0]["value"], "value1");
     }
 
+    void testSQueryBoundParameters()
+    {
```
aldo-expensify, line 974: What happens when we don't pass a value for a named param? I didn't find a test case for that

test/tests/LibStuffTest.cpp
tylerkaraszewski: Oh, there was a reason (not a great one) to allow this:

You can concatenate queries:
```
"SELECT 1; SELECT :accountID FROM accounts;"
```

The check is per query, so if you supplied `:accountID` the first query would throw an error if all are required.

Not sure that's an *important* case, but it's there.

test/tests/LibStuffTest.cpp
tylerkaraszewski: The query will fail with something like `no column named :accountID`

iwiznia (approved):

heekinho (approved):

tylerkaraszewski:
@coleaeason - care to final approve or should I merge?

aldo-expensify (approved):

iwiznia:
> Should we require all new code to use bound params via AI review? They seem better in almost every way, no?

Thoughts?

PR #2602, opened 2026-05-19
Reply to timed out commands without requeuing them

tylerkaraszewski:
### Details
Fixes the bug we saw where a command that was scheduled in the future with a timeout in the past would get into an infinite loop of being pulled off the command queue and re-queued immediately because it wasn't scheduled to run yet. This change makes these time out at the timeout* point, regardless of the scheduled run time.

### Fixed Issues
Fixes https://expensify.enterprise.slack.com/archives/C0B4AUG3FU7

### Tests

_________

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -751,8 +751,8 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
     // This functionality will go away as we remove the queues from bedrock, and so this can be removed at that time.
     {
         auto _syncNodeCopy = atomic_load(&_syncNode);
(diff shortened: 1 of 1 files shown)


danieldoglas (dismissed):
LGTM

NikkiWines (dismissed):

rafecolton (dismissed):

rafecolton (dismissed):

danieldoglas (dismissed):

danieldoglas (approved):

madmax330 (approved):

aldo-expensify (approved):

flodnv (approved):

PR #2610, opened 2026-05-25
Add try_lock_for on commit lock to respect the total command timeout

danieldoglas:
### Details

We saw a behavior where we were holding transactions for a long time, longer than our timeout for any command.

Considering a timeout of 110s, this is what was happening:

- Command executed peek/process in 2s. But waited for the commit lock for 45s, and it conflicted, so we ran it again.
- Command executed again in 2s. Waited for the commit lock for 45s, and conflicted again, so it went to the third try
- Command finished in 2s, but waited for the commit lock for 130s. in this case, it should have timed out after 14 seconds, but it didn't. It conflicted again.
- Before sending it to the blocking queue, we checked if it was timed out, and returned an error to the caller.

That means that we held that command transaction open for over 110s without need. With enough parallel transactions, this will increase WAL file size and make all other db reads/writes slow.

This PR changes the commit lock to wait for a maximum of the remaining command timeout.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/640539

### Tests

I had a script that ran 60 parallels command locally with a high timeout, and one "victim" command with a short timeout (50ms). Ran the script several times, and confrmed that sometimes the victim timed out early:

```
=== Victim response ===
555 Timeout
commitCount: 425318
Connection: close
nodeName: bedrock
peekTime: 52
processTime: 396
totalTime: 51473
unaccountedTime: 5020
Content-Length: 0


=== Victim status line ===
555 Timeout

=== Relevant log lines written after the victim was sent ===
2026-05-27T12:29:42.110203+00:00 expensidev-2404 bedrock: lHQMD1 we@dont.know (SQLite.cpp:868) prepare [socket32089_cmd] [info] Timed out after 23625us waiting for commit lock.
```
It's important to note that since the order in try_lock_for is not FIFO, this script might return success sometimes. 

<details><summary>Script</summary>
<p>

This script failed to shutdown it's own processes in the second run, so do `pkill -f "script_name.sh"

```bash
set -uo pipefail

HOST="${HOST:-127.0.0.1}"
PORT="${PORT:-8888}"
WORKERS="${WORKERS:-50}"
WARMUP_SECS="${WARMUP_SECS:-2}"
VICTIM_TIMEOUT_MS="${VICTIM_TIMEOUT_MS:-50}"
WORKER_TIMEOUT_MS="${WORKER_TIMEOUT_MS:-60000}"
VALUE_BYTES="${VALUE_BYTES:-262144}"
LOG_FILE="${LOG_FILE:-/var/log/syslog}"

cleanup() {
    [[ ${#WORKER_PIDS[@]:-0} -gt 0 ]] && kill "${WORKER_PIDS[@]}" 2>/dev/null
    pkill -P $$ 2>/dev/null
    wait 2>/dev/null
}
trap cleanup EXIT INT TERM
WORKER_PIDS=()

# Detect nc flavor (GNU has -q, BSD/macOS does not).
if nc -h 2>&1 | grep -q -- " -q "; then
    NC=(nc -q 1 -w 5)
else
    NC=(nc -w 5)
fi

# A single, in-memory value reused by all workers and the victim.
VALUE="$(head -c "$VALUE_BYTES" /dev/urandom | base64 | head -c "$VALUE_BYTES")"

send_writecache() {
    local name="$1" timeout_ms="$2"
    printf 'WriteCache\r\nname: %s\r\nvalue: %s\r\nTimeout: %s\r\nConnection: close\r\n\r\n' \
        "$name" "$VALUE" "$timeout_ms" | "${NC[@]}" "$HOST" "$PORT"
}

worker_loop() {
    local id="$1" i=0
    while :; do
        send_writecache "worker_${id}_$((i++))" "$WORKER_TIMEOUT_MS" >/dev/null 2>&1
    done
}

# 1. Confirm the target is leading. The new code path only runs on the leader
#    (followers escalate writes to the sync thread).
STATUS="$(printf 'Status\r\nConnection: close\r\n\r\n' | "${NC[@]}" "$HOST" "$PORT")"
if ! grep -q '"state"[[:space:]]*:[[:space:]]*"LEADING"' <<<"$STATUS"; then
    echo "ERROR: $HOST:$PORT is not LEADING; aborting." >&2
    echo "First 40 lines of Status response:" >&2
    head -40 <<<"$STATUS" >&2
    exit 1
fi

# 2. Spawn contention.
echo "Spawning $WORKERS WriteCache workers against $HOST:$PORT (value=${VALUE_BYTES}B)..."
for i in $(seq 1 "$WORKERS"); do
    worker_loop "$i" &
    WORKER_PIDS+=($!)
done

# 3. Warm up so the commit lock is consistently held by a worker.
sleep "$WARMUP_SECS"

# 4. Mark log position (if accessible) so we can isolate the victim's lines.
LOG_START_BYTES=0
if [[ -r "$LOG_FILE" ]]; then
    LOG_START_BYTES=$(wc -c <"$LOG_FILE")
fi

# 5. Fire the victim.
VICTIM_NAME="victim_$(date +%s%N)"
echo "Sending victim: name=$VICTIM_NAME Timeout=${VICTIM_TIMEOUT_MS}ms"
VICTIM_RESPONSE="$(send_writecache "$VICTIM_NAME" "$VICTIM_TIMEOUT_MS" || true)"

# 6. Let a few more commits flush through the log, then stop workers.
sleep "$WARMUP_SECS"
cleanup
trap - EXIT INT TERM

# 7. Report.
echo
echo "=== Victim response ==="
echo "$VICTIM_RESPONSE"

VICTIM_STATUS="$(printf '%s' "$VICTIM_RESPONSE" | head -1)"
echo
echo "=== Victim status line ==="
echo "$VICTIM_STATUS"

if [[ -r "$LOG_FILE" ]]; then
    echo
    echo "=== Relevant log lines written after the victim was sent ==="
    tail -c +"$((LOG_START_BYTES + 1))" "$LOG_FILE" | \
        grep -E "Timed out after .* waiting for commit lock|Command 'WriteCache' timed out before commit" || \
        echo "(no matching log lines found)"
else
    echo
    echo "(LOG_FILE=$LOG_FILE not readable; export LOG_FILE=... or read via 'docker logs ...')"
fi

# 8. Pass/fail. A `555` victim status is the contract for path 4.
echo
if [[ "$VICTIM_STATUS" == *"555"* ]]; then
    echo "PASS: victim returned 555 Timeout."
    exit 0
else
    echo "FAIL: victim did not return 555."
    exit 1
fi
```

</p>
</details> 

_________

### DIFF
--- a/BedrockCore.cpp
+++ b/BedrockCore.cpp
@@ -32,7 +32,7 @@ private:
     bool (*_handler)(int, const char*, string&);
 };
 
-uint64_t BedrockCore::getRemainingTime(const unique_ptr<BedrockCommand>& command, bool isProcessing)
+chrono::microseconds BedrockCore::getRemainingTime(const unique_ptr<BedrockCommand>& command, bool isProcessing)
 {
     int64_t timeout = command->timeout();
     int64_t now = STimeNow();
@@ -53,7 +53,7 @@ uint64_t BedrockCore::getRemainingTime(const unique_ptr<BedrockCommand>& command
     }
 
     // Both of these are positive, return the lowest remaining.
-    return isProcessing ? min(processTimeout, adjustedTimeout) : adjustedTimeout;
+    return chrono::microseconds(isProcessing ? min(processTimeout, adjustedTimeout) : adjustedTimeout);
 }
 
 bool BedrockCore::isTimedOut(unique_ptr<BedrockCommand>& command, SQLite* db, const BedrockServer* server)
@@ -444,11 +444,11 @@ void BedrockCore::_handleCommandException(unique_ptr<BedrockCommand>& command, c
 
 void BedrockCore::decreaseCommandTimeout(unique_ptr<BedrockCommand>& command, uint64_t timeoutMS)
 {
-    const uint64_t remainingTimeUS = getRemainingTime(command, false);
-    if ((timeoutMS * 1000ull) < remainingTimeUS) {
+    const auto remainingTimeUS = getRemainingTime(command, false);
+    if (chrono::microseconds(timeoutMS * 1000ull) < remainingTimeUS) {
         command->setTimeout(timeoutMS);
-        const int64_t newRemainingTimeUS = getRemainingTime(command, false);
+        const auto newRemainingTimeUS = getRemainingTime(command, false);
         _db.setTimeout(newRemainingTimeUS);
-        SINFO("Decreased command timeout from " << STIMESTAMP(STimeNow() + remainingTimeUS) << " to " << STIMESTAMP(STimeNow() + newRemainingTimeUS));
+        SINFO("Decreased command timeout from " << STIMESTAMP(STimeNow() + remainingTimeUS.count()) << " to " << STIMESTAMP(STimeNow() + newRemainingTimeUS.count()));
     }
 }
--- a/BedrockCore.cpp
+++ b/BedrockCore.cpp
@@ -32,7 +32,7 @@ private:
     bool (*_handler)(int, const char*, string&);
 };
 
-uint64_t BedrockCore::_getRemainingTime(const unique_ptr<BedrockCommand>& command, bool isProcessing)
+uint64_t BedrockCore::getRemainingTime(const unique_ptr<BedrockCommand>& command, bool isProcessing)
 {
     int64_t timeout = command->timeout();
     int64_t now = STimeNow();
@@ -59,7 +59,7 @@ uint64_t BedrockCore::_getRemainingTime(const unique_ptr<BedrockCommand>& comman
 bool BedrockCore::isTimedOut(unique_ptr<BedrockCommand>& command, SQLite* db, const BedrockServer* server)
 {
     try {
-        _getRemainingTime(command, false);
+        getRemainingTime(command, false);
     } catch (const SException& e) {
         // Yep, timed out.
         _handleCommandException(command, e, db, server);
@@ -82,7 +82,7 @@ void BedrockCore::prePeekCommand(unique_ptr<BedrockCommand>& command, bool isBlo
         try {
             SDEBUG("prePeeking at '" << request.methodLine << "'");
             command->prePeekCount++;
-            _db.setTimeout(_getRemainingTime(command, false));
+            _db.setTimeout(getRemainingTime(command, false));
             _db.setAbortRef(command->shouldAbort);
 
             // Make sure no writes happen while in prePeek command
@@ -142,7 +142,7 @@ BedrockCore::RESULT BedrockCore::peekCommand(unique_ptr<BedrockCommand>& command
     try {
         SDEBUG("Peeking at '" << request.methodLine << "'");
         command->peekCount++;
-        _db.setTimeout(_getRemainingTime(command, false));
+        _db.setTimeout(getRemainingTime(command, false));
         _db.setAbortRef(command->shouldAbort);
 
         try {
@@ -240,7 +240,7 @@ BedrockCore::RESULT BedrockCore::processCommand(unique_ptr<BedrockCommand>& comm
     try {
         SDEBUG("Processing '" << request.methodLine << "'");
         command->processCount++;
-        _db.setTimeout(_getRemainingTime(command, true));
+        _db.setTimeout(getRemainingTime(command, true));
         _db.setAbortRef(command->shouldAbort);
         if (!_db.insideTransaction()) {
             // If a transaction was already begun in `peek`, then this won't run. We call it here to support the case where
@@ -352,7 +352,7 @@ void BedrockCore::postProcessCommand(unique_ptr<BedrockCommand>& command, bool i
         try {
             SDEBUG("postProcessing at '" << request.methodLine << "'");
             command->postProcessCount++;
-            _db.setTimeout(_getRemainingTime(command, false));
+            _db.setTimeout(getRemainingTime(command, false));
             _db.setAbortRef(command->shouldAbort);
 
             // Make sure no writes happen while in postProcess command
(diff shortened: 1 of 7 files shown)


BedrockCore.cpp
```diff
@@ -32,7 +32,7 @@ class AutoScopeRewrite {
     bool (*_handler)(int, const char*, string&);
 };
 
-uint64_t BedrockCore::_getRemainingTime(const unique_ptr<BedrockCommand>& command, bool isProcessing)
+uint64_t BedrockCore::getRemainingTime(const unique_ptr<BedrockCommand>& command, bool isProcessing)
```
tylerkaraszewski: Since we're changing this anyway, have it return `std::chrono::microseconds`

BedrockCore.cpp
```diff
@@ -352,7 +352,7 @@ void BedrockCore::postProcessCommand(unique_ptr<BedrockCommand>& command, bool i
         try {
             SDEBUG("postProcessing at '" << request.methodLine << "'");
             command->postProcessCount++;
-            _db.setTimeout(_getRemainingTime(command, false));
+            _db.setTimeout(getRemainingTime(command, false));
```
tylerkaraszewski, line 355: allow `setTimeout` to take `std::chrono::microseconds`

BedrockCore.cpp
```diff
@@ -444,10 +444,10 @@ void BedrockCore::_handleCommandException(unique_ptr<BedrockCommand>& command, c
 
 void BedrockCore::decreaseCommandTimeout(unique_ptr<BedrockCommand>& command, uint64_t timeoutMS)
 {
-    const uint64_t remainingTimeUS = _getRemainingTime(command, false);
+    const uint64_t remainingTimeUS = getRemainingTime(command, false);
     if ((timeoutMS * 1000ull) < remainingTimeUS) {
         command->setTimeout(timeoutMS);
-        const int64_t newRemainingTimeUS = _getRemainingTime(command, false);
+        const int64_t newRemainingTimeUS = getRemainingTime(command, false);
```
tylerkaraszewski: With the above suggested changes, these can be `std::chrono::microseconds` or just `auto`.

sqlitecluster/SQLite.h
```diff
@@ -171,8 +171,10 @@ class SQLite {
     // Prepare to commit or rollback the transaction. This also inserts the current uncommitted query into the
     // journal; no additional writes are allowed until the next transaction has begun.
     // The transactionID and transactionHash, if passed, will be updated with the values prepared for this transaction.
+    // The commitLockTimeout, if passed, will limit the time we wait for the lock. If not, we'll use the max value, which
+    // is effectively no timeout.
     // Note that if this transaction fails to commit, these will not ultimately be accurate.
-    bool prepare(uint64_t* transactionID = nullptr, string* transactionHash = nullptr);
+    bool prepare(uint64_t* transactionID = nullptr, string* transactionHash = nullptr, chrono::microseconds commitLockTimeout = chrono::microseconds::max());
```
tylerkaraszewski, line 177: Can we use 0 as the default? Seems to do the same thing, and I like it better as a default.

BedrockServer.cpp
```diff
@@ -1021,11 +1021,24 @@ void BedrockServer::runCommand(unique_ptr<BedrockCommand>&& _command, bool isBlo
...
                             void (* onPrepareHandler)(SQLite& db, int64_t tableID) = nullptr;
                             bool enableOnPrepareNotifications = command->shouldEnableOnPrepareNotification(db, &onPrepareHandler);
-                            commitSuccess = core.commit(*_syncNode, transactionID, transactionHash, command->getMethodName(), enableOnPrepareNotifications, onPrepareHandler);
+
+                            // We want to calculate the remaining time for this command before we try to commit. It's possible that it will throw.
+                            // If that's the case, we will still allow the commit flow to happen passing a 0 timeout, making it behave as a try_lock
+                            // and still commit it if there's no wait time for the mutex.
+                            try {
+                                // We can use false for isProcessing here since we want to use the total command timeout.
+                                timeToCommit = chrono::microseconds(core.getRemainingTime(command, false));
```
tylerkaraszewski: I don't really get what we're doing here? Are we just trying to see how often this would happen? This seems to make the change largely ineffectual.

tylerkaraszewski (changes requested):

sqlitecluster/SQLite.h
danieldoglas: I didn't use 0 here because there are other calls to prepare that won't necessarily have a timeout defined, e.g.:

https://github.com/Expensify/Bedrock/blob/0785a140a6be6a3b8fd35db0f0462d18ba40d183/sqlitecluster/SQLiteNode.cpp#L2177
https://github.com/Expensify/Bedrock/blob/0785a140a6be6a3b8fd35db0f0462d18ba40d183/sqlitecluster/SQLiteNode.cpp#L2368
https://github.com/Expensify/Bedrock/blob/0785a140a6be6a3b8fd35db0f0462d18ba40d183/sqlitecluster/SQLiteNode.cpp#L1050


In those cases, if we pass 0 as the default, it will do a try_lock, which will automatically return false if we can't get the lock instead of waiting. I'd prefer we leave it as the max value so we can wait for the lock if needed.

BedrockServer.cpp
danieldoglas: I'm not sure what you mean here. we're getting the remaining time in the command and passing it to the try_lock_for, which is what we intended to do from start, right?

The try-catch I added was because `getRemainingTime` throws an SException if the command timed out, and we usually handle those exceptions in core.peek/process/postprocess. So I added a safeguard here to guarantee we wouldn't crash the server with an unexpected throw while commiting, and I treat that in the core.timeout below.

Does that make sense?

tylerkaraszewski: Doesn't passing `0` mean "try forever"? this just passes this number to `try_lock_for` and gets behavior the same as `try_lock`, right? So if the command is already timed out, we wait for as long as we need to get the lock?

danieldoglas: Not unless I'm reading this right? https://en.cppreference.com/cpp/thread/mutex/try_lock


> Tries to lock the mutex. Returns immediately. On successful lock acquisition returns true, otherwise returns false.

So when we pass 0, it won't wait forever, it will try to immediately lock and if it can't, it will return false and timeout

tylerkaraszewski: Oh, I was misunderstanding the behavior of try_lock. My mistake!

sqlitecluster/SQLite.h
danieldoglas: based on this thread https://github.com/Expensify/Bedrock/pull/2610#discussion_r3299710793, I'll leave this as the max value by default

tylerkaraszewski (approved):

flodnv (approved):

PR #2624, opened 2026-06-01
Allow live node priority change

danieldoglas:
### Details

This PR adds a new control command to allow us to change a node's priority while the cluster is running. This will allow us to promote a node without needing a full cluster restart. This will help us in 2 situations:

- In cases where the permafollower is needed for the cluster health (we have 2 other full cluster nodes down)
- When we want to promote changes specifically in one node to run as leader.

There are some safeguards:

- If we try to set a node with the same priority as another node, we reject the command call (if both nodes are currently up)
- If turning a node into a permafollower would make us lose quorum, we reject the command

This is not written to account for all possible concurrences. Meaning, if we run the command simultaneously on 2 different nodes, it's quite possible that we'd have unexpected results.

The tests were very flaky because changes didn't replicate to other nodes before I changed the priorities again, but now they should be stable. No failures in 100 runs locally.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/640540

### Tests

Implemented automated tests in `SetPriorityTest.cpp` to confirm the expected behavior of the command.

_________


sqlitecluster/SQLiteNode.cpp
```diff
@@ -1291,12 +1291,6 @@ void SQLiteNode::_onMESSAGE(SQLitePeer* peer, const SData& message)
             if (!message.isSet("Version")) {
                 STHROW("missing Version");
             }
-            if (peer->permaFollower && (message["Permafollower"] != "true" || message.calc("Priority") > 0)) {
-                STHROW("you're supposed to be a 0-priority permafollower");
-            }
-            if (!peer->permaFollower && (message["Permafollower"] == "true" || message.calc("Priority") == 0)) {
-                STHROW("you're *not* supposed to be a 0-priority permafollower");
-            }
```
danieldoglas, line 1299: We need to remove this since depending on what happens in the cluster, we will send a new LOGIN message. If that's the case, and we've already changed the node priority, the permafollower property might have been changed to the initial cluster stats.

sqlitecluster/SQLiteNode.cpp
```diff
@@ -2903,3 +2899,55 @@ void SQLiteNode::_dieIfForkedFromCluster()
...
+    // can change the state to SEARCHING.
+    unique_lock<decltype(_stateMutex)> lock(_stateMutex);
+    if (newPriority < 0 || newPriority == 1) {
+        STHROW("Priority must be non-negative");
+    }
+
+    const int oldPriority = _priority;
+    SINFO("Changing node priority", {{"oldPriority", to_string(oldPriority)}, {"newPriority", to_string(newPriority)}});
+    _priority.store(newPriority);
+    _originalPriority.store(newPriority);
```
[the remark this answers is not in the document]
danieldoglas: Good bot.

sqlitecluster/SQLiteNode.cpp
tylerkaraszewski: I don't think we have really needed these checks in a long time.

sqlitecluster/SQLiteNode.cpp
```diff
@@ -1314,12 +1308,13 @@ void SQLiteNode::_onMESSAGE(SQLitePeer* peer, const SData& message)
             peer->priority = message.calc("Priority");
             peer->version = message["Version"];
             peer->state = stateFromName(message["State"]);
+            peer->permaFollower = peer->priority == 0;
```
tylerkaraszewski, line 1311: I know it touches more code, but we could probably remove `permafollower` entirely and just check priority.

sqlitecluster/SQLiteNode.cpp
```diff
@@ -1314,12 +1308,13 @@ void SQLiteNode::_onMESSAGE(SQLitePeer* peer, const SData& message)
...
 
             // Is it on the same version as us?
             if (!_haveSeenPeerOnSameVersion && peer->version.load() == _version) {
                 _haveSeenPeerOnSameVersion = true;
                 if (_haveBeenWAITING) {
-                    _priority = _originalPriority;
+                    _priority = _originalPriority.load();
```
tylerkaraszewski: I would expect the existing line works just fine here.

sqlitecluster/SQLiteNode.cpp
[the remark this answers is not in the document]
tylerkaraszewski: We should rename `originalPriority` to `configuredPriority` or something, because it's confusing now.

Priority should either be `configuredPriority` or `-1` I think.

BedrockServer.cpp
```diff
@@ -2129,6 +2130,28 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
...
+    } else if (SIEquals(command->request.methodLine, "SetPriority")) {
+        if (!command->request.isSet("priority")) {
+            response.methodLine = "400 Missing priority";
+        } else {
+            int64_t newPriority = command->request.calc64("priority");
+
+            // Priority 1 is reserved for shutdown.
+            if (newPriority < 0 || newPriority == 1) {
+                response.methodLine = "400 Invalid priority";
+            } else if (!_syncNode) {
```
tylerkaraszewski: We should make a copy of syncNode like we do elsewhere.

tylerkaraszewski (changes requested):
Overall this look pretty good.

sqlitecluster/SQLiteNode.cpp
danieldoglas: So did I, but the build fails:

```
sqlitecluster/SQLiteNode.cpp:551:27: error: overload resolution selected deleted operator '='
  551 |                 _priority = _originalPriority;
      |                 ~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~
```

BedrockServer.cpp
danieldoglas: Done

sqlitecluster/SQLiteNode.cpp
danieldoglas: So, I tried doing this, but it generates other issues. Commit with the changes I tried here: https://github.com/Expensify/Bedrock/pull/2624/commits/7b70468458cde1b141845728af24096a5169f28b


I think the main issue would be that when we start the cluster, we don't send the priority that each node would have in the peerList, but we do send the `Permafollower` property.

So when we try to start the cluster, all peers start with priority 0 until they connect to each other

https://github.com/Expensify/Bedrock/blob/fa766cf72b773dd0f832fb51a6ae1feb0a427103/sqlitecluster/SQLitePeer.cpp#L24

Since everyone is priority 0, all nodes would be permafollowers in our checks, and we fail to start the cluster since there's no quorum.

I think we can remove this, but it should be a separate PR with specific tests related to it.

sqlitecluster/SQLiteNode.cpp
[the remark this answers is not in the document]
danieldoglas: renamed `originalPriority` to `configuredPriority`.

Your point about priority made me think a bit more that if the current node priority is -1 (syncing/waiting) or 1 (shutting down), we would allow it to change it. Let me fix that.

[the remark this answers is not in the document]
danieldoglas: Ok, now we won't change _priority if it's 1 or -1, nor broadcast STATE. That way, if the server is syncing or standing down, we won't "force" it to come back to leading causing weird cases for the cluster.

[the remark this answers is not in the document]
tylerkaraszewski: That's a good idea.

sqlitecluster/SQLiteNode.cpp
```diff
@@ -2895,11 +2891,108 @@ void SQLiteNode::_dieIfForkedFromCluster()
...
     }
 }
+
+int SQLiteNode::setPriority(int newPriority)
+{
+    // Let's lock _stateMutex here since we'll be changing the priority, which
+    // can change the state to SEARCHING.
+    unique_lock<decltype(_stateMutex)> lock(_stateMutex);
+    if (newPriority < 0 || newPriority == 1) {
+        STHROW("Priority must be non-negative");
```
tylerkaraszewski: Message could be better.

tylerkaraszewski (dismissed):

sqlitecluster/SQLiteNode.cpp
flodnv: Indeed, why can't priority be `1`? 

Edit: Ah, I see it further down: `Priority 1 is reserved for shutdown.`. Can you add a comment and udpate the STHROW to that effect? Maybe also make it `STHROW("500 Invalid priority")`?

sqlitecluster/SQLiteNode.cpp
```diff
@@ -2895,11 +2891,108 @@ void SQLiteNode::_dieIfForkedFromCluster()
...
+    }
+
+    // If demoting self to permafollower, ensure the remaining full peers still
+    // satisfy quorum. Minimum full peers required is ceil(totalClusterSize / 2):
+    // a 6-node cluster needs ≥3, a 4-node cluster needs ≥2, etc.
+    if (newPriority == 0 && _priority > 0) {
+        const int totalClusterSize = static_cast<int>(_peerList.size()) + 1;
+        const int minFullPeersForQuorum = (totalClusterSize + 1) / 2;
+        if (otherFullPeers < minFullPeersForQuorum) {
+            STHROW("409 Demoting would break quorym");
```
flodnv:
```suggestion
            STHROW("409 Demoting would break quorum");
```

flodnv (approved):

tylerkaraszewski (approved):

tylerkaraszewski:
@MelvinBot create a new issue, assign it to me. It should ask me to investigate if we broke any testing for permafollowers in terms of counting nodes available for quorum with this change.

PR #2625, opened 2026-06-02
Free listening port earlier in shutdown process

fukawi2:
Close the SQLiteNode peer listening port (`-nodeHost`) early during shutdown so the kernel releases it before the database memory is unmapped, instead of holding it bound for the duration of the `unmap()`.

When a node shuts down with a large memory-mapped database, the kernel keeps the peer listening port bound while it unmaps that memory, which can take a long time and prevents a new process starting.

The port was held because `~SQLiteNode` deletes `pluginDB` (a separate handle whose `sqlite3_close()` unmaps its pages) in the destructor body, while `_port` is a member only destroyed after the body returns. This closes `_port` before `delete pluginDB` in the destructor, and also frees it in `beginShutdown()` so it's released early and deterministically on the graceful path, with `update()` guarded so it won't reopen the port while shutting down.

### Fixed Issues
Fixes https://expensify.slack.com/archives/C094TGUTZ/p1780360333816159?thread_ts=1780356276.321669&cid=C094TGUTZ

### Tests

image

image

### DIFF
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -164,6 +164,12 @@ SQLiteNode::~SQLiteNode()
         delete peer;
     }
 
+    // Close the listening peer port before deleting pluginDB. pluginDB is a separate handle on the database, so closing
+    // it unmaps that handle's memory-mapped pages, which can take a long time when the database is large. _port is a
+    // member and would otherwise only be destroyed after this destructor body returns, leaving the kernel holding the
+    // listening port bound for the duration of that unmap.
+    _port = nullptr;
+
     if (pluginDB != nullptr) {
         delete pluginDB;
     }
@@ -259,6 +265,14 @@ bool SQLiteNode::beginShutdown()
         // Start graceful shutdown
         SINFO("Beginning graceful shutdown.");
         _isShuttingDown = true;
+
+        // Free the listening peer port now rather than waiting for the node to be destroyed during DB teardown. A
+        // shutting-down node won't accept new peer connections (established peer connections are independent of this
+        // listening socket), and the database teardown that follows can take a long time when the memory-mapped
+        // database is large. Until the FD is closed, the kernel keeps the listening port bound while it unmaps that
+        // memory, which prevents a restarted node from rebinding the port.
+        _port = nullptr;
+
         return true;
     }
 
@@ -516,8 +530,9 @@ bool SQLiteNode::update()
     }
     unique_lock<decltype(_stateMutex)> uniqueLock(_stateMutex);
 
-    // If we failed to open the port at creation time, try again on each upate.
-    if (_port == nullptr) {
+    // If we failed to open the port at creation time, try again on each update. We skip this while shutting down,
+    // because we intentionally close the listening port at that point and don't want to reopen it.
+    if (_port == nullptr && !_isShuttingDown) {
         _port = openPort(_host);
     }
 


sqlitecluster/SQLiteNode.cpp
```diff
@@ -259,6 +265,14 @@ bool SQLiteNode::beginShutdown()
         // Start graceful shutdown
         SINFO("Beginning graceful shutdown.");
         _isShuttingDown = true;
+
+        // Free the listening peer port now rather than waiting for the node to be destroyed during DB teardown. A
+        // shutting-down node won't accept new peer connections (established peer connections are independent of this
+        // listening socket), and the database teardown that follows can take a long time when the memory-mapped
+        // database is large. Until the FD is closed, the kernel keeps the listening port bound while it unmaps that
+        // memory, which prevents a restarted node from rebinding the port.
+        _port = nullptr;
```
tylerkaraszewski, line 274: This is broken, we shouldn't free this until cluater communication is done.

sqlitecluster/SQLiteNode.cpp
```diff
@@ -164,6 +164,12 @@ SQLiteNode::~SQLiteNode()
         delete peer;
     }
 
+    // Close the listening peer port before deleting pluginDB. pluginDB is a separate handle on the database, so closing
+    // it unmaps that handle's memory-mapped pages, which can take a long time when the database is large. _port is a
+    // member and would otherwise only be destroyed after this destructor body returns, leaving the kernel holding the
+    // listening port bound for the duration of that unmap.
+    _port = nullptr;
```
tylerkaraszewski, line 171: This doesn't do anything. Well, it does what it says, which is to move the port freeing to before deleting `pluginDB`, but deleting `pluginDB` is not the slow part of this that you're worried about.

The show part of closing the DB is closing the *last* existing filehandle to it. This is owned by `BedrockServer::_dbPool`. 

This currently gets deleted at the end of `BedrockServer::sync()`, but it's right *after* we delete the syncNode (by calling `atomic_store` for it and passing `nullptr`), which should release the peer port.

If the peer port is not released before the final DB deletion starts, it is probably because some other thread is holding a reference to `_syncNode`, but even that shouldn't be an issue, as that thread can release it while the deletion happens.

tylerkaraszewski (changes requested):
I think the diagnosis of this issue is wrong, or at least incomplete.

sqlitecluster/SQLiteNode.cpp
fukawi2: Cluster communication should be being done over established connections though? This is only closing the _listening_ port for _new_ connections?

sqlitecluster/SQLiteNode.cpp
fukawi2: Yeah I'll be honest, I'm out of my depth. What I saw yesterday when bouncing the cluster was that port 4445 was the one that was held during unmmap() - but only on the hosts that had their command port closed when shutdown was initiated. I'm not sure if that really helps knowing though.

sqlitecluster/SQLiteNode.cpp
tylerkaraszewski: Existing connections are not guaranteed to stay open.

sqlitecluster/SQLiteNode.cpp
tylerkaraszewski: Do you have the time and date and host for when this happened so I can look at the logs?

fukawi2: `db2.lax` [Logs](https://vl.expensify.com/select/vmui/#/?step=5s&query=host%3A%3Ddb2.lax&g0.range_input=45m&g0.end_input=2026-06-02T01%3A00%3A00&g0.relative_time=none&limit=10000)

`db1.rno` [Logs](https://vl.expensify.com/select/vmui/#/?step=2m&query=host%3A%3Ddb1.rno&g0.range_input=4h15m&g0.end_input=2026-06-02T04%3A30%3A00&g0.relative_time=none&limit=10000)

tylerkaraszewski:
Does this seem likely to be the log line for when db2.lax was shut down and this issue started on that node?

```
2026-06-02T00:27:35.987610+00:00 db2.lax bedrock: xxxxxx we@dont.know (BedrockServer.cpp:2159) _beginShutdown [main] [info] Beginning graceful shutdown due to 'Terminated', closing command port on '0.0.0.0:4444'.
```

fukawi2:
I can't be positive since it was in the middle of a long saltfab run, but it seems highly likely.

PR #2627, opened 2026-06-04
Add per-identifier time-consumed rate limit to BlockingCommandQueue

stitesExpensify:
### Details
Adds a per-identifier time limit to the blocking queue, alongside the existing count limit. Count only tracks how many commands a user has queued, so it misses the case where one user sends a few very slow commands that tie up the blocking commit thread. This tracks total execution time per user instead.

Both metrics are log-only for now so we can compare them against production data before deciding which to enforce. Set the threshold with `SetBlockingQueueTimeRateLimit`, view it in `Status`, and like the count limit it resets once the queue has been empty for 30 seconds.

According to claude about 0.2% of commands total over 120s so we should get this about 2x/hour

### Fixed Issues
Follow-up to https://github.com/Expensify/Bedrock/pull/2555
https://github.com/Expensify/Expensify/issues/568969

### Tests
Added automated tests
_________


BedrockServer.cpp
```diff
@@ -2072,6 +2081,18 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
         if (command->request.test("ClearBlocks")) {
             _blockingCommandQueue.clearRateLimits();
         }
+    } else if (SIEquals(command->request.methodLine, "SetBlockingQueueTimeRateLimit")) {
```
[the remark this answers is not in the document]
stitesExpensify: updated

BedrockBlockingCommandQueue.cpp
```diff
@@ -137,6 +177,20 @@ size_t BedrockBlockingCommandQueue::setMaxRequestsPerIdentifier(size_t value)
...
+    return _maxTimePerIdentifier.exchange(valueMicros);
+}
+
+void BedrockBlockingCommandQueue::recordExecutionTime(const string& identifier, uint64_t elapsedMicros)
+{
+    if (_maxTimePerIdentifier.load() == 0 || identifier.empty()) {
+        return;
+    }
+    lock_guard<decltype(_rateLimitMutex)> lock(_rateLimitMutex);
+    _identifierTimes[identifier] += elapsedMicros;
```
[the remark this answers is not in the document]
stitesExpensify: Intended behavior. The 30s empty-queue reset is by design (see BedrockBlockingCommandQueue.h:17-21). It only fires after the queue is continuously empty for 30s, which means nothing was queued behind that command and contention has resolved. Leaving as-is.

BedrockBlockingCommandQueue.cpp
```diff
@@ -20,27 +20,47 @@ void BedrockBlockingCommandQueue::push(unique_ptr<BedrockCommand>&& command)
...
-        if (count > maxPerIdentifier) {
-            SINFO("Blocking queue rate limit: rejecting '" << command->request.methodLine
-                  << "' for identifier '" << identifier << "' (count=" << count
-                  << ", threshold=" << maxPerIdentifier << ")");
-            // TODO: enable enforcement after monitoring confirms thresholds are correct in production.
-            // count--;
-            // STHROW("503 Blocking queue rate limited");
+        if (shouldCheckTime) {
+            auto it = _identifierTimes.find(identifier);
+            const uint64_t timeMicros = (it == _identifierTimes.end()) ? 0 : it->second;
```
flodnv: What's "Micros" refering to?

BedrockBlockingCommandQueue.cpp
```diff
@@ -137,6 +177,20 @@ size_t BedrockBlockingCommandQueue::setMaxRequestsPerIdentifier(size_t value)
     return _maxPerIdentifier.exchange(value);
 }
 
+uint64_t BedrockBlockingCommandQueue::setMaxTimePerIdentifier(uint64_t valueMicros)
+{
+    return _maxTimePerIdentifier.exchange(valueMicros);
+}
+
+void BedrockBlockingCommandQueue::recordExecutionTime(const string& identifier, uint64_t elapsedMicros)
```
flodnv: Same Q here

BedrockBlockingCommandQueue.h
```diff
@@ -48,6 +56,8 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
     mutex _rateLimitMutex;
 
     map<string, size_t> _identifierCounts;
+    map<string, uint64_t> _identifierTimes;
     atomic<size_t> _maxPerIdentifier{10};
+    atomic<uint64_t> _maxTimePerIdentifier{120'000'000}; // 120 seconds, in microseconds
```
flodnv: How can one account consume 120 seconds of a single blockingCommit thread within 30 seconds? 😕

BedrockBlockingCommandQueue.cpp
danieldoglas: I think `ms`?

BedrockBlockingCommandQueue.h
danieldoglas: I don't think I understand your question. What do you mean `within 30s`?

The logic is: if the blocking queue is empty for 30s, then clear the identifiers. If not, we will continuously add time to the identifiers, and it can build up for more than 120s for sure.

am I wrong here @stitesExpensify ?

stitesExpensify: Yep what Daniel said.

So one account hitting 120s doesn't mean 120s of work in 30s of clock time, it means 120s of blockingCommit time accumulated across however long the queue stays busy.

1. User A sends a command that takes 30s on the blocking thread → A's total = 30s.
2. While A's command is in the queue, User B adds a command. The queue is now not-empty, so the 30s-empty reset never starts counting.
4. A sends another 30s command → A's total = 60s. B (and others) keep the queue fed the whole time.
5. A sends two more → 90s, then 120s. Now A is over the threshold and we log it.

BedrockBlockingCommandQueue.cpp
flodnv: Ah, so let's call it `timeMS` like we do elsewhere: https://github.com/Expensify/Bedrock/blob/d90029b80d93fc1acf95635694abc40e43c2d505/sqlitecluster/SQLiteNode.cpp#L2403-L2404

BedrockBlockingCommandQueue.h
flodnv: Ahhh, got it, thanks. That feels a bit high, but we can start with that.

> According to claude about 0.2% of commands total over 120s so we should get this about 2x/hour

I'm curious about this, can you share this analysis?

stitesExpensify:
```
. Per-command total on the blockingCommit thread, last 7 days (152,796 commands):

┌────────────────────────┬────────┬───────────────┐
│ Single command exceeds │ Count  │ % of commands │
├────────────────────────┼────────┼───────────────┤
│ 5s                     │ 67,556 │ 44%           │
├────────────────────────┼────────┼───────────────┤
│ 15s                    │ 48,085 │ 31%           │
├────────────────────────┼────────┼───────────────┤
│ 30s                    │ 31,262 │ 20%           │
├────────────────────────┼────────┼───────────────┤
│ 60s                    │ 15,233 │ 10%           │
├────────────────────────┼────────┼───────────────┤
│ 120s                   │ 315    │ 0.2%          │
└────────────────────────┴────────┴───────────────┘

44% of individual blocking commands already take >5s. Since one command over the threshold trips the per-identifier accumulator (and then every subsequent push from that identifier keeps tripping it until the 30s-empty reset), 5s would fire on nearly half of all normal activity. Even 30s fires on 20%.

The clean break is at 120s — only 0.2% of single commands (315 over a week, ~2/hour) exceed it, and those are the genuine outliers. That also lines up with the threat: the worst documented monopolization in the PR was ~230s on one accountID, which blows past 120s reliably.

One caveat: the metric accumulates per identifier across a busy window, so it'll trip somewhat more often than the 0.2% single-command figure — but for log-only monitoring that's the point (we want to see who's accumulating), and 120s keeps it from drowning in normal slow commands.
```

BedrockBlockingCommandQueue.cpp
stitesExpensify: Good call on changing this to match our other code, it's going to be US though FYI since these are microseconds not milliseconds

stitesExpensify: updated!

BedrockBlockingCommandQueue.h
stitesExpensify: re-reading that, I'm wondering if we just had bursts where things took a really long time and that we wouldn't normally hit that with regular traffic since it's just an average of 315/7 days. Reducing the time might be a good idea

danieldoglas: Let's reduce it to 60s then and merge this?

stitesExpensify: done

danieldoglas (approved):

tylerkaraszewski (approved):

PR #2631, opened 2026-06-09
Check if we have already rolled back

tylerkaraszewski:
### Fixed Issues

https://github.com/Expensify/Expensify/issues/646725

Fixes crash after OOM error seen here: https://expensify.slack.com/archives/C094TGUTZ/p1781010136152849

### Tests

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -481,9 +481,6 @@ bool SQLite::beginTransaction(SQLite::TRANSACTION_TYPE type, bool beginOnly)
         STHROW("Attempted to begin transaction while in invalid state: _uncommittedQuery not empty");
     }
 
-    // Reset before the query, as it's possible the query sets these.
-    _autoRolledBack = false;
-
     // We actively track transaction counts incrementing and decrementing to log the number of active open transactions at any given moment.
     _sharedData.openTransactionCount++;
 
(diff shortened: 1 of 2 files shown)


tylerkaraszewski:
@codex review

sqlitecluster/SQLite.cpp
```diff
@@ -1115,9 +1105,16 @@ void SQLite::rollback(const string& commandName)
         _totalTransactionElapsed = _transactionTimer.stop();
 
         // Cancel this transaction
-        if (_autoRolledBack) {
+        if (sqlite3_get_autocommit(_db)) {
+            // Some failures will cancel the current transcation and do an automatic rollback.
```
cead22:
```suggestion
            // Some failures will cancel the current transaction and do an automatic rollback.
```

sqlitecluster/SQLite.cpp
```diff
@@ -1115,9 +1105,16 @@ void SQLite::rollback(const string& commandName)
...
+            // See: https://sqlite.org/c3ref/interrupt.html
+            // and: https://www.sqlite.org/lang_transaction.html ("Response To Errors Within A Transaction")
+            // In these cases, we can test that we are no longer in a transaction using `sqlite3_get_autocommit()`.
+            // If we are in autocommit mode, our transaction has been rolled back.
+            // When this happens, we do not do a rollback here. This allows the externally visible SQLite API to be
+            // consistent and not have to handle this special case. Consumers can just always call `rollback` after a
+            // failed query, regardless of whether or not it was already rolled back internally.
             SINFO("Transaction was automatically rolled back, not sending 'ROLLBACK'.");
```
danieldoglas, line 1117: NAB: do we need to add a comment that this behavior only works like this because we use `BEGIN` in every transaction? Since it would also return true if there was no `BEGIN` call, right?

danieldoglas (dismissed):
LGTM, just a small comment

cead22 (dismissed):

sqlitecluster/SQLite.cpp
tylerkaraszewski: If we don't use begin we don't need to roll back. I don't think we need an extra comment for it, `ROLLBACK` doesn't make sense without `BEGIN` first.

tylerkaraszewski:
@danieldoglas @cead22 want to re-approve after the typo fix?

danieldoglas (approved):

PR #2635, opened 2026-06-11
More reliably abort commands if client disconnects.

tylerkaraszewski:
### Details
This extends our Abort for commands where the client has disconnected from just inside sql queries to before and after each command phase, and during HTTPS requests.

### Fixed Issues
Fixes https://expensify.slack.com/archives/C0B96MS9B8X/p1781194408074179

### Tests

_________

### DIFF
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -123,15 +123,21 @@ void BedrockCommand::_waitForHTTPSRequests()
         // from the the point of the list of known completed request.
         auto requestIt = (_lastContiguousCompletedTransaction == httpsRequests.end()) ? httpsRequests.begin() : _lastContiguousCompletedTransaction;
 
-        // Wait until the command's timeout, or break early if the command has timed out.
+        // Wait until the command's timeout, or break early if the command has timed out or the connection that
+        // originated this command has dropped (in which case there's nobody left to receive the response).
         uint64_t maxWaitUs = 0;
         uint64_t now = STimeNow();
         if (!startTime) {
             startTime = now;
         }
-        if (now < timeout()) {
+        bool aborted = socket && shouldAbort.load();
+        if (now < timeout() && !aborted) {
             maxWaitUs = timeout() - now;
         } else {
+            if (aborted) {
+                SINFO("Abandoning in-flight HTTPS requests, originating connection dropped.");
+            }
+
             // Look at all our uncompleted requests, mark them failed.
             while (requestIt != httpsRequests.end()) {
                 if (!(*requestIt)->response) {
@@ -140,7 +146,7 @@ void BedrockCommand::_waitForHTTPSRequests()
                 requestIt++;
             }
 
-            // Timed everything out, can return.
+            // Timed everything out (or abandoned them), can return.
             break;
         }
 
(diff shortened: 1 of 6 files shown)


tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

danieldoglas (dismissed):
LGTM.

Should we try to set this after acquiring the lock in `SQLite::prepare`? That way, even if we acquired the lock after waiting 60s, we can abort without commiting if the client disconnected.

danieldoglas:
@codex review

tylerkaraszewski:
> Should we try to set this after acquiring the lock in SQLite::prepare? That way, even if we acquired the lock after waiting 60s, we can abort without committing if the client disconnected.

So I think this is a good idea, and thought about adding it, but I think I'm OK omitting or at least deferring it. It requires passing the command or a pointer to the `shouldAbort` flag to be passed all the way down into SQLite::prepare, which isn't that bad, but is a bit messy, and I don't really love it from an encapsulation standpoint. But also, once we've waited a long time for the commit lock, it seems like a waste not to do anything with it, we're basically done at that point.

But I think what I'd like more is to adjust this to be interruptible so that instead of potentially waiting 20 seconds for the lock and then doing nothing with it, we could at least interrupt waiting for the lock when the client disconnects. This is also a little bigger change.

What if we do this as a followup?

tylerkaraszewski:
@danieldoglas - updated, comments addressed.

tylerkaraszewski:
@codex review

danieldoglas:
> But I think what I'd like more is to adjust this to be interruptible so that instead of potentially waiting 20 seconds for the lock and then doing nothing with it, we could at least interrupt waiting for the lock when the client disconnects. This is also a little bigger change.


I agree with this. Let's address in a followup then!

BedrockCore.cpp
```diff
@@ -276,6 +290,7 @@ BedrockCore::RESULT BedrockCore::processCommand(unique_ptr<BedrockCommand>& comm
                 command->_inDBWriteOperation = true;
                 command->process(_db);
                 command->_inDBWriteOperation = false;
+                _throwIfAborted(command);
```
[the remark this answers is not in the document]
tylerkaraszewski: This is the case we are addressing in a followup, I think it's fine.

tylerkaraszewski:
Followup issue: https://github.com/Expensify/Expensify/issues/648226

BedrockCommand.cpp
```diff
@@ -140,7 +146,7 @@ void BedrockCommand::_waitForHTTPSRequests()
                 requestIt++;
             }
 
-            // Timed everything out, can return.
+            // Timed everything out (or abandoned them), can return.
```
danieldoglas, line 149: While I think these changes make sense, I'm kind of conflicted with aborting the command if we did HTTP requests. This could be a problem for cases like when we do billing, right?

danieldoglas (approved):
LGTM, just one question. Not sure if we need to discuss that here or in the slack.

BedrockCommand.cpp
tylerkaraszewski: I know what you mean but there's no real guarantee these succeed anyway. They can time out and fail after an HTTPS request anyway, or we can send a HTTPS request and get disconnected from the remote server while receiving the response. Billing is designed to be resilient to failures anyway, so if we start abandoning commands mid-billing, we could re-run it, but it doesn't actually affect normal billing because those commands are fire-and-forget and are exempt to being cancelled.

PR #2647, opened 2026-06-19
Re-anchor missed SCHEDULED job repeats to preserve stagger

justinpersaud:
### Details
When a scheduled job misses it run due to bwm being offline for any reason, the job will run immediately when bwm is back up again. This results in us losing a stagger between jobs that was previously set for jobs that were run in the past.

For any jobs that have an interval of less than 4h and their last run was missed, find the next scheduled slot when bwm is back to preserve the stagger time.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/649216

### Tests
I added automated tests

I did a manual test as well.

1. create a job scheduled to run in the past with a repeat of scheduled +5m
2. run it
3. confirm the next run after it finishes is still staggered for the future run instead of also being scheduled in the past

note that this does not prevent the initial flood of jobs that will just run immediately when BWM is restarted if the scheduled next run is in the past because that is a much larger change that has a larger blast radius

AI helped me run the manual test case. `bcmd` is just a wrapper to send a command to bedrock

```
expensidev-2404->/vagrant (main) > date
Mon Jun 22 06:11:48 PM UTC 2026
expensidev-2404->/vagrant (main) > which bcmd
expensidev-2404->/vagrant (main) > PAST=$(date -u -d '1 hour ago' '+%Y-%m-%d %H:%M:00')
expensidev-2404->/vagrant (main) > echo "firstRun (missed) = $PAST"
bcmd "CreateJob\r\nname: test/StaggerProof\r\nrepeat: SCHEDULED, +5 MINUTES\r\nfirstRun: $PAST"
firstRun (missed) = 2026-06-23 17:52:00
200 OK
commitCount: 364332
Connection: close
nodeName: bedrock
peekTime: 179
processTime: 979
totalTime: 2873
unaccountedTime: 838
Content-Length: 28

{"jobID":458899167125282596}
expensidev-2404->/vagrant (main) > bcmd "Query\r\nformat: json\r\nquery: SELECT state, nextRun, CAST(STRFTIME('%s',nextRun)-STRFTIME('%s','now') AS INT) AS secs_from_now FROM jobs WHERE name='test/StaggerProof';"
200 OK
commitCount: 364332
Connection: close
nodeName: bedrock
peekTime: 682
totalTime: 1162
unaccountedTime: 479
Content-Length: 95

{"headers":["state","nextRun","secs_from_now"],"rows":[["QUEUED","2026-06-23 17:52:00",-3669]]}
expensidev-2404->/vagrant (main) > bcmd "GetJob\r\nname: test/StaggerProof"
200 OK
commitCount: 364333
Connection: close
nodeName: bedrock
peekTime: 78
processTime: 2681
totalTime: 3651
unaccountedTime: 360
Content-Length: 207

{"created":"2026-06-23 18:53:00","data":{},"jobID":458899167125282596,"lastRun":"","name":"test\/StaggerProof","nextRun":"2026-06-23 17:52:00","priority":500,"repeat":"SCHEDULED, +5 MINUTES","retryAfter":""}
expensidev-2404->/vagrant (main) > date
Tue Jun 23 06:53:38 PM UTC 2026
expensidev-2404->/vagrant (main) > bcmd "FinishJob\r\njobID: 458899167125282596"
200 OK
commitCount: 364334
Connection: close
nodeName: bedrock
peekTime: 179
processTime: 1214
totalTime: 3012
unaccountedTime: 795
Content-Length: 0


expensidev-2404->/vagrant (main) > bcmd "Query\r\nformat: json\r\nquery: SELECT nextRun, CAST(STRFTIME('%s',nextRun)-STRFTIME('%s','now') AS INT) AS secs_from_now, (STRFTIME('%s',nextRun)-STRFTIME('%s','$PAST')) % 300 AS phase_offset FROM jobs WHERE name='test/StaggerProof';"
200 OK
commitCount: 364334
Connection: close
nodeName: bedrock
peekTime: 500
totalTime: 774
unaccountedTime: 273
Content-Length: 93

{"headers":["nextRun","secs_from_now","phase_offset"],"rows":[["2026-06-23 18:57:00",172,0]]}
expensidev-2404->/vagrant (main) > date
Tue Jun 23 06:54:11 PM UTC 2026
```
_________

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -1403,32 +1403,32 @@ string BedrockJobsCommand::_constructNextRunDATETIME(SQLite& db, const string& l
     // Apply the repeat's date modifiers to a base timestamp, returning the result as a quoted SQL
     // literal (or "" on a syntax error).
     auto applyModifiers = [&](const string& startExpr) -> string {
-        string nextRun = startExpr;
+        string current = startExpr;
         for (const string& part : parts) {
             // This isn't supported natively by SQLite, so do it manually here instead.
             if (SToUpper(part) == "START OF HOUR") {
                 SQResult result;
-                if (!db.read("SELECT STRFTIME('%Y-%m-%d %H:00:00', " + nextRun + ");", result) || result.empty()) {
+                if (!db.read("SELECT STRFTIME('%Y-%m-%d %H:00:00', " + current + ");", result) || result.empty()) {
                     SWARN("Syntax error, failed parsing repeat " + part);
                     return "";
                 }
 
-                nextRun = SQ(result[0][0]);
+                current = SQ(result[0][0]);
             } else if (!SIsValidSQLiteDateModifier(part)) {
                 // Validate the sqlite date modifiers
                 SWARN("Syntax error, failed parsing repeat " + part);
                 return "";
             } else {
                 SQResult result;
-                if (!db.read("SELECT DATETIME(" + nextRun + ", " + SQ(part) + ");", result) || result.empty()) {
+                if (!db.read("SELECT DATETIME(" + current + ", " + SQ(part) + ");", result) || result.empty()) {
                     SWARN("Syntax error, failed parsing repeat " + part);
                     return "";
                 }
 
-                nextRun = SQ(result[0][0]);
+                current = SQ(result[0][0]);
             }
         }
-        return nextRun;
+        return current;
     };
 
     string nextRun = applyModifiers(baseExpr);
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -7,6 +7,10 @@
 #undef SLOGPREFIX
 #define SLOGPREFIX "{" << getName() << "} "
 
+// Only SCHEDULED repeats shorter than this re-anchor to now after a missed window; longer intervals
+// keep their scheduled time since they may need to run on a specific calendar date.
+static constexpr int64_t REPEAT_REANCHOR_THRESHOLD_SECONDS = 4 * 60 * 60;
+
 const int64_t BedrockPlugin_Jobs::JOBS_DEFAULT_PRIORITY = 500;
 const string BedrockPlugin_Jobs::name("Jobs");
 const string& BedrockPlugin_Jobs::getName() const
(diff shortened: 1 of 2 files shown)


iwiznia:
> This results in us losing a stagger between jobs that was previously set for jobs that were run in the past

This is not correct as seen in the ReceiptScan jobs. What we did see (according to @danieldoglas) is that we set the nextRun to the past, since it repeat was anchored on SCHEDULED.

iwiznia:
> create a job scheduled to run in the past

With what repeat?

iwiznia:
Why is this targeting `expensify_prod` and not main? 😕 
image

justinpersaud:
oh that was a mistake, going to move to a new pr to fix

iwiznia:
You can change the target branch in the UI

justinpersaud:
@iwiznia i updated the test section to show the repeat

plugins/Jobs.cpp
```diff
@@ -1382,40 +1386,89 @@ string BedrockJobsCommand::_constructNextRunDATETIME(SQLite& db, const string& l
...
+        // Only fixed-offset modifiers (eg "+30 SECONDS") have a constant period we can skip by arithmetic.
+        // Snapping modifiers like "START OF HOUR" round the result, so their realized delta isn't the period;
+        // leave those to their own (coarse, low-volume) catch-up rather than re-anchor them off-schedule.
+        bool fixedInterval = true;
+        for (const string& part : parts) {
+            if (part.empty() || (part.front() != '+' && part.front() != '-')) {
+                fixedInterval = false;
+                break;
             }
+        }
```
iwiznia: Not sure I get this... what's an example of a valid repeat that does not have a `+` or `-` in one of its parts?

justinpersaud: We have two types of repeats

+/- interval
and then a snap, which i something like `START OF HOUR`

so if we have one of the snaps in there, we can't really modify the repeat logic since it's already fixed

iwiznia: Oh got it, had read this code backwards.
Something like `SCHEDULED, +1 MONTH, WEEKDAY 2` could get this fix but it won't right? I think that's ok anyway

iwiznia (dismissed):

justinpersaud:
@danieldoglas did you want to review this one too?

plugins/Jobs.cpp
justinpersaud: yeah that is right

danieldoglas:
reviewing now

plugins/Jobs.cpp
```diff
@@ -1382,40 +1386,89 @@ string BedrockJobsCommand::_constructNextRunDATETIME(SQLite& db, const string& l
...
+                     "STRFTIME('%s', " + nowExpr + ") - STRFTIME('%s', " + SQ(lastScheduled) + ");", result) || result.empty()) {
+            SWARN("Failed evaluating SCHEDULED repeat staleness for '" << repeat << "'");
             return "";
-        } else {
-            SQResult result;
-            if (!db.read("SELECT DATETIME(" + nextRun + ", " + SQ(part) + ");", result) || result.empty()) {
-                SWARN("Syntax error, failed parsing repeat " + part);
-                return "";
+        }
+        const bool missedWindow = result[0][0] == "1";
```
danieldoglas, line 1450: all the logic below here only needs to be executed if missed window is true, right? Should we early return here?

danieldoglas: we can also check for `intervalSeconds < REPEAT_REANCHOR_THRESHOLD_SECONDS` around here so we don't do the rest of the logic.

danieldoglas (dismissed):
LGTM, 2 small comments that are NAB imo

plugins/Jobs.cpp
justinpersaud: added that in

danieldoglas (approved):

iwiznia (approved):

PR #2656, opened 2026-06-26
Make SQLite::prepare lock acquisition interruptible on command abort

tylerkaraszewski:
### Details
Took a few iterations to get all the nuances for this correct.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/648226

### Tests
No new tests. This is very difficult to test.

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -1,5 +1,6 @@
 #include "SQLite.h"
 
+#include <chrono>
 #include <linux/limits.h>
 #include <string.h>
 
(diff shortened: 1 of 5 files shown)


tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

tylerkaraszewski:
@codex review

sqlitecluster/SQLite.cpp
```diff
@@ -864,16 +865,55 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash, chrono::m
 
     // We lock this here, so that we can guarantee the order in which commits show up in the database.
     if (!_mutexLocked) {
-        auto start = STimeNow();
-        if (!_sharedData.commitLock.try_lock_for(commitLockTimeout)) {
-            // Couldn't get the lock in time. Roll back the open transaction.
+        auto start = chrono::steady_clock::now();
+        auto finalLockTimeout = start + commitLockTimeout;
+        auto nextLockTimeout = start + min(commitLockTimeout, 1'000'000us);
```
danieldoglas, line 870: Just so we use the same representation we have in line 887
```suggestion
        auto nextLockTimeout = start + min(commitLockTimeout, 1s);
```

sqlitecluster/SQLite.cpp
```diff
@@ -864,16 +865,55 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash, chrono::m
...
+
+        if (abortPtr) {
+            setAbortRef(*abortPtr);
+        }
```
danieldoglas, line 874: NAB for this PR: It seems we call `setAbortRef` and `clearAbortRef` a bunch of times inside `BedrockCore`. So we add and remove it to the db handle several times in the same command, starting on prepeek, then peek, then process, etc.

Since we start core injecting the DB handle to it, and the same core will process the command through its lifecycle, should we add a new method to core itself to pass the ref and have that same ref in the db through all steps of a command instead of adding and removing it?

If we do that, we wouldn't need to pass it down to this function here.

danieldoglas (approved):
changes make sense to me, just 2 comments, one for style and one for discussion.

sqlitecluster/SQLite.cpp
tylerkaraszewski: `min` doesn't work here, the types of the two values need to be identical for it, and `commitLockTimeout` is `chrono::microseconds`. We could cast the second param, but the cast I think is messier than the longer number.

sqlitecluster/SQLite.cpp
danieldoglas: or even better - can we set the abortRef directly here: https://github.com/Expensify/Bedrock/blob/cf3316c70ec0d0d733284b9968424a614ad0f5d0/BedrockServer.cpp#L890

tylerkaraszewski: Several of the earlier commits in this PR were attempting something like what you're suggesting but I kept running into a bunch of different edge cases where it was difficult to ensure that actually works. For example, if `_shouldAbort` is true, then [this code](https://github.com/Expensify/Bedrock/blob/main/sqlitecluster/SQLite.cpp#L1028) can throw because the command is aborted, and it throws up into a `noexcept` function (`SQLiteCore::commit`).

Or another case was that [`_onPrepareHandler`](https://github.com/Expensify/Bedrock/blob/main/sqlitecluster/SQLite.cpp#L886) might throw if it calls `read` or `write` and you get the same thing happening.

danieldoglas (approved):

PR #2661, opened 2026-07-02
Fix silent truncation of large command responses in BedrockServer::_reply()

roryabraham:
### Details

`BedrockServer::_reply()` sent a command's response with a single blocking `socket->send()` call and then immediately shut the socket down, without checking whether the entire serialized response was actually queued to the kernel first. If that one `send()` call queued fewer bytes than the full response (e.g. because a signal interrupts the blocking wait), the remainder sat in the socket's in-process `sendBuffer` and was silently discarded — the client would receive a truncated response with no error indication.

This PR adds `STCPManager::Socket::sendAll()`, which loops `send()` until the socket's `sendBuffer` is fully drained (or a hard error occurs), and switches `_reply()` to use it instead of a single `send()` call — mirroring the drain pattern already used in `SQLiteClusterMessenger` and `STCPManager::postPoll`.

The other `Socket::send()` call sites (`plugins/MySQL.cpp`, `SQLitePeer::sendMessage`, `SHTTPSManager`'s outbound request) are all driven by the ongoing `prePoll`/`postPoll` loop rather than a one-shot blocking thread, so a single `send()` there is correct as-is — the poll loop keeps draining `sendBuffer` on subsequent iterations, and switching those to `sendAll()` would risk busy-spinning the shared poll thread on a non-blocking socket instead. Left those alone.

`sendAll()` only avoids busy-spinning because command-port sockets are accepted as blocking (`S_accept(..., true)`), so the kernel itself blocks `send()` until it can accept more data instead of returning `EAGAIN`. To make sure this doesn't quietly regress if `sendAll()` is ever called on a non-blocking socket in the future, it now asserts the fd is blocking before looping.

### Fixed Issues

N/A — this is a proactively-identified correctness bug, not filed as a separate GitHub issue.

### Tests

Covered by existing unit and integration tests that exercise the command port response path (e.g. `CommandPortTest`).


BedrockServer.cpp
```diff
@@ -2444,7 +2423,9 @@ unique_ptr<BedrockCommand> BedrockServer::buildCommandFromRequest(SData&& reques
         SINFO("Firing and forgetting '" << request.methodLine << "'");
         SData response("202 Successfully queued");
         response["Connection"] = "close";
-        socket.send(response.serialize());
+        // sendAll() loops until the whole response is queued to the kernel (or a hard error occurs); we're about to
+        // shut down the socket, so we won't get another chance to send anything left behind by a partial send().
+        socket.sendAll(response.serialize());
```
tylerkaraszewski, line 2428: This fix is "correct" because we actually shut down the socket here but is sort of pointless because there is no message being sent. We are literally sending `202 Successfully queued` and `close`.

BedrockServer.cpp
```diff
@@ -1648,8 +1639,9 @@ void BedrockServer::_reply(unique_ptr<BedrockCommand>& command)
                 SERROR("Couldn't find plugin '" << pluginName << ".");
             }
         } else {
-            // Otherwise we send the standard response.
-            if (!command->socket->send(command->response.serialize())) {
+            // Otherwise we send the standard response. sendAll() loops until the whole response is queued to
+            // the kernel (or a hard error occurs), rather than assuming a single send() call queued everything.
+            if (!command->socket->sendAll(command->response.serialize())) {
```
tylerkaraszewski, line 1644: So this really isn't correct. If there's more data than we can send here we don't want to block, we want this to get sent on the next loop.

tylerkaraszewski: I think a better way to handle this (maybe it is already handled, I'm not sure) is to check if the sendbuffer is empty when we attempt to close a socket, and then refuse to until it's emptied.

tylerkaraszewski (changes requested):
This is the wrong fix for this. `STCPManager::Socket::Send`

PR #2660, opened 2026-07-01
Append the command to all log lines

iwiznia:
### Details

Needs https://github.com/Expensify/Salt/pull/16927 so command get parsed

### Fixed Issues
$https://github.com/Expensify/Expensify/issues/651951

### Tests
Run some commands, checked log lines had the command in it:
```
2026-07-01T11:12:37.354383+00:00 expensidev-2404 php-fpm: xmx4Kx /api.php submita@submit.com !ecash9.4.25-1! ?api? [info] Bedrock\Client - Starting a request ~~ command: 'GetPersonalDetailsForEmails' clusterName: 'auth' headers: '[emailList: 'submita@submit.com' authToken: '<REDACTED>' shouldCheckIfKnownUser: '' idempotent: '1' logParam: 'submita@submit.com' urlToNewDot: 'https://dev.new.expensify.com:8082/' shouldReadUsingThreads: '1' shouldHandleOnyxUpdates: '1' isNewDotRequest: '1' clientUpdateID: '77329' isFromIntegrationServer: '' maxNumberOfUpdates: '500' commitCount: '41386' requestID: 'xmx4Kx' lastIP: '10.2.2.1' writeConsistency: 'ASYNC' priority: '500' timeout: '110000']'
2026-07-01T11:12:37.354429+00:00 expensidev-2404 php-fpm: xmx4Kx /api.php submita@submit.com !ecash9.4.25-1! ?api? [dbug] Bedrock\Client - APC fetch host configs ~~ 0: 'auth'
2026-07-01T11:12:37.354467+00:00 expensidev-2404 php-fpm: xmx4Kx /api.php submita@submit.com !ecash9.4.25-1! ?api? [info] Bedrock\Client - Possible hosts ~~ nonBlacklistedHosts: '[0: 'auth']'
2026-07-01T11:12:37.354504+00:00 expensidev-2404 php-fpm: xmx4Kx /api.php submita@submit.com !ecash9.4.25-1! ?api? [info] Bedrock\Client - Reusing socket ~~ host: 'auth' cluster: 'auth' pid: '1300'
2026-07-01T11:12:37.354653+00:00 expensidev-2404 bedrock: xmx4Kx submita@submit.com (BedrockServer.cpp:1582) getCommandFromPlugins [socket2] [dbug] Plugin Auth handling command GetPersonalDetailsForEmails ~~ command: 'GetPersonalDetailsForEmails'
[...]

PR #2665, opened 2026-07-02
Log transaction timings for standalone queries.

danieldoglas:
### Details

When we have a rolled back command, all read queries for it still hold a transaction for the duration of the read, but we don't really log those. 

This PR will log the transaction held time when we don't have a transaction.

### Fixed Issues
https://github.com/Expensify/Expensify/issues/655406

### Tests

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -596,12 +596,12 @@ bool SQLite::addColumn(const string& tableName, const string& column, const stri
     return false;
 }
 
-string SQLite::read(const string& query) const
+string SQLite::read(const string& query)
 {
     return read(query, map<string, Parameter>{});
 }
 
-string SQLite::read(const string& query, const map<string, Parameter>& params) const
+string SQLite::read(const string& query, const map<string, Parameter>& params)
 {
     SQResult result;
     if (!read(query, params, result)) {
@@ -613,12 +613,12 @@ string SQLite::read(const string& query, const map<string, Parameter>& params) c
     return result[0][0];
 }
 
-int SQLite::read(const string& query, sqlite3_qrf_spec* spec) const
+int SQLite::read(const string& query, sqlite3_qrf_spec* spec)
 {
     return read(query, {}, spec);
 }
 
-int SQLite::read(const string& query, const map<string, Parameter>& params, sqlite3_qrf_spec* spec) const
+int SQLite::read(const string& query, const map<string, Parameter>& params, sqlite3_qrf_spec* spec) 
 {
     // Execute the read-only query. Skips caching.
     uint64_t before = STimeNow();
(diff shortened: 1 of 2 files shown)


sqlitecluster/SQLite.h
```diff
@@ -99,16 +99,16 @@ class SQLite {
...
-    virtual string read(const string& query) const;
-    virtual string read(const string& query, const map<string, Parameter>& params) const;
+    virtual string read(const string& query);
+    virtual string read(const string& query, const map<string, Parameter>& params);
 
     // Performs a read-only query (eg, SELECT) that uses a query result formatter to format the response.
-    virtual int read(const string& query, sqlite3_qrf_spec* spec) const;
-    virtual int read(const string& query, const map<string, Parameter>& params, sqlite3_qrf_spec* spec) const;
+    virtual int read(const string& query, sqlite3_qrf_spec* spec);
+    virtual int read(const string& query, const map<string, Parameter>& params, sqlite3_qrf_spec* spec);
```
danieldoglas: const functions should not modify the object they're part of, but these change methods are changing local variables (e.g. _readQueryCount)

sqlitecluster/SQLite.cpp
```diff
@@ -653,7 +654,16 @@ bool SQLite::read(const string& query, const map<string, Parameter>& params, SQR
             _queryCache.emplace(make_pair(query, result));
         }
     }
-    _readElapsed += STimeNow() - before;
+    uint64_t timeSpent = STimeNow() - before;
+    _readElapsed += timeSpent;
+
+    // If we're not inside a transaction, SQLite will automatically start a transaction when we execute the read,
+    // and it will also hold the WAL files with those. Let's add a specific log line for this case.
```
tylerkaraszewski: I think the mention of the WAL file here is just confusing for someone reading this a year from now.

tylerkaraszewski:
```suggestion
    // If we're not inside a transaction, this doesn't get added to our normal transaction timing.
```

tylerkaraszewski:
```suggestion
    // If we're not inside a transaction, this doesn't get added to our normal transaction timing.
```

sqlitecluster/SQLite.cpp
```diff
@@ -653,7 +654,16 @@ bool SQLite::read(const string& query, const map<string, Parameter>& params, SQR
...
+    if(!_insideTransaction) {
+        SINFO("Read with auto-transaction timing info", {
+            {"totalTransactionElapsed", to_string(timeSpent / 1000)},
+        });
```
danieldoglas, line 664: no need to worry about logging command here since https://github.com/Expensify/Bedrock/pull/2660 was merged

flodnv (dismissed):

tylerkaraszewski (approved):

PR #2666, opened 2026-07-02
Add a Bedrock job blacklist (CrashBedrockJob) to disable individual jobs without stopping BWM

chiragsalian:
### Details

Adds a Bedrock job blacklist so a single misbehaving job can be surgically disabled without stopping BWM for everyone.

- New localhost control commands: `CrashBedrockJob` (accepts a comma-separated `names` list of GLOB patterns) and `ClearCrashedBedrockJobs`.
- At `GetJob`/`GetJobs` time, any candidate job whose name matches a blacklisted GLOB pattern is transitioned to `FAILED` and never returned to the worker. The extra query only runs when the blacklist is non-empty, so the normal path is unaffected.
- Matching uses SQLite `GLOB`, so `www-prod/UserActivityEvaluator*` matches both `www-prod/UserActivityEvaluator?accountID=856` and `www-prod/UserActivityEvaluatorV2`. This also lets you scope to a single account by including it in the pattern.
- Patterns are broadcast across the cluster and resent to peers on login, so a new leader still enforces them after a failover.
- Surfaced in the `Status` command as `crashedBedrockJobs` / `crashedBedrockJobList`, mirroring how crash commands are shown.
- In-memory only (cleared on a full cluster restart), matching the existing crash-command design.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/649219

### Tests
1. I tested this locally by running the new `CrashBedrockJob` command,
```
CrashBedrockJob
names: www-prod/ProcessAgentZeroRequest*

200 OK
nodeName: bedrock
totalTime: 2500
unaccountedTime: 2500
Content-Length: 0

status

200 OK
nodeName: bedrock
totalTime: 1007
unaccountedTime: 1007
Content-Length: 972

{"blockedIdentifiers":0,"blockedTimeIdentifiers":0,"blockingRateLimitThreshold":10,"blockingTimeRateLimitThresholdMs":60000,"commandCount":1,"commandPortBlockReasons":[],"CommitCount":266693,"crashCommandList":[],"crashCommands":0,"crashedBedrockJobList":["www-prod\/ProcessAgentZeroRequest*"],"crashedBedrockJobs":1,"freelistCount":0,"host":"0.0.0.0:8889","isDetached":false,"isLeader":true,"multiWriteEnabled":true,"multiWriteManualBlacklist":[],"outstandingFramesToCheckpoint":0,"pageCount":44576,"peerList":[],"plugins":[{"name":"Cache"},{"name":"Concierge","version":"62907ef765"},{"name":"DB"},{"name":"ExpensifyTableManager","version":"e2de05c782"},{"name":"Jobs"},{"name":"MySQL"},{"name":"expensifyBackupManager","version":"2a9911f0af"}],"priority":200,"queuedCommandList":[],"state":"LEADING","syncNodeAvailable":true,"syncThreadQueuedCommandList":[],"version":"63a96a407b:Concierge_62907ef765:ExpensifyTableManager_e2de05c782:expensifyBackupManager_2a9911f0af"}
```
2. Then i send a newDot message and confirmed the job failed immediately while bwm was running,
```
SQLite main.db> select * from jobs where name glob 'www-prod/ProcessAgentZeroRequest*';
      created              jobID        state                                      name                                           nextRun       lastRun repeat                                                                                                                                                    data                                                                                                                                                    priority parentJobID retryAfter
------------------- ------------------- ------ ---------------------------------------------------------------------------- ------------------- ------- ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- -------- ----------- -----------
2026-07-02 22:40:04 6986622538182774669 FAILED www-prod/ProcessAgentZeroRequest?reportID=3833388386142844&persona=Concierge 2026-07-02 22:40:04                {"reportID":3833388386142844,"persona":"Concierge","email":"chirag@testjuly2.com","didChatJobExistWhenQueued":false,"pageHTML":"","userAccountID":1583,"triggeringRequestElapsedMS":45,"additionalInputMessages":[],"sidePanelContext":[],"triggeringReportActionID":0,"trigger":"chat_message","               750           0 +10 MINUTES
                                                                                                                                                               _commitCoun...
```
3. Then i ran `ClearCrashedBedrockJobs` and it was cleared. 
```
vagrant@expensidev-2404:/vagrant$ nc localhost 8888
ClearCrashedBedrockJobs

200 OK
nodeName: bedrock
totalTime: 322
unaccountedTime: 322
Content-Length: 0

status

200 OK
nodeName: bedrock
totalTime: 1634
unaccountedTime: 1634
Content-Length: 936

{"blockedIdentifiers":0,"blockedTimeIdentifiers":0,"blockingRateLimitThreshold":10,"blockingTimeRateLimitThresholdMs":60000,"commandCount":1,"commandPortBlockReasons":[],"CommitCount":266718,"crashCommandList":[],"crashCommands":0,"crashedBedrockJobList":[],"crashedBedrockJobs":0,"freelistCount":0,"host":"0.0.0.0:8889","isDetached":false,"isLeader":true,"multiWriteEnabled":true,"multiWriteManualBlacklist":[],"outstandingFramesToCheckpoint":0,"pageCount":44577,"peerList":[],"plugins":[{"name":"Cache"},{"name":"Concierge","version":"62907ef765"},{"name":"DB"},{"name":"ExpensifyTableManager","version":"e2de05c782"},{"name":"Jobs"},{"name":"MySQL"},{"name":"expensifyBackupManager","version":"2a9911f0af"}],"priority":200,"queuedCommandList":[],"state":"LEADING","syncNodeAvailable":true,"syncThreadQueuedCommandList":[],"version":"63a96a407b:Concierge_62907ef765:ExpensifyTableManager_e2de05c782:expensifyBackupManager_2a9911f0af"}
```

Automated coverage: `GetJobTest::testCrashBedrockJobBlacklist` (passes locally, full `GetJob` suite green).

_________

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -1846,6 +1846,13 @@ void BedrockServer::_status(unique_ptr<BedrockCommand>& command)
             content["crashCommandList"] = SComposeJSONArray(crashCommandListArray);
         }
 
+        {
+            // Make it known which bedrock jobs are blacklisted from running.
+            shared_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
+            content["crashedBedrockJobs"] = _crashedBedrockJobPatterns.size();
+            content["crashedBedrockJobList"] = SComposeJSONArray(vector<string>(_crashedBedrockJobPatterns.begin(), _crashedBedrockJobPatterns.end()));
+        }
+
         for (auto& p : _blockingCommandQueue.getState()) {
             content[p.first] = p.second;
         }
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -1959,6 +1959,8 @@ bool BedrockServer::_isControlCommand(const unique_ptr<BedrockCommand>& command)
         SIEquals(command->request.methodLine, "SetBlockingQueueTimeRateLimit") ||
         SIEquals(command->request.methodLine, "ClearBlockingQueue") ||
         SIEquals(command->request.methodLine, "SetPriority") ||
+        SIEquals(command->request.methodLine, "CrashBedrockJob") ||
+        SIEquals(command->request.methodLine, "ClearCrashedBedrockJobs") ||
         SIEquals(command->request.methodLine, "CRASH_COMMAND")
     ) {
         return true;
@@ -2054,6 +2056,20 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
             totalCount += s.second.size();
         }
         SALERT("Blacklisting command (now have " << totalCount << " blacklisted commands): " << request.serialize());
+    } else if (SIEquals(command->request.methodLine, "CrashBedrockJob")) {
+        const list<string> names = SParseList(command->request["names"]);
+        if (names.empty()) {
+            SINFO("Got CrashBedrockJob with no 'names'. Nothing to blacklist.");
+            return;
+        }
+        unique_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
+        for (const auto& name : names) {
+            _crashedBedrockJobPatterns.insert(name);
+        }
+        SALERT("Blacklisting bedrock jobs (now have " << _crashedBedrockJobPatterns.size() << " patterns): " << SComposeList(names));
+    } else if (SIEquals(command->request.methodLine, "ClearCrashedBedrockJobs")) {
+        unique_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
+        _crashedBedrockJobPatterns.clear();
     } else if (SIEquals(command->request.methodLine, "ClearBlockingQueue")) {
         auto commands = _blockingCommandQueue.getAll();
         list<string> methodLines;
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -1975,6 +1975,8 @@ bool BedrockServer::_isNonSecureControlCommand(const unique_ptr<BedrockCommand>&
     // sent from other nodes on the private command port.
     return SIEquals(command->request.methodLine, "SuppressCommandPort") ||
            SIEquals(command->request.methodLine, "ClearCommandPort") ||
+           SIEquals(command->request.methodLine, "CrashBedrockJob") ||
+           SIEquals(command->request.methodLine, "ClearCrashedBedrockJobs") ||
            SIEquals(command->request.methodLine, "CRASH_COMMAND");
 }
 
@@ -2062,14 +2064,28 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
             SINFO("Got CrashBedrockJob with no 'names'. Nothing to blacklist.");
             return;
         }
-        unique_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
-        for (const auto& name : names) {
-            _crashedBedrockJobPatterns.insert(name);
+        {
+            unique_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
+            for (const auto& name : names) {
+                _crashedBedrockJobPatterns.insert(name);
+            }
+            SALERT("Blacklisting bedrock jobs (now have " << _crashedBedrockJobPatterns.size() << " patterns): " << SComposeList(names));
         }
-        SALERT("Blacklisting bedrock jobs (now have " << _crashedBedrockJobPatterns.size() << " patterns): " << SComposeList(names));
+
+        // If this originated locally (from an operator on this node), propagate it to the rest of the cluster so a new
+        // leader still enforces the blacklist after a failover. Commands arriving from a peer have a non-empty _source,
+        // so they update this node only and don't re-broadcast, which prevents an infinite broadcast loop.
+        if (command->request["_source"].empty()) {
+            broadcastCommand(command->request);
+        }
     } else if (SIEquals(command->request.methodLine, "ClearCrashedBedrockJobs")) {
-        unique_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
-        _crashedBedrockJobPatterns.clear();
+        {
+            unique_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
+            _crashedBedrockJobPatterns.clear();
+        }
+        if (command->request["_source"].empty()) {
+            broadcastCommand(command->request);
+        }
     } else if (SIEquals(command->request.methodLine, "ClearBlockingQueue")) {
         auto commands = _blockingCommandQueue.getAll();
         list<string> methodLines;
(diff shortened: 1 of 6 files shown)


BedrockServer.cpp
```diff
@@ -2054,6 +2065,34 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
...
+                _crashedBedrockJobPatterns.insert(name);
+            }
+            SALERT("Blacklisting bedrock jobs (now have " << _crashedBedrockJobPatterns.size() << " patterns): " << SComposeList(names));
+        }
+
+        // If this originated locally (from an operator on this node), propagate it to the rest of the cluster so a new
+        // leader still enforces the blacklist after a failover. Commands arriving from a peer have a non-empty _source,
+        // so they update this node only and don't re-broadcast, which prevents an infinite broadcast loop.
+        if (command->request["_source"].empty()) {
+            broadcastCommand(command->request);
```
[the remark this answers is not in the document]
chiragsalian: Done — switched to an explicit `isPropagated` marker in 63a96a40 so peer-delivered commands don't re-broadcast.

BedrockServer.cpp
```diff
@@ -1973,6 +1982,8 @@ bool BedrockServer::_isNonSecureControlCommand(const unique_ptr<BedrockCommand>&
     // sent from other nodes on the private command port.
     return SIEquals(command->request.methodLine, "SuppressCommandPort") ||
            SIEquals(command->request.methodLine, "ClearCommandPort") ||
+           SIEquals(command->request.methodLine, "CrashBedrockJob") ||
+           SIEquals(command->request.methodLine, "ClearCrashedBedrockJobs") ||
```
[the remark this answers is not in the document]
chiragsalian: Done — removed both from `_isNonSecureControlCommand` in 63a96a40; propagation works via the private-port localhost branch.

BedrockServer.cpp
```diff
@@ -1973,6 +1982,8 @@ bool BedrockServer::_isNonSecureControlCommand(const unique_ptr<BedrockCommand>&
...
            SIEquals(command->request.methodLine, "CRASH_COMMAND");
```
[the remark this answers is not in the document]
chiragsalian: Done — removed from the non-secure control list in 63a96a40 for the same reason.

plugins/Jobs.cpp
```diff
@@ -762,13 +762,50 @@ void BedrockJobsCommand::process(SQLite& db)
...
         list<string> nonRetriableJobs;
         list<STable> retriableJobs;
         list<string> jobList;
         for (size_t c = 0; c < result.size(); ++c) {
             SASSERT(result[c].size() == 10); // jobID, name, data, parentJobID, retryAfter, created, repeat, lastRun, nextRun, priority
 
+            // Skip any job we just failed for matching the blacklist so it's never returned to the worker.
+            if (crashedJobIDs.count(result[c][0])) {
+                continue;
+            }
```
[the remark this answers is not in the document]
chiragsalian: Done — 1d1e05f1 now fails matching jobs before the candidate query, so it backfills up to numResults.

chiragsalian:
Code LGTM and it works well for me locally. Assigning @tylerkaraszewski to review. Can you check it out and make sure the direction seems correct. The idea is that we want to blacklist a bedrock job instead of stopping bwm when there is a fire. For blacklisting the job im using a glob pattern to make it easier on us for the first version.

chiragsalian:
Bump @tylerkaraszewski to review

tylerkaraszewski:
Back from OOO, will plan to review today.

tylerkaraszewski (changes requested):
I think that basically this entire change should live in `Jobs.cpp/h` and the only real change in `BedrockServer` should be that we add something like:
```
    for (auto plugin : plugins) {
        plugin.second->onNodeLogin(SQLitePeer* peer);
    }
```
To `onNodeLogin`.

And we add a virtual `onNodeLogin` method to `BedrockPlugin` with an empty default implementation.

chiragsalian:
Done in `7b801f6e` + `7dc96b68`. Added a virtual `onNodeLogin` hook to `BedrockPlugin` (BedrockServer just loops plugins), and moved the entire blacklist — state, `CrashBedrockJob`/`ClearCrashedBedrockJobs` (now handled in the Jobs command's `peek()`), Status via `Jobs::getInfo()`, and the peer-login resend — into the Jobs plugin. BedrockServer no longer carries any of it.

BedrockServer.h
```diff
@@ -180,7 +180,7 @@ class BedrockServer : public SQLiteServer {
     // Exposes the replication state to plugins.
     SQLiteNodeState getState() const;
 
-    // When a peer node logs in, we'll send it our crash command list.
+    // When a peer node logs in, we'll send it our crash command list and our blacklisted bedrock job patterns.
```
tylerkaraszewski: Comment is out of date.

plugins/Jobs.cpp
```diff
@@ -18,6 +19,42 @@ const string& BedrockPlugin_Jobs::getName() const
...
+        shared_lock<decltype(_crashedBedrockJobPatternMutex)> lock(_crashedBedrockJobPatternMutex);
+        jobPatterns.assign(_crashedBedrockJobPatterns.begin(), _crashedBedrockJobPatterns.end());
+    }
+    if (jobPatterns.empty()) {
+        return;
+    }
+
+    // Send the current blacklist to the joining peer so it enforces the same patterns if it becomes leader. Tag it as
+    // propagated so the peer stores it without re-broadcasting, and set a timeout so this send can't block for long.
+    SALERT("Sending " << jobPatterns.size() << " blacklisted bedrock job patterns to node " << peer->name << " on login");
```
tylerkaraszewski: Do we want this to alert? It's going to perpetually open bugbot issues that are not resolvable.

chiragsalian:
Updated and retested. Thoughts on the changes.

tylerkaraszewski (dismissed):
One minor critique and a comment that may or may not be addressable.

BedrockServer.h
chiragsalian: Done — reworded in `f71aa366`.

plugins/Jobs.cpp
chiragsalian: Good call — downgraded this and the other two blacklist logs to SINFO in `f71aa366` so they don't open bugbot issues on routine events. I don't see any reason for these to alert.

chiragsalian:
Addressed both comments, ready for review.

tylerkaraszewski (approved):

plugins/Jobs.cpp
```diff
@@ -769,6 +878,11 @@ void BedrockJobsCommand::process(SQLite& db)
         for (size_t c = 0; c < result.size(); ++c) {
             SASSERT(result[c].size() == 10); // jobID, name, data, parentJobID, retryAfter, created, repeat, lastRun, nextRun, priority
 
+            // Skip any candidate we just failed for matching the blacklist so it's never returned to the worker.
+            if (crashedJobIDs.count(result[c][0])) {
+                continue;
+            }
+
```
danieldoglas, line 885: If we fail all jobs present in the result, are there any implications we might not be seeing? 

asking because there was an SASSERT above to confirm that results.size > 0

chiragsalian: No i don't think so. Plus there are many other cases where results.size would be greater than 0 but we don't return it. Also when i tested locally things seemed fine.

PR #2672, opened 2026-07-07
Only log read only queries if they take more than 100ms

danieldoglas:
### Details

Filter log so we only log transactions that took more than 100ms

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/655406

### Tests
N/A
_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -658,7 +658,8 @@ bool SQLite::read(const string& query, const map<string, Parameter>& params, SQR
     _readElapsed += timeSpent;
 
     // If we're not inside a transaction, this doesn't get added to our normal transaction timing.
-    if (!_insideTransaction) {
+    // This log line will happen a lot. Let's log only when it takes more than 100ms to complete.
+    if (!_insideTransaction && _readElapsed > 100'000) {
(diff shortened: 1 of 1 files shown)


aldo-expensify (approved):

sqlitecluster/SQLite.cpp
```diff
@@ -658,7 +658,8 @@ bool SQLite::read(const string& query, const map<string, Parameter>& params, SQR
     _readElapsed += timeSpent;
 
     // If we're not inside a transaction, this doesn't get added to our normal transaction timing.
-    if (!_insideTransaction) {
+    // This log line will happen a lot. Let's log only when it takes more than 100ms to complete.
+    if (!_insideTransaction && _readElapsed > 100'000) {
```
rlinoz, line 662: I am seeing a lot of 

```
Read with auto-transaction timing info ~~ totalTransactionElapsed: '0' command: 'OpenApp'
```

Maybe we should update the if to also check timeSpent?

aldo-expensify: +1

danieldoglas: I used the wrong var 😬 
https://github.com/Expensify/Bedrock/pull/2683

PR #2673, opened 2026-07-08
Exclude volatile per-request fields from Jobs crash blacklist

MelvinBot:
### Details

`crashIdentifyingValues` was populated with **every** request field, now it's only populate with name and data. Adding other values like requestID made it so a blacklisted entry couldn't match another

### Fixed Issues
https://github.com/Expensify/Expensify/issues/656214

### Tests
None (@tylerkaraszewski any suggestions?)

_________


plugins/Jobs.cpp
```diff
@@ -121,8 +121,15 @@ bool BedrockJobsCommand::peek(SQLite& db)
 {
     const string& requestVerb = request.getVerb();
 
-    // Jobs commands can only crash if they look identical.
+    // Jobs commands can only crash if they look identical. We key the crash blacklist on the
+    // semantically meaningful request fields, but exclude volatile, per-request fields that are
+    // never identical across two requests (e.g. `requestID`). Including them would make the
+    // blacklist entry impossible to match, rendering the poison-pill protection a no-op.
```
cead22:
```suggestion
    // Jobs commands can only crash if they look identical. 
    // We exclude unique request fields so blacklist entries can be matched
```

cead22 (changes requested):

plugins/Jobs.cpp
```diff
@@ -116,15 +116,38 @@ void BedrockPlugin_Jobs::upgradeDatabase(SQLite& db)
...
 // ==========================================================================
 bool BedrockJobsCommand::peek(SQLite& db)
 {
     const string& requestVerb = request.getVerb();
 
     // Jobs commands can only crash if they look identical.
-    for (const auto& name : request.nameValueMap) {
-        crashIdentifyingValues.insert(name.first);
-    }
+    populateCrashIdentifyingValues();
```
tylerkaraszewski: Instead of doing the AI thing where it adds 50 lines to make a 3 line change, I would simply change the loop removed here with:

```
crashIdentifyingValues.insert("name");
crashIdentifyingValues.insert("data");
```

No changes to the .h file, -3/+2 here.

test/tests/jobs/CrashIdentifyingValuesTest.cpp
```diff
@@ -0,0 +1,74 @@
+#include <libstuff/SData.h>
```
tylerkaraszewski: I find this entire test quite useless. It inserts two items in a `map` and then spends 74 lines asserting that the map contains those two items. We already know that maps work without this test class.

tylerkaraszewski (changes requested):

cead22:
Any suggestions for tests?

tylerkaraszewski (approved):

cead22 (approved):

danieldoglas (approved):

PR #2677, opened 2026-07-13
Reject blocking queue commands

danieldoglas:
### Details
Start enforcing blocking queue rate limits

### Fixed Issues
https://github.com/Expensify/Expensify/issues/568969#top

### Tests

_________


BedrockBlockingCommandQueue.h
```diff
@@ -58,6 +58,7 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
     map<string, size_t> _identifierCounts;
     map<string, uint64_t> _identifierTimes;
     atomic<size_t> _maxPerIdentifier{10};
-    atomic<uint64_t> _maxTimePerIdentifier{30'000'000}; // 30 seconds, in microseconds
+    atomic<uint64_t> _maxTimePerIdentifier{60'000'000}; // 60 seconds, in microseconds
+    atomic<uint64_t> _maxTimePerIdentifierToLog{10'000'000}; // 10 seconds, in microseconds
```
tylerkaraszewski, line 62: Why are these units different? I know we already use `timeUS` above, but it makes more sense to either divide that by 1000 or multiply this by 1000 and use the same units for all these values.

tylerkaraszewski (changes requested):

BedrockBlockingCommandQueue.h
danieldoglas: I'm confused, we're using US everywhere. We only use MS on the log lines. Am I missing something?

tylerkaraszewski: I don't know. I think I just read one of these wrong. Sorry.

tylerkaraszewski (approved):

PR #2681, opened 2026-07-16
Log duplicate escalated write requests

rlinoz:
### Details

Duplicate escalated write requests may reach the leader and be executed more than once, causing write amplification (identified during fire analysis).

This is a protective, observability-first step: on the leader, we now track the ids of escalated requests currently in flight and detect when the same request arrives again while a previous copy is still being processed. For now duplicates are only **logged** (via `SWARN`) — they are **not** dropped — so we can first confirm whether/how often this actually happens in production.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/649211

### Tests

Added `EscalateTest::duplicateEscalatedRequestDetected`, which escalates two identical write commands sharing one id from a follower (kept overlapping in flight on the leader via a slow process step) and verifies the leader's `duplicateEscalatedRequestCount` increased.

_________

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -2631,10 +2631,7 @@ void BedrockServer::handleSocket(Socket&& socket, bool fromControlPort, bool fro
                             command->response.methodLine = "500 Server Shutting Down";
                             _reply(command);
                         } else {
-                            // Escalated requests arrive on the private command port. Track each one by its cluster-unique
-                            // id so we can spot duplicates: the same request arriving again while a previous copy is still
-                            // being processed on the leader, causing write amplification. For now we only log these, we
-                            // don't drop them.
+                            // Track each request by its cluster-unique id so we can spot if we have a duplicates problem
                             string escalatedID;
                             if (fromPrivateCommandPort) {
                                 escalatedID = command->id;
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -2454,6 +2454,56 @@ string BedrockServer::generateCommandID()
     return args["-nodeName"] + "#" + to_string(_requestCount++);
 }
 
+size_t BedrockServer::escalatedRequestFingerprint(const SData& request)
+{
+    // A follower stamps these transport headers onto the request while escalating (see
+    // SQLiteClusterMessenger::_sendCommandOnSocket), and they differ between two escalations of the same write: `ID`
+    // is the follower's own command id, `timeout` is the recomputed remaining time, and the rest carry per-escalation
+    // state. We skip them so retries of one write (identical `requestID` and payload, re-sent by PHP to a different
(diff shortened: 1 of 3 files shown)


rlinoz:
@danieldoglas is this what you had in mind?

danieldoglas:
@rlinoz do you know where the ID is coming from? Does it come from PHP?

rlinoz:
From what I understand It comes from here https://github.com/Expensify/Bedrock/blob/ac42ad646117ece1eb35e91ad3b9286a94891b7d/BedrockServer.cpp#L1574

We generate the ID in Bedrock to build the command.

danieldoglas:
Yeah, so that's kind of a problem. The issue is that PHP is resending the command to followers, which are escalated to leader. So leader would always receive a different ID on those cases

rlinoz:
aaah I got it wrong, from the issue I thought followers were duplicating the commands, that makes more sense. I will rethink this, thanks!

BedrockServer.cpp
```diff
@@ -2453,6 +2454,56 @@ string BedrockServer::generateCommandID()
...
+        if (excludedHeaders.count(name)) {
+            continue;
+        }
+        canonical += name;
+        canonical += ':';
+        canonical += value;
+        canonical += '\n';
+    }
+    canonical += request.content;
+    return hash<string>{}(canonical);
```
rlinoz, line 2481: I think hashing the request is better than sending another unique ID from PHP, mostly because it is less moving parts.

I say another because there is the `requestID` but we forward that to other commands, so not unique enough.

Do you have any thoughts @danieldoglas ?

danieldoglas: hm yes this one might work. cc @tylerkaraszewski wtyt?

tylerkaraszewski (changes requested):
I don't think this is particularly useful the way that it's implemented. It also adds another global mutex that locks on every command, which is a thing that I really try *not* to do as it limits total throughput in the cluster. I am especially wary of doing it for a logging change.

I also think it basically doesn't work. PHP resends requests when it has already sent to one host and did not receive a response. It is expected that the host that receives the duplicated request is a different bedrock node than the one that received the first request. For write requests, it's true, they will all eventually end up on leader, but there's no guarantee they end up there at the same time. I'm not sure what the timeout PHP currently has for resending after not receiving a response, but even if it's 5 seconds, there's a good chance the first command has already finished before the second one arrives.

We know where the duplicated requests are coming from already, PHP sends them twice because it never got a response to the first one. Why don't we just log in PHP when we send a request more than once, and associate all of the requestIDs that we sent with that logline. We can probably write a logging tool that pulls all of the duplicated commands from the 5 minutes worth of logs before this PHP logline fires, and it would be both more performant and more definitive than the bedrock implementation.

rlinoz:
Closing this PR for now based on https://expensify.slack.com/archives/C05CBC62HGW/p1784829434479869

PR #2686, opened 2026-07-17
Create new getReadQueryCount acessor

Valforte:
### Details
Add a new acessor `getReadQueryCount` so we can count the number of read queries a test does using `db.getReadQueryCount` in an Auth test.

### Fixed Issues
Requirement to fix https://github.com/Expensify/Expensify/issues/659611

### Tests
Compiled everything and ran all Auth tests

_________


flodnv (changes requested):
https://github.com/Expensify/Expensify/issues/659611#issuecomment-5023793944

PR #2687, opened 2026-07-22
Refuse HTTPS requests on the blockingCommit thread

danieldoglas:
### Details

Some Auth commands only fire their optional HTTPS request once they've been escalated to the serialized `blockingCommit` thread (worker 0), rather than during `peek()` on a regular worker thread. This happens when the earlier worker-thread `peek()` runs saw different data/conditions and queued no request, so Bedrock's "skip peek if there were httpsRequests" optimization doesn't apply — after the command conflicts >3 times (e.g. on the `nameValuePairsName` index) it lands in the blocking queue, and its `peek()` there finally issues the HTTP call.

Firing an HTTPS request from that single serialized thread holds it for the full network round-trip (and any retry storms), serializing the whole blocking queue behind one slow third-party call. In one trace there were 613 commands waiting behind a single `SendDataToHubspot` that spent ~8.5s in `peek()`.

This PR refuses to start an HTTPS request on the `blockingCommit` thread, failing the command cleanly instead of stalling the queue. It adds a `thread_local bool isBlockingCommitThread` flag (mirroring the existing `isSyncThread`), set for worker 0 in `BedrockServer::worker`, and guards a single choke-point:

- **`SStandaloneHTTPSManager::Transaction`'s constructor** throws `500 Refused` when `isBlockingCommitThread` is set. Every outbound request — regardless of manager — constructs exactly one `Transaction` before opening its socket, so this one guard covers them all. That matters because `Stripe::_httpsSend` and `Hubspot::_httpsSend` are their *own* methods (not overrides of the base) that open sockets via their own `socketStarter`; a guard in the base `_httpsSend` would miss them, and Hubspot is the primary offender here.

Because the `Transaction` is built during the `peek()` that issues the request, the throw:
- fires **before the socket opens** (the ctor runs before `new Socket` in every manager), and
- propagates out of `peek()` where `BedrockCore::peekCommand` turns it into a clean `500` reply — analogous to the existing sync-thread guard, but failing the instant a request would start rather than after `peek()` completes.

It only fires for requests *started* on the blocking thread. Reconstructing already-completed requests (`BedrockCommand::deserializeHTTPSRequests`, when a leader receives an escalated command) happens in `buildCommandFromRequest` on the command-reception thread — never worker 0 — so those are unaffected (verified by the passing Escalate tests).

Expected volume of failures is low (~400/week) and acceptable — better to fail these and address the fallout in follow-up issues than keep serializing the blocking queue behind slow external calls.

Note: this PR is the primary fix from the issue. The secondary metric-accuracy fix (`exclusiveTransactionLockTime`) will follow in a separate PR.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/661688

### Tests

Added `HTTPSBlockingCommitTest` (clustertest). It adds a `httpsblockingcommit` test command that queues an HTTPS request only when it runs on the blockingCommit thread and is otherwise a conflicting write, then fires many concurrent copies with `MaxConflictRetries=1` so some escalate to worker 0. It asserts that:
- the command that re-peeks on the blockingCommit thread comes back `500 Refused - HTTPS request attempted on blockingCommit thread` (covering both the wait-in-peek path and the return-false-then-worker-loop path), and
- every request still gets a response (nothing hangs).

Verified in the local VM:
1. `./clustertest -only HTTPSBlockingCommit` → Passed: 2, Failed: 0.
2. Regression: `./clustertest -only HTTPS,Escalate,BlockingQueueRateLimit` → all passed (normal HTTPS requests, escalation with reconstructed HTTPS requests, and blocking-queue mechanics all still work).

Manual/prod validation after deploy: re-run the issue's LogSearch for `thread:blockingCommit AND "for network requests"` — the count should drop to ~0, replaced by `500 Refused` at the expected low volume.

_________

### DIFF
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -118,13 +118,8 @@ void BedrockCommand::_waitForHTTPSRequests()
 {
     // The blockingCommit thread (worker 0) runs commands under an exclusive transaction, serialized against the whole
     // blocking queue. Waiting on an HTTPS request here would block that thread for the full network round-trip,
-    // stalling every command behind it. This catches commands that wait from within `peek()` (the throw propagates
-    // out and `BedrockCore::peekCommand` turns it into a clean failure), including Stripe and Hubspot, which bypass
-    // `SStandaloneHTTPSManager::_httpsSend` with their own implementations. The worker loop's own call to
-    // `waitForHTTPSRequests` pre-checks this same condition and fails the command without relying on the throw.
-    // We only refuse when there is genuine pending network work; blocking-queue commands with no HTTPS request must
-    // pass through untouched. Scheduled requests are started from the loop below, so throwing now stops them before
-    // the socket opens.
+    // stalling every command behind it. This will guarantee that we're throwing an error if the command calls
+    // waitForHTTPSRequests(db) in peek. If this is running between peek/process, the check will be in runCommand.
     if (isBlockingCommitThread && !areHttpsRequestsComplete()) {
         STHROW("500 Refused - HTTPS request attempted on blockingCommit thread");
     }
--- a/BedrockCommand.cpp
+++ b/BedrockCommand.cpp
@@ -118,11 +118,13 @@ void BedrockCommand::_waitForHTTPSRequests()
 {
     // The blockingCommit thread (worker 0) runs commands under an exclusive transaction, serialized against the whole
     // blocking queue. Waiting on an HTTPS request here would block that thread for the full network round-trip,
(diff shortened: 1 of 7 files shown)


tylerkaraszewski (changes requested):
What if we threw from the constructor for STCPManager::Socket? I think that would cover the cases in auth that use alternate implementations of _httpsSend as well.

tylerkaraszewski (approved):

flodnv (approved):

PR #2693, opened 2026-07-29
Make the default nextRun for jobs when not passed have ms precision

iwiznia:
### Details
When many jobs run in a very short timeframe and they all create onyx updates, we then create `SendFetchableOnyxUpdates` jobs for them. If those jobs run out of order, they force the App to call `GetMissingOnyxMessages` which means the SequentialQueue stays paused and App can't do other requests. If the updates arrive in order, then there's no need to call `GetMissingOnyxMessages` at all.
Since all `SendFetchableOnyxUpdates` jobs were created in the same second, the order is determined by the random jobID, which means that a job created first might run after a job created later. This adds ms precision so that jobs run in order.

### Fixed Issues
Context https://expensify.slack.com/archives/C08CZDJFJ77/p1785332347559139?thread_ts=1785235703.173929&cid=C08CZDJFJ77

### Tests
- Sent a comment in App
- Checked `SendFetchableOnyxUpdates` jobs had ms in the nextRun:
```
sqlite> select * from jobs;
      created              jobID        state                            name                                   nextRun         lastRun repeat                  data                  priority parentJobID retryAfter
------------------- ------------------- ------ -------------------------------------------------------- ----------------------- ------- ------ -------------------------------------- -------- ----------- -----------
2026-07-29 13:47:43 1011095460230348294 QUEUED www-prod/ProcessAgentZeroRequest?reportID=               2026-07-29 13:47:43                    {"reportID":213777350431729,"persona        750           0 +10 MINUTES
                                               213777350431729&persona=Concierge                                                               ":"Concierge","email":"ionatan@
                                                                                                                                               expensify.com","didChatJobExistWhenQue
                                                                                                                                               ued":false,"pageHTML":"","
                                                                                                                                               userAccountID":4,"triggeringRequestEla
                                                                                                                                               ...

2026-07-29 13:47:43 5128217579886465770 QUEUED www-prod/NotifyOfflineUsersAboutActivity?reportID=       2026-07-29 13:57:43                    {"reportID":213777350431729,"accountID      500           0
                                               213777350431729                                                                                 ":4,"targetAccountIDs":null,"
                                                                                                                                               actorAccountID":4,"_commitCounts":{"
                                                                                                                                               auth":42469,"webrock":126271}}

2026-07-29 13:47:43 5234437901448725819 QUEUED www-prod/SendFetchableOnyxUpdates?updates=               2026-07-29 13:47:43.769                {"pusherSocketID":"899894.703119","         750           0 +5 MINUTES
                                               7926cc6d5efa753877e69a7f80be730f                                                                updateIDs":["80284","80285","80286"],"
                                                                                                                                               originalRequestID":"kr0sId","logEmail
                                                                                                                                               ":"ionatan@expensify.com","
                                                                                                                                               _commitCounts":{"auth":42469,"webrock
                                                                                                                                               ...
```
- Run bwm and checked jobs were correctly processed
_________

### DIFF
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -603,7 +603,7 @@ void BedrockJobsCommand::process(SQLite& db)
             const string& currentTime = SCURRENT_TIMESTAMP();
 
             // If no "firstRun" was provided, use right now
-            const string& safeFirstRun = !SContains(job, "firstRun") || job["firstRun"].empty() ? currentTime : SQ(job["firstRun"]);
+            const string& safeFirstRun = !SContains(job, "firstRun") || job["firstRun"].empty() ? SQ(SCURRENT_TIMESTAMP_MS()) : SQ(job["firstRun"]);
 
             // If no data was provided, use an empty object
             const string& safeData = !SContains(job, "data") || job["data"].empty() ? SQ("{}") : SQ(job["data"]);
--- a/plugins/Jobs.cpp
+++ b/plugins/Jobs.cpp
@@ -748,7 +748,7 @@ void BedrockJobsCommand::process(SQLite& db)
                 "FROM jobs "
                 "WHERE state IN ('QUEUED', 'RUNQUEUED') "
                 "AND priority=" + SQ(request.calc("jobPriority")) + " "
-                "AND " + SCURRENT_TIMESTAMP() + ">=nextRun "
+                "AND " + SQ(SCURRENT_TIMESTAMP_MS()) + ">=nextRun "
                 "AND +name " + (nameList.size() > 1 ? "IN (" + SQList(nameList) + ")" : "GLOB " + SQ(request["name"])) + " " +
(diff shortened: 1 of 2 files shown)


rlinoz (approved):
When retrying we are still using seconds, but I think it is fine for the issue we are fixing

https://github.com/Expensify/Bedrock/blob/121d50df0feb0511c7a9b054cc4cf993c3eb1b76/plugins/Jobs.cpp#L985

A good follow up might be to make bedrock work with ms everywhere.

iwiznia:
> When retrying we are still using seconds, but I think it is fine for the issue we are fixing

Hmmm yeah but that's only for jobs with retryAfter. We could change it, but not sure is worth it. Won't help at all for the problem at hand

mountiny (approved):
LGTM

rafecolton:
FYI reverting this in https://github.com/Expensify/Bedrock/pull/2696, it broke a bunch of integration tests. More in https://expensify.slack.com/archives/C03TQ48KC/p1785353285725229

PR #2694, opened 2026-07-29
Track all commit waits, including no wait

tylerkaraszewski:
### Details
For diagnostic purposes, we want to see how much time we can remove from committing with HC-Tree not requiring waiting on a commit lock.

### Fixed Issues
Fixes GH_LINK

### Tests

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -890,7 +890,7 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash, chrono::m
             // If we never waited at all, we record that.
             waited = false;
         }
-        while (true && !lockAcquired) {
+        while (!lockAcquired) {
             lockAcquired = _sharedData.commitLock.try_lock_until(nextLockTimeout);
             if (lockAcquired || chrono::steady_clock::now() >= finalLockTimeout) {
                 break;
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -884,7 +884,13 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash, chrono::m
         }
 
         bool lockAcquired = false;
-        while (true) {
+        lockAcquired = _sharedData.commitLock.try_lock();
+        bool waited = true;
+        if (lockAcquired) {
+            // If we never waited at all, we record that.
+            waited = false;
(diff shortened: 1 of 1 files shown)


tylerkaraszewski:
@codex review

sqlitecluster/SQLite.cpp
```diff
@@ -884,7 +884,13 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash, chrono::m
...
 
         bool lockAcquired = false;
-        while (true) {
+        lockAcquired = _sharedData.commitLock.try_lock();
+        bool waited = true;
+        if (lockAcquired) {
+            // If we never waited at all, we record that.
+            waited = false;
+        }
+        while (true && !lockAcquired) {
```
iwiznia:
```suggestion
        while (!lockAcquired) {
```

sqlitecluster/SQLite.cpp
```diff
@@ -921,9 +928,10 @@ bool SQLite::prepare(uint64_t* transactionID, string* transactionhash, chrono::m
             return false;
         }
 
-        auto elapsed = chrono::steady_clock::now() - start;
-        if (elapsed > 5ms) {
+        if (waited) {
             SINFO("Waited " << chrono::duration_cast<chrono::microseconds>(elapsed) << " for commit lock.");
+        } else {
+            SINFO("Acquired commit lock immediately.");
```
iwiznia, line 934: Hmmm maybe we should remove this and log the elapsed every time? That way graph is still correct as it would track 0?

iwiznia (changes requested):

sqlitecluster/SQLite.cpp
tylerkaraszewski: I think it may still track `0` though. I'm not sure how likely it is we can get `0us` as a time when we waited, but I wouldn't doubt that it's possible, and that's a different case from successfully getting the lock on the first try. It might not be enough of a difference to matter, but I feel this provides more info. I could change it though, I'm not sure if the increase in accuracy makes a real difference.

iwiznia (approved):

tylerkaraszewski:
@MelvinBot - create an issue, make it weekly, assign it to me. It should be to undo this change after we've gathered the required data to see how long we spend waiting on this lock.

PR #2696, opened 2026-07-29
Revert "Make the default nextRun for jobs when not passed have ms precision"

neil-marcellini:
Reverts Expensify/Bedrock#2693

See the [Slack conversation here](https://expensify.slack.com/archives/C03TQ48KC/p1785356139769739?thread_ts=1785353285.725229&cid=C03TQ48KC) for an explanation of the problem.


rafecolton (approved):
cc @iwiznia @mountiny @rlinoz

PR #2697, opened 2026-07-30
Add journalEntries view

aldo-expensify:
### Details

Adds a view `journalEntries` to UNION all journal tables so it is easier to write custom queries targeting all journal tables

<details>
<summary>The view (click to view it)</summary>

```sql
CREATE VIEW journalEntries AS
SELECT
    id,
    query,
    hash
FROM
    journal
UNION
ALL
SELECT
    id,
    query,
    hash
FROM
    journal0000
UNION
ALL
SELECT
    id,
    query,
    hash
FROM
    journal0001
UNION
ALL
SELECT
    id,
    query,
    hash
FROM
    journal0002
UNION
ALL
SELECT
    id,
    query,
    hash
FROM
    journal0003
UNION
ALL
SELECT
    id,
    query,
    hash
FROM
    journal0004
UNION
ALL
....
```

</details>

The code here should keep the view up to date with all journal tables if the table number ever changes.

Currently this is done in the readdb.sh here: https://github.com/Expensify/Salt/blob/554c37ac5d10959f8d2ae6890a925c50ec614f13/shared_pkgs/bedrock/files/readdb.sh#L309-L355

I think we can drop that special code and param `-j` from readdb.sh if we create this view.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/661721

### Tests

1. Compile Bedrock
2. Restart bedrock
3. Verify that the `journalEntries` view is created correctly in the bedrock db
3. Compile Auth using this version of bedrock
3. Restart auth
3. Verify that the `journalEntries` view is created correctly in the auth db

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -213,7 +213,6 @@ vector<string> SQLite::initializeJournal(sqlite3* db, int minJournalTables)
         }
     }
 
-    // Expose one read-only surface for ad-hoc journal queries without changing the critical internal journal queries.
     string journalEntriesQuery;
     for (const string& journalName : journalNames) {
         if (!journalEntriesQuery.empty()) {
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -212,6 +212,41 @@ vector<string> SQLite::initializeJournal(sqlite3* db, int minJournalTables)
             break;
         }
     }
+
+    // Expose one read-only surface for ad-hoc journal queries without changing the critical internal journal queries.
+    string journalEntriesQuery;
+    for (const string& journalName : journalNames) {
+        if (!journalEntriesQuery.empty()) {
+            journalEntriesQuery += " UNION ALL ";
+        }
+        journalEntriesQuery += "SELECT id, query, hash FROM " + journalName;
+    }
+    const string journalEntriesViewQuery = "CREATE VIEW journalEntries AS " + journalEntriesQuery;
+
+    // Avoid changing the schema on every restart, but refresh the view if the set of journal tables changes.
+    SQResult journalEntriesView;
+    SASSERT(!SQuery(db, "SELECT type, sql FROM sqlite_master WHERE name = 'journalEntries'", journalEntriesView));
+    if (journalEntriesView.empty()) {
+        SASSERT(!SQuery(db, journalEntriesViewQuery));
+        SHMMM("Created journalEntries view.");
+    } else {
+        SASSERT(journalEntriesView.size() == 1);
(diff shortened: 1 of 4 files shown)


sqlitecluster/SQLite.cpp
```diff
@@ -212,6 +212,40 @@ vector<string> SQLite::initializeJournal(sqlite3* db, int minJournalTables)
...
+        }
+        journalEntriesQuery += "SELECT id, query, hash FROM " + journalName;
+    }
+    const string journalEntriesViewQuery = "CREATE VIEW journalEntries AS " + journalEntriesQuery;
+
+    // Avoid changing the schema on every restart, but refresh the view if the set of journal tables changes.
+    SQResult journalEntriesView;
+    SASSERT(!SQuery(db, "SELECT type, sql FROM sqlite_master WHERE name = 'journalEntries'", journalEntriesView));
+    if (journalEntriesView.empty()) {
+        SASSERT(!SQuery(db, journalEntriesViewQuery));
```
iwiznia, line 229: I assume this is instant and thus safe to run when we deploy this code?

iwiznia (approved):

sqlitecluster/SQLite.cpp
aldo-expensify: That is my understanding

tylerkaraszewski (approved):
I'm approving this because I think this code is fine, though I wonder if anything that actually uses it will have performance issues.

sqlitecluster/SQLite.cpp
```diff
@@ -212,6 +212,40 @@ vector<string> SQLite::initializeJournal(sqlite3* db, int minJournalTables)
...
+        SHMMM("Created journalEntries view.");
+    } else {
+        SASSERT(journalEntriesView.size() == 1);
+        SASSERT(journalEntriesView[0][0] == "view");
+        if (journalEntriesView[0][1] != journalEntriesViewQuery) {
+            SASSERT(!SQuery(db, "BEGIN IMMEDIATE"));
```
danieldoglas, line 235: what does this do?

aldo-expensify: Seems like it starts the transaction as write transaction from the start. Without it SQlite would wait to see a write statement before making it a write transaction. That is what I understood from the documentation at least: https://sqlite.org/lang_transaction.html

I'm guessing this would still work if you drop `IMMEDIATE`.

aldo-expensify:
> though I wonder if anything that actually uses it will have performance issues.

Same issues we would have with `readdb.sh -j`, right? or do you see any difference?

danieldoglas (approved):

flodnv:
NAR

PR #2701, opened 2026-08-03
Enforce blocking-queue time limit at dequeue

danieldoglas:
### Details
This PR adds a check when dequeueing commands from the blocking queue to confirm if the identifier is blocked at execution time.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/662853

### Tests
- Added unit tests for this case.
_________

### DIFF
--- a/BedrockBlockingCommandQueue.cpp
+++ b/BedrockBlockingCommandQueue.cpp
@@ -77,14 +77,23 @@ unique_ptr<BedrockCommand> BedrockBlockingCommandQueue::_dequeue()
 {
     auto command = BedrockCommandQueue::_dequeue();
 
+    const string blockingIdentifier = command->blockingQueueRateLimitIdentifier;
+
     // Decrement rate limit count when a command leaves the queue.
-    if (!command->blockingQueueRateLimitIdentifier.empty() && _maxPerIdentifier.load() > 0) {
+    if (!blockingIdentifier.empty() && _maxPerIdentifier.load() > 0) {
         lock_guard<decltype(_rateLimitMutex)> lock(_rateLimitMutex);
         _decrementIdentifierCount(command->blockingQueueRateLimitIdentifier);
     }
 
     if (_queue.empty() && _emptyTime.load() == 0) {
         _emptyTime.store(STimeNow());
+    }
+
+    // If this command has a blocking queue identifier, check if it's over the time limit. If so, fill the response methodLine with 503
+    // and set the command as completed. By doing so, we will skip processing the command in `BedrockServer::runCommand`.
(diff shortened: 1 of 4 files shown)


BedrockServer.cpp
```diff
@@ -716,6 +716,16 @@ void BedrockServer::worker(int threadId)
...
             const string blockingIdentifier = (threadId == 0) ? command->blockingQueueRateLimitIdentifier : "";
+
+            // If this command has a blocking queue identifier, check if it's over the time limit. If so, respond with 503 and continue.
+            if (!blockingIdentifier.empty() && _blockingCommandQueue.isIdentifierOverTimeLimit(blockingIdentifier, command->request.methodLine)) {
+                command->response.methodLine = "503 Blocking queue rate limited (time)";
+                command->complete = true;
+                _reply(command);
+                command = nullptr;
+                continue;
+            }
```
tylerkaraszewski: I think we could override `BedrockBlockingCommandQueue::get()` instead of doing this.

It would become something like:
```
BedrockBlockingCommandQueue::get(string blockingIdentifier) {
    auto command = BedrockCommandQueue::get();
    if (blockingIdentifier.size() && isIdentifierOverTimeLimit(blockingIdentifier, command->request.methodLine)) {
    command->response.methodLine = "503 Blocking queue rate limited (time)";
    command->complete = true;
    }
    return command;
}
```

I think this has better encapsulation and keeps all this logic in the blocking queue.

tylerkaraszewski (changes requested):

BedrockServer.cpp
danieldoglas: Done

tylerkaraszewski (approved):

BedrockBlockingCommandQueue.cpp
```diff
@@ -20,55 +20,32 @@ void BedrockBlockingCommandQueue::push(unique_ptr<BedrockCommand>&& command)
...
-            if (timeUS > maxTimePerIdentifierToLog) {
-                SINFO("Blocking queue rate limit (time), logging", {
-                    {"command", command->request.methodLine},
-                    {"identifier", identifier},
-                    {"timeMS", to_string(timeUS / 1000)},
-                    {"thresholdMS", to_string(maxTimePerIdentifier / 1000)}
-                });
-            }
+        if (count > maxPerIdentifier) {
+            SINFO("Blocking queue rate limit: rejecting '" << command->request.methodLine
```
flodnv, line 43: NAB: Maybe we can remove this, as we are pretty sure that "command count" is not as good a metric as "command total time"

flodnv (approved):

PR #2709, opened 2026-08-05
Per-identifier locking for the blocking queue rate limiter

danieldoglas:
### Details
This PR changes how we handle the blocking commit thread rate limits.

The logic we applied here is:
- We have a 180s time window
- Each command can spend up to 40s in the blocking queue during that period
- Each accountID can spend up to 20s in the blocking queue during that period
- If a command or accountID reach their limits, we will block them from going to the blocking queue for 60s.

Every time a command finishes in the blocking queue, we will:

- add the time spent in that command in _commandStates, keyed by command name
- add the time spent for that account in _accountStates, keyed by accountID

The StateMap object has a mutex that is briefly held while we retrieve an entry from the map.

The entry in the map is an IdentifierState, which also has a mutex, a blockedUntil property (so the checkes when we're inserting the commands to the blocking queue is O(1)) and a list of <startTime, timeSpent> for commands of that identifier, which we'll use to calculate how long they spent in the blocking queue in the last 180s every time a command is finished.

So the max we'll hold a mutex for someone is if they have several commands going to the blocking queue, or if we have the same command going to the blocking queue too much. But if that happens, there's a high chance we will block them either way.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/662853

### Tests
Compiles clean; existing unit + cluster tests unaffected (no behavior change yet).

_________


BedrockBlockingCommandQueue.cpp
```diff
@@ -16,47 +16,31 @@ BedrockBlockingCommandQueue::BedrockBlockingCommandQueue() :
 {
 }
 
-void BedrockBlockingCommandQueue::push(unique_ptr<BedrockCommand>&& command)
+uint64_t BedrockBlockingCommandQueue::_now() const
 {
-    const string identifier = command->blockingQueueRateLimitIdentifier;
+    return STimeNow();
+}
```
danieldoglas, line 22: This is here so we can implement automated tests

BedrockServer.cpp
```diff
@@ -713,15 +713,17 @@ void BedrockServer::worker(int threadId)
...
                   << commandQueue.size() << " commands in " << (threadId ? "" : "blocking") << " queue.");
 
-            // Capture the identifier and start time so we can attribute worker-0 execution time
-            // back to the per-identifier rate-limit accumulator after the command finishes.
+            // Capture the account and command name (before the command is moved) so we can attribute worker-0
+            // execution time back to the blocking-queue rate limiter after the command finishes. We time every
+            // command run on the blocking thread, recording against its command name (always) and account (when set).
             const string blockingIdentifier = (threadId == 0) ? command->blockingQueueRateLimitIdentifier : "";
-            const uint64_t blockingStart = (threadId == 0 && !blockingIdentifier.empty()) ? STimeNow() : 0;
+            const string blockingCommandName = (threadId == 0) ? command->request.methodLine : "";
```
tylerkaraszewski: How about just `commandName = command->request.methodLine`?

There's no advantage to not setting it for non-blocking threads, it just wont get used.

BedrockServer.cpp
```diff
@@ -2097,12 +2099,36 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
...
         _enableConflictPageLocks = command->request.test("enable");
     } else if (SIEquals(command->request.methodLine, "SetBlockingQueueTimeRateLimit")) {
-        if (command->request.isSet("MaxTimePerIdentifierMs")) {
-            int64_t maxTimeMs = command->request.calc64("MaxTimePerIdentifierMs");
-            if (maxTimeMs >= 0) {
-                uint64_t previousUS = _blockingCommandQueue.setMaxTimePerIdentifier(static_cast<uint64_t>(maxTimeMs) * 1000);
-                response["previousMaxBlockingQueueTimePerIdentifierMs"] = to_string(previousUS / 1000);
-                SINFO("Setting blocking queue max time per identifier to " << maxTimeMs << "ms");
+        if (command->request.isSet("WindowMS")) {
+            int64_t windowMS = command->request.calc64("WindowMS");
```
tylerkaraszewski: You can collapse the above two lines into just `int64_t windowMS = command->request.calc64("WindowMS");`

it will end up `0` if it's not set. Similar below.

BedrockBlockingCommandQueue.cpp
```diff
@@ -72,96 +56,169 @@ void BedrockBlockingCommandQueue::clear()
...
-        uint64_t emptyTime = _emptyTime.load();
-        if (emptyTime > 0 && STimeNow() - emptyTime >= 30'000'000) {
-            _identifierTimes.clear();
+    const uint64_t now = _now();
+
+    // Collect the tracked count and the currently-blocked identifiers in a map. Holds the map mutex while briefly
+    // locking each entry to read `blockedUntil`; no code path takes an entry lock before the map mutex, so this
+    // can't deadlock.
+    auto inspect = [now](StateMap& map, size_t& tracked, list<string>& blocked) {
+        lock_guard<decltype(map.mapMutex)> lock(map.mapMutex);
```
tylerkaraszewski, line 77: I don't really like this lock wrapping a full walk of the entire map, even more so since it waits for a lock on every single item in the map. It seems like it's only used in `Status` so maybe performance wise it's not that big of a deal.

However, the two locks are worth considering. Are you sure there's no way to cause a deadlock with this?

tylerkaraszewski (changes requested):

BedrockBlockingCommandQueue.cpp
```diff
@@ -16,47 +16,31 @@ BedrockBlockingCommandQueue::BedrockBlockingCommandQueue() :
...
-    const string identifier = command->blockingQueueRateLimitIdentifier;
+    return STimeNow();
+}
 
-    // Reject before enqueuing if the identifier is over the allowed time spent in the blocking queue.
-    if (isIdentifierOverTimeLimit(identifier, command->request.methodLine)) {
+void BedrockBlockingCommandQueue::push(unique_ptr<BedrockCommand>&& command)
+{
+    if (isBlocked(command->blockingQueueRateLimitIdentifier, command->request.methodLine)) {
         STHROW("503 Blocking queue rate limited (time)");
```
flodnv, line 27: NAB: Auth uses `429 Too Many Requests` for something very similar, maybe we should standardize these.

BedrockServer.cpp
```diff
@@ -713,15 +713,17 @@ void BedrockServer::worker(int threadId)
             SINFO("Dequeued command " << command->request.methodLine << " (" << command->id << ") in worker, "
                   << commandQueue.size() << " commands in " << (threadId ? "" : "blocking") << " queue.");
 
-            // Capture the identifier and start time so we can attribute worker-0 execution time
-            // back to the per-identifier rate-limit accumulator after the command finishes.
+            // Capture the account and command name (before the command is moved) so we can attribute worker-0
```
flodnv:
```suggestion
            // Capture the account and command name so we can attribute worker-0
```

BedrockServer.cpp
```diff
@@ -2097,12 +2099,36 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
     } else if (SIEquals(command->request.methodLine, "SetConflictPageLocks")) {
         _enableConflictPageLocks = command->request.test("enable");
     } else if (SIEquals(command->request.methodLine, "SetBlockingQueueTimeRateLimit")) {
```
flodnv, line 2103: Please add this to Step 4 of https://stackoverflowteams.com/c/expensify/questions/17450/17451#17451

BedrockBlockingCommandQueue.cpp
```diff
@@ -16,47 +16,31 @@ BedrockBlockingCommandQueue::BedrockBlockingCommandQueue() :
...
-    try {
-        // Base class acquires its own (non-recursive) `_queueMutex`.
-        BedrockCommandQueue::push(move(command));
-    } catch (...) {
-        // The command never entered the queue. Restore the empty timestamp so the
-        // 30-second auto-reset timer isn't lost.
-        _emptyTime.store(previousEmptyTime);
-        throw;
-    }
+    // Base class acquires its own (non-recursive) `_queueMutex`.
```
flodnv:
```suggestion
    // Base class acquires its own `_queueMutex`.
```

BedrockBlockingCommandQueue.cpp
```diff
@@ -16,47 +16,31 @@ BedrockBlockingCommandQueue::BedrockBlockingCommandQueue() :
...
+    BedrockCommandQueue::push(move(command));
 }
 
 /**
- * Dequeues command and inspects _queue to update _emptyTime, and rejects the command if its identifier is over the time limit.
+ * Dequeues a command and rejects it if its account or command name is rate limited.
```
flodnv: NAB, function doc should be in .h files

BedrockBlockingCommandQueue.h
```diff
@@ -11,49 +15,98 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
 
-    // Enforce per-identifier rate limits before enqueuing. Overrides BedrockCommandQueue::push().
-    // Throws SException("503 ...") if the identifier is rate limited; caller should catch and reply.
-    //
-    // Rate limit state auto-resets when the queue has been continuously empty for 30 seconds — this
-    // is a global reset, not per-identifier. A brief drain (even one dequeue that empties the queue)
-    // starts the timer and will unblock all identifiers once 30 seconds elapse. This is intentional:
-    // the primary threat is a sustained burst from one identifier, so a quiet period is a safe signal
-    // to restore normal operation.
+    // Reject a command before enqueuing if its account or command name is rate limited. Overrides
```
flodnv:
```suggestion
    // Rejects a command before enqueuing if its account or command name is rate limited. Overrides
```

BedrockBlockingCommandQueue.h
```diff
@@ -11,49 +15,98 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+    // BedrockCommandQueue::push(). Throws SException("503 ...") when blocked; the caller catches and replies.
     void push(unique_ptr<BedrockCommand>&& command) override;
 
     // Clear the queue and all rate limiting state.
     void clear();
 
-    // Reset rate limit counters without emptying the queue. Returns the number of tracked identifiers cleared.
+    // Clear all rate-limit state without emptying the queue. Returns the number of tracked accounts and commands cleared.
```
flodnv:
```suggestion
    // Clear all rate-limit states without emptying the queue. Returns the number of tracked accounts and commands cleared.
```

BedrockBlockingCommandQueue.h
```diff
@@ -11,49 +15,98 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
     size_t clearRateLimits();
 
     // Return a table of rate limiting status info for the Status command.
     STable getState();
 
-    // Set the max accumulated worker-0 execution time (microseconds) per identifier. Returns the previous value.
-    uint64_t setMaxTimePerIdentifier(uint64_t valueUS);
+    // Configure the sliding window and thresholds, all in microseconds. A threshold of 0 disables that
+    // dimension. Each setter returns the previous value.
+    uint64_t setWindow(uint64_t windowUS);
```
flodnv: Should all these params be consts?

BedrockBlockingCommandQueue.h
```diff
@@ -11,49 +15,98 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+    uint64_t setCommandThreshold(uint64_t thresholdUS);
+    uint64_t setBlockDuration(uint64_t durationUS);
 
-    // Accumulate elapsed worker-0 execution time for `identifier`. Called by the blockingCommit worker
-    // after each command finishes running. No-op when the time threshold is disabled (== 0). Time is
-    // cumulative per identifier until the empty-queue reset clears it — it is never decremented per command.
-    void recordExecutionTime(const string& identifier, uint64_t elapsedUS);
+    // Record that a command finished on the blocking queue after `elapsedUS` of worker-0 time. Records the
+    // sample against the account (when `accountID` is non-empty) and against the command name.
+    void recordExecutionTime(const string& accountID, const string& commandName, uint64_t elapsedUS);
```
flodnv: NAB: Should `accountID` be abstracted to `identifier`?

BedrockBlockingCommandQueue.h
```diff
@@ -11,49 +15,98 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
 private:
-    // Guards `_identifierTimes`. Separate from the base class `_queueMutex` because the base
-    // mutex is non-recursive and is held while `_dequeue` runs.
-    mutex _rateLimitMutex;
-
-    map<string, uint64_t> _identifierTimes;
-    atomic<uint64_t> _maxTimePerIdentifier{60'000'000}; // 60 seconds, in microseconds
-    atomic<uint64_t> _maxTimePerIdentifierToLog{10'000'000}; // 10 seconds, in microseconds
-    atomic<uint64_t> _emptyTime{0};
+    // One command an identifier finished on the blocking queue. Both times are in microseconds.
```
flodnv:
```suggestion
    // A command that finished on the blocking queue. Both times are in microseconds.
```

BedrockBlockingCommandQueue.h
```diff
@@ -11,49 +15,98 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+    {
+        uint64_t finishTime = 0;
+        uint64_t elapsedTime = 0;
+    };
+
+    // An identifier's recently finished blocking-queue commands, oldest first.
+    typedef deque<RecentlyFinishedCommand> RecentlyFinishedCommandList;
+
+    // Rate-limit state for one identifier (an account or a command name). Each entry has its own mutex, so
+    // different identifiers never contend on one lock. `blockedUntil` is the time (microseconds) an active
```
flodnv, line 67:
```suggestion
    // different identifiers each have their own lock. `blockedUntil` is the time (microseconds) an active
```

BedrockBlockingCommandQueue.h
```diff
@@ -11,49 +15,98 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+    {
+        mutable mutex mapMutex;
+        unordered_map<string, shared_ptr<IdentifierState>> states;
+    };
+
+    // Return a shared_ptr to the state for `key` in `map`, creating it if absent. Holds map.mapMutex only briefly.
+    static shared_ptr<IdentifierState> _getOrCreateState(StateMap& map, const string& key);
+
+    // Return the state for `key` in `map`, or nullptr if absent. Holds map.mapMutex only briefly.
+    static shared_ptr<IdentifierState> _getState(StateMap& map, const string& key);
```
flodnv, line 90: Why can't we always use `_getOrCreateState`?

BedrockBlockingCommandQueue.cpp
```diff
@@ -72,96 +56,169 @@ void BedrockBlockingCommandQueue::clear()
...
-            {"identifier", identifier},
-            {"timeMS", to_string(timeUS / 1000)},
-            {"thresholdMS", to_string(maxTimePerIdentifier / 1000)}
-        });
-        return true;
+    const uint64_t windowUS = _windowUS.load();
+
+    // Drop samples that finished before the window and sum the rest in a single pass, crediting each remaining
+    // sample only for the part that lies inside the window. Samples are ordered oldest-first, so expired ones are
+    // a contiguous prefix at the front. Guard the subtraction so a future-dated sample (clock skew) can't underflow.
```
flodnv: > expired ones are a contiguous prefix at the front.

Can you please rephrase this? 😅

flodnv: > Guard the subtraction so a future-dated sample (clock skew) can't underflow.

Should this comment be above line 184 instead?

BedrockBlockingCommandQueue.cpp
```diff
@@ -72,96 +56,169 @@ void BedrockBlockingCommandQueue::clear()
...
+    uint64_t total = 0;
+    for (auto it = state->commands.begin(); it != state->commands.end();) {
+        const uint64_t age = now > it->finishTime ? now - it->finishTime : 0;
+        if (age >= windowUS) {
+            it = state->commands.erase(it);
+            continue;
+        }
+        total += min(it->elapsedTime, windowUS - age);
```
flodnv: I don't understand this. Why not `total += it->elapsedTime`?

BedrockBlockingCommandQueue.cpp
```diff
@@ -72,96 +56,169 @@ void BedrockBlockingCommandQueue::clear()
...
-        });
-        return true;
+    const uint64_t windowUS = _windowUS.load();
+
+    // Drop samples that finished before the window and sum the rest in a single pass, crediting each remaining
+    // sample only for the part that lies inside the window. Samples are ordered oldest-first, so expired ones are
+    // a contiguous prefix at the front. Guard the subtraction so a future-dated sample (clock skew) can't underflow.
+    uint64_t total = 0;
+    for (auto it = state->commands.begin(); it != state->commands.end();) {
+        const uint64_t age = now > it->finishTime ? now - it->finishTime : 0;
```
flodnv: Can we rename the variable `age` to something that makes the code easier to read? Reading `age >= windowUS` below is quite cryptic.

BedrockBlockingCommandQueue.cpp
```diff
@@ -72,96 +56,169 @@ void BedrockBlockingCommandQueue::clear()
...
+            continue;
+        }
+        total += min(it->elapsedTime, windowUS - age);
+        ++it;
     }
 
-    if (timeUS > _maxTimePerIdentifierToLog.load()) {
+    if (total > thresholdUS) {
+        state->blockedUntil = now + _blockDurationUS.load();
+        SINFO("Blocking queue rate limit (time), blocking", {
```
flodnv:
```suggestion
        SINFO("Blocking queue rate limit reached, blocking", {
```

BedrockBlockingCommandQueue.cpp
```diff
@@ -72,96 +56,169 @@ void BedrockBlockingCommandQueue::clear()
...
+            {"dimension", dimension},
+            {"identifier", key},
+            {"timeMS", to_string(total / 1000)},
+            {"thresholdMS", to_string(thresholdUS / 1000)},
+            {"blockDurationMS", to_string(_blockDurationUS.load() / 1000)}
+        });
+    } else if (total > LOG_THRESHOLD_US) {
+        // Log-only monitoring: the identifier is heavy but still under its block threshold.
         SINFO("Blocking queue rate limit (time), logging", {
```
flodnv:
```suggestion
        SINFO("Blocking queue rate limit above logging threshold", {
```

flodnv (changes requested):

BedrockBlockingCommandQueue.cpp
danieldoglas: Done

danieldoglas: removed everything here

BedrockBlockingCommandQueue.cpp
danieldoglas: Done

BedrockBlockingCommandQueue.cpp
danieldoglas: added a comment

BedrockServer.cpp
danieldoglas: I don't think we want that. If you pass identifierThresholdMS or commandThresholdMS as 0, that means we're not gonna check for those anymore, which means disabling the rate limit. I think we want to allow disabling it, so having the check if the property exists makes sense.

BedrockBlockingCommandQueue.cpp
danieldoglas: removed

BedrockBlockingCommandQueue.h
danieldoglas: done

BedrockBlockingCommandQueue.h
danieldoglas: Done

BedrockBlockingCommandQueue.cpp
danieldoglas: let me know if it's clearer or not

BedrockBlockingCommandQueue.h
danieldoglas: getState won't create a new entry in the map. When we call `isblocked`, we will call `_isBlocked` twice, one for the identifier and another for the command.

If the command is already blocked, we don't want to add that account to the map and waste resources. or vice versa.

BedrockServer.cpp
danieldoglas: Will do once we approve this and there are not more changes

BedrockBlockingCommandQueue.cpp
danieldoglas: I don't see how we would cause a deadlock here... maybe you're seeing something I'm not seeing?

BedrockServer.cpp
danieldoglas: done

BedrockBlockingCommandQueue.cpp
danieldoglas: getState only uses the mapMutex, same as clearRateLimits. The other places always use mapMutex, then the state mutex in that order, so I don't think a deadlock would happen.

BedrockServer.cpp
```diff
@@ -2188,6 +2216,28 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
                 }
             }
         }
+    } else if (SIEquals(command->request.methodLine, "GetCommitHash")) {
```
flodnv, line 2219: Why is this here?

BedrockServer.cpp
```diff
@@ -713,15 +713,17 @@ void BedrockServer::worker(int threadId)
             SINFO("Dequeued command " << command->request.methodLine << " (" << command->id << ") in worker, "
                   << commandQueue.size() << " commands in " << (threadId ? "" : "blocking") << " queue.");
 
-            // Capture the identifier and start time so we can attribute worker-0 execution time
-            // back to the per-identifier rate-limit accumulator after the command finishes.
+            // Capture the account and command name so we can attribute worker-0
+            // execution time back to the blocking-queue rate limiter after the command finishes. We time every
+            // command run on the blocking thread, recording against its command name (always) and account (when set).
```
flodnv:
```suggestion
            // command run on the blocking thread, recording against its command name (always) and identifier (when set).
```

BedrockServer.cpp
danieldoglas: I think I did a wrong pull from main, but this was in main before. Was it reverted?

danieldoglas: merged main again and seems like it worked now

tylerkaraszewski (approved):

flodnv (approved):

PR #2711, opened 2026-08-06
Report the errno from the syscall that actually failed

fukawi2:
### Details

`S_sendconsume` passed `SGetPeerName(s)` and `errno` as sibling arguments to `SCheckNetworkErrorType`. Argument evaluation order is unspecified in C++, and clang evaluates left to right, so `getpeername()` — which fails with `ENOTCONN` on a socket that is still connecting — overwrote `errno` before it was read.

A `send()` that failed with a recoverable `EAGAIN` was therefore reported as a fatal `ENOTCONN` and the socket was closed. This is the source of `send((errno#107)) failed with response 'Transport endpoint is not connected' (#107), closing.`, where both the peer name and the errno come from the same failed `getpeername()`.

Affects any path reaching raw `S_sendconsume` — the proxy `CONNECT` in `SHTTPSProxySocket::send`, and plaintext sockets such as auth→webrock. Direct HTTPS is unaffected because it uses `ssl->sendConsume()`.

Three changes:

- Capture `errno` immediately after `send()` in `S_sendconsume` and use it throughout.
- Capture `errno` immediately after the failing `recvfrom()` in `S_recvappend`.
- Make `SGetPeerName` restore the caller's `errno` so it cannot clobber an unrelated failure.

### Fixed Issues

Fixes https://github.com/Expensify/Expensify/issues/668981

### Tests

1. Build the test binary with `make -j12 test`.
2. Run `./test -only LibStuff` and confirm 36 tests pass.
3. Comment out the `errno = callerErrno;` line in `SGetPeerName`, rebuild, and re-run — `LibStuff::testSGetPeerNamePreservesErrno` fails.
4. Restore the line, rebuild, and confirm it passes again.
5. Run the full suite with `./test` and compare against main — 203 passed / 15 failed with this change, 202 passed / 15 failed on main. The 15 `GetJobTest` failures are pre-existing and unrelated.

Compiler behaviour was verified separately: clang-18 at both `-O0` and `-O2` evaluates left to right and reproduces the clobber, g++-13 does not. Bedrock is built with clang-18 (`Salt/build/pkgs.sls`).

Auth was compiled against this branch: 64/64 steps, zero errors, `auth.a` and `auth.so` linked against the rebuilt `libstuff.a`.

_________

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -2262,6 +2262,9 @@ bool S_recvappend(int s, SFastBuffer& recvBuffer)
         }
     }
 
+    // Save errno from the recvfrom() that ended the loop, before the string building below can overwrite it.
+    const int recvErrno = errno;
+
     // See how we finished
     if (numRecv == 0) {
         return false; // Graceful shutdown; socket closed
@@ -2269,7 +2272,7 @@ bool S_recvappend(int s, SFastBuffer& recvBuffer)
     // Some kind of error -- what happened?
     stringstream addrStr;
     addrStr << fromAddr;
-    return SCheckNetworkErrorType("recv", addrStr.str(), S_errno);
+    return SCheckNetworkErrorType("recv", addrStr.str(), recvErrno);
 }
 
 // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


arosiclair (approved):
Wow, nice find 👍

PR #2716, opened 2026-08-07
Remove QUORUM commits

tylerkaraszewski:
~HOLD for https://github.com/Expensify/Salt/pull/17123 to be on auth.~

Note: follow-ups after this are documented here: https://github.com/Expensify/Expensify/issues/669612

This removes `QUORUM` (and completely obsolete `ONE` level) commits, and with them the whole `ConsistencyLevel` concept. Every commit is async now.

#### We only used this for two cases:
1. The `upgradeDB` path at the beginning of going `LEADING`, which is useful on dev but practically a NOOP on prod anyway, (we keep this, but it is a `blockingCommit` task now).
2. We forced it for `BatchReimbursements` specifically, because this moves lots of money and we don't ever want to lose it. We moved the same functionality here into the batch reimbursements script.

### This removes a lot of complexity that we weren't using.

**`-synchronousCommands`**

This used to force commands to be `QUORUM`, now it forces them to the blocking commit queue (so that they still can't possibly conflict).

**Follower commit counts**

The leader still needs to know how much data each follower has for `COMMITS_LAGGING_BEHIND` command port logic and `freshestPeer`. This was done by `APPROVE_TRANSACTION`: followers sent one for every tenth async transaction with `AsyncNotification`. This remains (see followup to remove it) for now. We added `FOLLOWER_STATUS` which will replace this after all nodes can recognize it.

- new follower + old leader: the follower still sends real approvals for un-prefixed transaction IDs, so an old leader doesn't park forever holding the commit lock. This is the invariant I'd most like a reviewer to not "simplify" away — the sender at `_handlePrepareTransaction` is untouched on purpose.
- old follower + new leader: sees only `ASYNC_` IDs and acks every tenth, as it does now. The new leader still accepts `APPROVE_TRANSACTION`, it just doesn't act on it.

Followup is here:
https://github.com/Expensify/Expensify/issues/669612

We can't remove old messages from the wire protocol until no nodes still use them.

**Replication protocol changes**

The leader now broadcasts a transaction only after its local commit succeeds, rather than sending `BEGIN_TRANSACTION` first and retracting with `ROLLBACK_TRANSACTION` on a commit conflict. That deletes the only `ROLLBACK_TRANSACTION` sender.

The `ROLLBACK_TRANSACTION` *receiver* stays for now, can be removed once the entire cluster is upgraded.

With BEGIN and COMMIT now always adjacent and carrying identical `ID`/`NewCount`/`NewHash`, they can collapse into one message, so this also adds a `TRANSACTION` receiver that applies and commits in one step. Nothing sends it yet — leaders switch over a release later, same staging as `FOLLOWER_STATUS`.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/208443

Really for: https://github.com/Expensify/Expensify/issues/337537

### Tests

Added `SynchronousCommandsTest`, since `-synchronousCommands` had no coverage at all, which felt uncomfortable given it just changed meaning. It checks a listed command lands on the blocking thread both sent directly to the leader and escalated from a follower, plus a negative control — `-synchronousCommands` matches method lines exactly while the test plugin dispatches on a prefix, so `reportcommitthreadunlisted` hits the same handler without being listed and has to come back on a regular worker. Without that third case a bug that sent *everything* to the blocking thread would still pass.

Also stripped the `writeConsistency` lines out of the existing tests. Almost all of them said `"ASYNC"`, which was already the default. Two were interesting: `PermafollowerTest` set `"QUORUM"`, which `SData::calc()` parses as `0`, so despite its comments it has always been testing that commits work with a minority of full peers up. And `TimeoutTest::quorumHTTPS` set a real `"2"` to shove a command with an HTTPS request onto the sync thread — it still reaches `555 Timeout` through the worker path, so I renamed it `httpsRequestTimeout` rather than dropping the coverage.

```
expensidev-2404->Bedrock/test (remove-quorum-commits) > ./test -threads 16
[ TEST RESULTS ] Passed: 217, Failed: 1

Failures:
SSLTest::proxyTest
```

```
expensidev-2404->Bedrock/test/clustertest (remove-quorum-commits) > ./clustertest -threads 16
[ TEST RESULTS ] Passed: 79, Failed: 1

Failures:
ClusterUpgrade::BEFORE_CLASS
```

Same 79 pass with the leader wired to send `TRANSACTION` and no `COMMIT_TRANSACTION` at all, which is what convinced me the merged receiver is right.

Both failures are environmental, not from this change. `SSLTest::proxyTest` needs squid running. `ClusterUpgrade::BEFORE_CLASS` does `ASSERT_EQUAL(system("git fetch --all --tags > /dev/null"), 0)` and my VM has no GitHub key:

```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
...
256 != 0
   assertion #1 at test/clustertest/tests/ClusterUpgradeTest.cpp:26
```

`GracefulFailover`, `MultipleLeaderSync`, `ForkCheck`, `Permafollower` and `ConflictSpam` all pass, which are the ones I actually cared about here.

Compiled Auth against it — `auth.a`, `auth.so` and `authtest` all link clean.

_________


BedrockCommand.h
```diff
@@ -42,6 +40,12 @@ class BedrockCommand : public SQLiteCommand {
     static const uint64_t DEFAULT_TIMEOUT_FORGET = 60'000 * 60; // 1 hour for `connection: forget` commands.
     static const uint64_t DEFAULT_PROCESS_TIMEOUT = 5'000; // 5 seconds.
 
+    // How long a command named in `-synchronousCommands` may hold the blocking commit thread. That thread owns the
+    // commit lock for its whole transaction, and SQLiteNode::_changeState waits on the same lock with no timeout, so
+    // this also bounds how long a node state change can be stalled. It has to stay well under the 30 second peer
+    // receive timeout that would otherwise cost us the leader.
+    static const uint64_t BLOCKING_COMMAND_TIMEOUT = 20'000; // 20 seconds.
```
danieldoglas: I think we should specifically call it synchronous here since we are not applying that timeout for other blocking commits.


```suggestion
    // How long a command named in `-synchronousCommands` may hold the blocking commit thread. That thread owns the
    // commit lock for its whole transaction, and SQLiteNode::_changeState waits on the same lock with no timeout, so
    // this also bounds how long a node state change can be stalled. It has to stay well under the 30 second peer
    // receive timeout that would otherwise cost us the leader.
    static const uint64_t SYNCHRONOUS_COMMAND_TIMEOUT = 20'000; // 20 seconds.
```

BedrockCommand.cpp
```diff
@@ -357,11 +351,9 @@ void BedrockCommand::finalizeTimingInfo()
           "blockingCommitThreadTime:" << blockingCommitThreadTime / 1000 << ", "
           "exclusiveTransactionLockTime:" << exclusiveTransactionLockTime / 1000 <<
           ". Commit: "
-          "worker:" << commitWorkerTotal / 1000 << ", "
-          "sync:" << commitSyncTotal / 1000 <<
+          "worker:" << commitWorkerTotal / 1000 <<
           ". Queue: "
           "worker:" << queueWorkerTotal / 1000 << ", "
-          "sync:" << queueSyncTotal / 1000 << ", "
           "blocking:" << queueBlockingTotal / 1000 << ", "
```
danieldoglas, line 357: this will break log parsing, please address that too in https://github.com/Expensify/Salt/blob/6611eb24400e844781a3387c9106eb3d5b22ee26/shared_pkgs/bedrock/files/logsToStatsd.pl#L107-L176

BedrockCommand.cpp
```diff
@@ -357,11 +351,9 @@ void BedrockCommand::finalizeTimingInfo()
           "blockingCommitThreadTime:" << blockingCommitThreadTime / 1000 << ", "
           "exclusiveTransactionLockTime:" << exclusiveTransactionLockTime / 1000 <<
           ". Commit: "
-          "worker:" << commitWorkerTotal / 1000 << ", "
-          "sync:" << commitSyncTotal / 1000 <<
+          "worker:" << commitWorkerTotal / 1000 <<
           ". Queue: "
           "worker:" << queueWorkerTotal / 1000 << ", "
-          "sync:" << queueSyncTotal / 1000 << ", "
```
iwiznia, line 364: We will need to update `logsToStatsd.pl` so we don't break our stats

sqlitecluster/SQLiteNode.h
```diff
@@ -199,34 +165,13 @@ class SQLiteNode : public STCPManager {
     // to call from other public functions.
     SQLitePeer* getPeerByName(const string& name) const;
 
-    // Pointer to the current on prepare handler to be passed on commit to SQLite
-    void (*onPrepareHandler)(SQLite& _db, int64_t tableID);
-    bool onPrepareHandlerEnabled;
```
danieldoglas, line 204: how are we dealing with onyx updates if we remove this?

test/clustertest/tests/PermafollowerTest.cpp
```diff
@@ -38,17 +38,16 @@ struct PermafollowerTest : tpunit::TestFixture
         tester.stopNode(1);
         tester.stopNode(2);
 
-        // Now 4 out of 6 nodes (1 a permafollower) Do a full quorum commit
+        // Now 4 out of 6 nodes (1 a permafollower). Commit.
         SData request;
         request.methodLine = "Query";
         request["query"] = "INSERT OR REPLACE INTO test (id, value) VALUES(12345, 1 );";
-        request["writeConsistency"] = "QUORUM";
         node4.executeWaitVerifyContent(request, "200");
```
danieldoglas, line 46: Is this test still needed?

BedrockServer.cpp
```diff
@@ -361,253 +322,24 @@ void BedrockServer::sync()
...
-        } catch (const out_of_range& e) {
-            // _syncNodeQueuedCommands had no commands to work on, we'll need to re-poll for some.
+            auto upgrade = make_unique<BedrockServerUpgradeCommand>(*this);
+            upgradeCommand = upgrade.get();
+            upgrade->destructionCallback = &upgradeCommandFinished;
+            upgradeCommandRunning = true;
+            runCommand(move(upgrade), true);
+
+            // This interrupts the next poll loop immediately. This prevents a 1-second wait when running as a single server.
+            _notifyDoneSync.push(true);
```
danieldoglas, line 342: How to we guarantee that this will be the first command to run? What happens if this is a follower that is becoming leader, do we wait for all commands to finish before shifting to LEADING?

iwiznia (dismissed):
> -synchronousCommands
> This used to force commands to be QUORUM, now it forces them to the blocking commit queue (so that they still can't possibly conflict).

Were the quorum commands run in the blocking queue too?

sqlitecluster/SQLiteNode.cpp
```diff
@@ -2403,7 +2125,8 @@ void SQLiteNode::_handlePrepareTransaction(SQLite& db, SQLitePeer* peer, const S
     float transitTimeMS = (float) transitTimeUS / 1000.0;
     float applyTimeMS = (float) applyTimeUS / 1000.0;
     PINFO("[performance] Replicated transaction " << message.calcU64("NewCount") << ", sent by leader at " << leaderSentTimestamp
-          << ", transit/dequeue time: " << transitTimeMS << "ms, applied in: " << applyTimeMS << "ms, should COMMIT next.");
+          << ", transit/dequeue time: " << transitTimeMS << "ms, applied in: " << applyTimeMS << "ms.");
```
danieldoglas, line 2128: this will break https://github.com/Expensify/Salt/blob/6611eb24400e844781a3387c9106eb3d5b22ee26/shared_pkgs/bedrock/files/logsToStatsd.pl#L262, please adjust it

sqlitecluster/SQLiteNode.cpp
```diff
@@ -406,31 +390,21 @@ void SQLiteNode::_sendOutstandingTransactions(const set<uint64_t>& commitOnlyIDs
...
-
-        // If this is marked as "commitOnly", we won't send the BEGIN for it.
-        if (commitOnlyIDs.find(id) == commitOnlyIDs.end()) {
-            // Any commit where we can send a BEGIN and a COMMIT without waiting for acknowledgement is ASYNC.
-            idHeader = "ASYNC_" + idHeader;
-            SData transaction("BEGIN_TRANSACTION");
-            transaction["NewCount"] = to_string(id);
-            transaction["NewHash"] = hash;
-            transaction["leaderSendTime"] = sendTime;
-            transaction["dbCountAtStart"] = to_string(dbCountAtStart);
```
danieldoglas, line 420: is this unused?

sqlitecluster/SQLiteNode.cpp
```diff
@@ -1125,11 +901,6 @@ bool SQLiteNode::update()
             // At this point, we're no longer committing. We'll have returned false above, or we'll have completed any
             // outstanding transaction, we can complete standing down if that's what we're doing.
             if (_state == SQLiteNodeState::STANDINGDOWN) {
-                // See if we're done
-                // We can only switch to SEARCHING if the server has no outstanding write work to do.
-                if (_standDownTimeout.ringing()) {
-                    SWARN("Timeout STANDINGDOWN, giving up on server and continuing.");
-                }
```
danieldoglas, line 1132: I don't really understand what this does, but why did we remove it?

danieldoglas (commented):
LGTM overall, just a few questions

sqlitecluster/SQLiteNode.h
iwiznia: I had the same thought but I think this one is unused now, the one we still use is in `SQLite.cpp`

tylerkaraszewski:
> > -synchronousCommands
> > This used to force commands to be QUORUM, now it forces them to the blocking commit queue (so that they still can't possibly conflict).
> 
> Were the quorum commands run in the blocking queue too?

No, they were run in the sync thread, which also blocked other commands from running. With the sync thread no longer running commands, the blocking commit queue is the closest thing that guarantees no conflicts.

sqlitecluster/SQLiteNode.h
tylerkaraszewski: Yes, we still have it for normal commits, we just don't need it in the sync thread as well anymore.

BedrockServer.cpp
tylerkaraszewski: The check on line 525 `dbReadyToHandleRequests()` will fail and will get stuck in that loop except for the explicit gate we added to let `BedrockServerUpgradeCommand` through there. All other commands are stuck waiting on `dbReadyToHandleRequests()`.

If an individual command is *already running* on a follower becoming leader, we check again on line 714 when peek finishes.

tylerkaraszewski:
Statsd PR:
https://github.com/Expensify/Salt/pull/17166

sqlitecluster/SQLiteNode.cpp
tylerkaraszewski: It's actually so unused that there's some more stuff we can remove. Will update.

test/clustertest/tests/PermafollowerTest.cpp
tylerkaraszewski: It's perhaps not the most useful test but it does still check that a permafollower isn't treated the same as a full peer.

sqlitecluster/SQLiteNode.cpp
tylerkaraszewski: This was already dead code that hasn't been useful in years.

tylerkaraszewski:
@danieldoglas @iwiznia 

Comments addressed and updated! PR created for the logToStatsd change.

iwiznia (approved):

BedrockServer.cpp
danieldoglas: Cool, thanks for explaining!

danieldoglas (approved):
LGTM. It's a lot of changes, let's try to CP instead of making a full deploy to guarantee we're testing only this.

flodnv:
Looks like Tyler did not agree with that approach 😆 🙏

flodnv:
Oh, looks like you tried but failed 😄 Well, we're deploying to both clusters now.

PR #2726, opened 2026-08-17
Replication migration deploy 3: delete the old dialect

tylerkaraszewski:
Hold for https://github.com/Expensify/Bedrock/pull/2725 to be *deployed* everywhere, not just merged. As of writing it's on `main` but hasn't been tagged, so no node is running it yet.

### Details

Deploy 3 of 3, the one that deletes the old dialect. Deploy 1 was [Remove QUORUM commits](https://github.com/Expensify/Bedrock/pull/2716), deploy 2 was [#2725](https://github.com/Expensify/Bedrock/pull/2725).

- Deleted the `BEGIN_TRANSACTION` and `COMMIT_TRANSACTION` dispatch arms in `_replicate` and their `_onMESSAGE` enqueue arms. `_handleBeginTransaction`, `_handlePrepareTransaction` and `_handleCommitTransaction` all stay — the `TRANSACTION` path calls all three. It's the message names that go away, not the handlers.
- Deleted the `ASYNC_` transaction ID prefix. The `ID` header itself stays: a deploy-2 follower still throws `missing ID` without it, even though it now duplicates `NewCount` exactly. Collapsing those two into one is a wire change for a later deploy, not this one.
- Deleted the `APPROVE_TRANSACTION` / `DENY_TRANSACTION` receiver, and the `AsyncNotification` field docs with it. `MIN_APPROVE_FREQUENCY` stays, it paces `FOLLOWER_STATUS`.
- Deleted the four rollback paths that only existed because a follower could hold a prepared transaction between two messages: `_replicate`'s missing-COMMIT_TRANSACTION rollback on exit, the leader-stepped-down rollback in `update()`, the STANDINGDOWN rollback in the `STATE` handler, and the lost-our-leader rollback in `_onDisconnect`.

On that last group, one thing worth saying out loud since it makes them more clearly dead than the issue claims. Three of the four test `_db`, which is the sync thread's handle, but a follower replicates on a handle it takes from `_dbPool`. So even before any of this, a follower's prepared transaction was never visible to the code that was checking for it. They're dead twice over. The fourth one, in `_replicate`, is the one that operated on the right handle, and it's dead because every iteration of that loop now either commits or throws, and the exit check is only reachable between two iterations.

`_onDisconnect`'s comment was the one the issue called out as already describing a world that doesn't exist, and it did — it said we were waiting for leader to "approve or deny" a transaction.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/669612

`For` rather than `$` on purpose: this finishes the migration, but the "adjacent cleanups" checkboxes on that issue (the inert `writeConsistency` headers in `AuthTestHelper.cpp`, `UpgradeToCorporateTest.cpp` and `Client.php`, and the `batchReimbursements.sh` update) aren't in here and would get lost if merging this closed it. Happy to switch it to `$` and move those to their own issue if that's cleaner.

### Tests

`clustertest` on this branch has `ClusterUpgradeTest` failing, and that failure is the point, so I want to be clear about it rather than bury it.

The test builds the most recent release tag as the "old" side of the rolling restart. Right now that's `2026-08-14-373619`, which is deploy 1 — deploy 2 isn't tagged yet. So what the test currently runs is deploy-1 leader against a deploy-3 follower, which is exactly the out-of-order pairing this whole three-deploy dance exists to avoid:

```
    assertion failed: tester->getTester(node).waitForStatusTerm("CommitCount", to_string(leaderCommitCount)) resolved to FALSE
❌ !FAILED! ❌ ClusterUpgradeTest::test (126631ms 🐌)

[ TEST RESULTS ] Passed: 79, Failed: 1
```

Syslog from that run says what happened. `cluster_node_2` is the upgraded one:

```
2026-08-17T21:10:56.963738+00:00 expensidev-2404 bedrock10207: (SQLiteNode.cpp:1397) _onMESSAGE [sync] [info] {cluster_node_2/FOLLOWING} ->{cluster_node_0} unrecognized message: BEGIN_TRANSACTION
2026-08-17T21:11:57.208480+00:00 expensidev-2404 bedrock10207: (SQLiteNode.cpp:1397) _onMESSAGE [sync] [info] {cluster_node_2/FOLLOWING} ->{cluster_node_1} unrecognized message: BEGIN_TRANSACTION
```

A deploy-1 leader sends `BEGIN_TRANSACTION`, the follower has no arm for it, so it drops it and stops replicating, and the commit count it reports never catches up to leader's. Note it drops it at `_onMESSAGE` rather than enqueueing it, so it stalls quietly instead of dying — no `already in a transaction` out of an uncaught thread. Small mercy, and not one to rely on.

To check the pairing that will actually happen, I pinned the old side to deploy 2's merge commit locally:

```
-        // bedrockTagName = "2022-05-06";
+        bedrockTagName = "9a68a32a54eead058e10949e1b16252175f6e68e";
```

```
vagrant@expensidev-2404:/vagrant/Bedrock/test/clustertest$ ./clustertest -only ClusterUpgrade
```

That pin is not committed. Once deploy 2 gets tagged, the test picks it up on its own and this passes without anyone touching it — so the way to confirm this PR is ready to merge is to re-run `clustertest` and see `ClusterUpgrade` go green on its own.

Unit tests, unchanged from `main`:

```
vagrant@expensidev-2404:/vagrant/Bedrock/test$ ./test -threads 16
[ TEST RESULTS ] Passed: 222, Failed: 1

Failures:
SSLTest::proxyTest
```

`proxyTest` needs squid, which my VM doesn't run.

Auth compiles clean against this, `authtest` included:

```
vagrant@expensidev-2404:/vagrant/Auth$ make -j16
2026-08-17 21:18  280678096 ./test/authtest
2026-08-17 21:18  184626424 ./auth/auth.so
2026-08-17 21:18  516593758 ./auth/auth.a
```

_________

### DIFF
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -1393,17 +1393,11 @@ void SQLiteNode::_onMESSAGE(SQLitePeer* peer, const SData& message)
             } else {
                 SINFO("Discarding replication message, stopping FOLLOWING");
             }
-        } else if (SIEquals(message.methodLine, "FOLLOWER_STATUS") ||
-                   SIEquals(message.methodLine, "APPROVE_TRANSACTION") ||
-                   SIEquals(message.methodLine, "DENY_TRANSACTION")) {
-            // These all tell the leader how much data a follower has, and nothing else that we act on. The work is
-            // already done: every message carries CommitCount and Hash, and the code above recorded them with
-            // `peer->setCommit()`. These branches exist only so the messages aren't logged as unrecognized.
-            //
-            // FOLLOWER_STATUS is the message meant for this, and what a follower sends. APPROVE_TRANSACTION and
-            // DENY_TRANSACTION are what a follower running older code sends, so they're accepted here too.
-            // Deliberately no validation: we no longer act on an approval, and throwing would cost us the peer
-            // connection for a message we're about to ignore either way.
+        } else if (SIEquals(message.methodLine, "FOLLOWER_STATUS")) {
+            // This tells the leader how much data a follower has, and nothing else that we act on. The work is already
+            // done: every message carries CommitCount and Hash, and the code above recorded them with
+            // `peer->setCommit()`. This branch exists only so the message isn't logged as unrecognized. Deliberately no
+            // validation: throwing would cost us the peer connection for a message we're about to ignore either way.
         } else if (SIEquals(message.methodLine, "FORKED")) {
             peer->forked = true;
             PINFO("Peer said we're forked, believing them.");
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -34,11 +34,8 @@
 // *** REPLICATION MESSAGES ***
 // TRANSACTION:         A complete transaction: the follower applies it and commits it, with no second message needed.
 //                      A leader only ever broadcasts a transaction it has already committed itself, so there is nothing
-//                      it could later ask the follower to take back. This is what a leader sends.
-// BEGIN_TRANSACTION:   Apply a transaction but don't commit it; wait to be told which way it went. The older,
-//                      two-message form of the above. Nothing in this version sends it; the receiver is still here
-//                      because a leader running older code does, and it goes away once none can.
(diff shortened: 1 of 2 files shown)


danieldoglas (approved):
LGTM!

PR #2731, opened 2026-08-19
Add afterCommit callbacks to SQLite

danieldoglas:
### Details

Adds an `afterCommit` callback mechanism so a plugin can be notified every time this node commits.

This is 1/4 PRs to introduce actions after commit to plugins.

- `SQLite::SharedData` holds the callback list, so one registration covers every handle to the same database file
- `SQLite::commit()` runs the callbacks in the success, after releasing the commitLock and before the checkpoint `BedrockServer::registerAfterCommitCallback()` forwards registration to the pool's base handle.

Plugins are constructed before the DB pool exists. `BedrockServer` builds its plugins in its own constructor, but `_dbPool` is not created until `sync()` runs. `BedrockServer` now holds early registrations and hands them to the pool once it is created. 

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/671183

### Tests

Added automated tests

_________

### DIFF
--- a/BedrockServer.cpp
+++ b/BedrockServer.cpp
@@ -79,21 +79,6 @@ shared_ptr<SQLitePool> BedrockServer::getDBPool()
     return _dbPool;
 }
 
-void BedrockServer::registerAfterCommitCallback(function<void()>&& callback)
-{
-    lock_guard<decltype(_afterCommitCallbackMutex)> lock(_afterCommitCallbackMutex);
-
-    // Kept even after being handed off, so they can be handed to a different SharedData if the pool is re-created
-    // against a different resolved filename.
-    _afterCommitCallbacks.push_back(callback);
-
-    if (_dbPool) {
-        // Registering on the base handle is enough, the callback list lives in the SharedData that every handle to
-        // this database file shares.
-        _dbPool->getBase().registerAfterCommitCallback(move(callback));
-    }
-}
-
 void BedrockServer::sync()
 {
     // Parse out the number of worker threads we'll use. The DB needs to know this because it will expect a
@@ -120,18 +105,18 @@ void BedrockServer::sync()
     // We use fewer FDs on test machines that have other resource restrictions in place.
 
     SQLite::journalZstdDictionaryID = args.calc("-journalZstdDictionaryID");
-    SINFO("Setting dbPool size to: " << _dbPoolSize);
-    _dbPool = make_shared<SQLitePool>(_dbPoolSize, args["-db"], args.calc("-cacheSize"), args.calc("-maxJournalSize"), journalTables, mmapSizeGB, args.isSet("-newDBsUseHctree"), args["-checkpointMode"]);
-    SQLite& db = _dbPool->getBase();
-
-    // Hand over any callbacks that plugins registered before this pool existed
-    {
-        lock_guard<decltype(_afterCommitCallbackMutex)> lock(_afterCommitCallbackMutex);
-        for (const auto& callback : _afterCommitCallbacks) {
-            function<void()> callbackCopy = callback;
-            db.registerAfterCommitCallback(move(callbackCopy));
-        }
+    // Every plugin gets one of these. The base implementation does nothing, so plugins that don't care about commits
+    // cost an empty virtual call each, which is nothing next to the commit that just happened.
+    vector<function<void()>> callbacks;
+    for (const auto& plugin : plugins) {
+        callbacks.emplace_back([pluginPtr = plugin.second]() {
+            pluginPtr->afterCommitCallback();
+        });
     }
+
+    SINFO("Setting dbPool size to: " << _dbPoolSize);
+    _dbPool = make_shared<SQLitePool>(_dbPoolSize, args["-db"], args.calc("-cacheSize"), args.calc("-maxJournalSize"), journalTables, mmapSizeGB, args.isSet("-newDBsUseHctree"), args["-checkpointMode"], callbacks);
+    SQLite& db = _dbPool->getBase();
 
     // Allow plugins to read from the DB at startup.
     for (auto plugin : plugins) {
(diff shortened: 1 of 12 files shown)


sqlitecluster/SQLite.h
```diff
@@ -220,6 +220,14 @@ class SQLite {
     // The main purpose of this is to allow replications in SQLiteNode to notify other waiting threads that the commit has finished even before the checkpoint is done.
     int commit(const string& description = "UNSPECIFIED", const string& commandName = "", function<void()>* preCheckpointCallback = nullptr);
 
+    // Registers a callback to run after every successful commit to this database, whether the commit came from a
+    // command running on the leader or from replication on a follower. Callbacks are shared across every handle to
+    // the same database file, so one registration covers all of them.
+    // The callback runs after the commit lock is released, but it still runs on the committing thread: it must not
+    // block, and it must not touch the database.
```
danieldoglas: This is explicitly this way because we will just use the after commit callback to wake up our local thread in Auth

iwiznia (dismissed):

sqlitecluster/SQLite.cpp
```diff
@@ -1673,3 +1677,30 @@ map<uint64_t, pair<string, string>> SQLite::SharedData::popCommittedTransactions
...
+    _afterCommitCallbacks.push_back(move(callback));
+}
+
+void SQLite::SharedData::runAfterCommitCallbacks()
+{
+    shared_lock<decltype(_afterCommitCallbackLock)> lock(_afterCommitCallbackLock);
+    for (const auto& callback : _afterCommitCallbacks) {
+        // The commit already happened, and SQLiteCore::commit is noexcept, so letting an exception out of here would
+        // terminate the process over a callback that has no bearing on whether the commit succeeded.
+        try {
```
tylerkaraszewski, line 1686: Instead of noting that a potential caller of this function is `noexcept` in a comment here, drop the comment and make this function `noexcept` instead. This makes it clear why we have the try/catch, and given that there's no real recourse to exceptions caught here, logging them to later allow callbacks to be updated not to throw them seems the right call.

danieldoglas: done!

tylerkaraszewski (changes requested):
So I have an entirely different interface for this in mind that might come out simpler.

What if we add:

virtual void BedrockPlugin::afterCommitCallback();

We change SQLite and SQLitePool to take a list of afterCommit callbacks.

SQLitepool does nothing with that except pass it to the constructor for `_baseDB`.

SQLite's `_afterCommitCallbacks` is `const` and set at construction.

The only change in BedrockServer then is that in `sync()` we change this line:

```
_dbPool = make_shared<SQLitePool>(_dbPoolSize, args["-db"], args.calc("-cacheSize"), args.calc("-maxJournalSize"), journalTables, mmapSizeGB, args.isSet("-newDBsUseHctree"), args["-checkpointMode"]);
```

To something like:
```
vector<function<void()>> callbacks;
for (auto& plugin : plugins) {
    if (plugin.afterCommitCallback) {
        callbacks.emplace_back(plugin.afterCommitCallback);
    }
}
_dbPool = make_shared<SQLitePool>(_dbPoolSize, args["-db"], args.calc("-cacheSize"), args.calc("-maxJournalSize"), journalTables, mmapSizeGB, args.isSet("-newDBsUseHctree"), args["-checkpointMode"], callbacks);
```

The interface to BedrockSever doesn't need to change, we don't need a mutex, we don't need to worry about lifetimes, etc.

BedrockServer.cpp
```diff
@@ -105,8 +105,14 @@ void BedrockServer::sync()
     // We use fewer FDs on test machines that have other resource restrictions in place.
 
     SQLite::journalZstdDictionaryID = args.calc("-journalZstdDictionaryID");
+    vector<function<void()>> callbacks;
+    for (const auto& plugin : plugins) {
+        callbacks.emplace_back([pluginPtr = plugin.second]() {
+            pluginPtr->afterCommitCallback();
+        });
+    }
```
danieldoglas, line 113: @tylerkaraszewski your code suggestion was

```
vector<function<void()>> callbacks;
for (auto& plugin : plugins) {
    if (plugin.afterCommitCallback) {
        callbacks.emplace_back(plugin.afterCommitCallback);
    }
}
_dbPool = make_shared<SQLitePool>(_dbPoolSize, args["-db"], args.calc("-cacheSize"), args.calc("-maxJournalSize"), journalTables, mmapSizeGB, args.isSet("-newDBsUseHctree"), args["-checkpointMode"], callbacks);
```

This is a bit different because we can't pass a member function straight to that vector, and I don't think we can check if `plugin.afterCommitCallback` exists like in that code suggestion.

What we could do is implement something like `shouldEnableOnPrepareNotification`.

Or we can just leave it as is and it will run nothing for the plugins that don't implement it.

iwiznia (dismissed):

BedrockServer.cpp
tylerkaraszewski: What if we did this:

add:
```
const function<void()> BedrockPlugin::afterCommitCallback;
```

The default `function` is falsy.

Add a second parameter to the BedrockPlugin constructor:
```
BedrockPlugin::BedrockPlugin(BedrockServer& s, function<void()> afterCommitCallback = {}) : server(s), afterCommitCallback(afterCommitCallback)
{
}
```

And then here we *can* do the `emplace_back` like:
```
if (plugin.second->afterCommitCallback) {
    callbacks.emplace_back(plugin.second->.afterCommitCallback);
}
```

I think this is roughly as clean to implement and saves us the extra function calls (both for the plugins that don't implement this, and the lambda wrapper we currently are using). They aren't a huge cost to run, but we do want to keep this path fast.

tylerkaraszewski (changes requested):

BedrockServer.cpp
danieldoglas: Done

tylerkaraszewski (approved):

PR #2734, opened 2026-08-20
Add different error messages for user vs command triggered 503

cead22:
### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/673310

### Tests
Auto
_________

### DIFF
--- a/BedrockBlockingCommandQueue.cpp
+++ b/BedrockBlockingCommandQueue.cpp
@@ -23,14 +23,14 @@ uint64_t BedrockBlockingCommandQueue::_now() const
 
 void BedrockBlockingCommandQueue::push(unique_ptr<BedrockCommand>&& command)
 {
-    const char* dimension = _getBlockingDimension(command->blockingQueueRateLimitIdentifier, command->request.methodLine);
-    if (dimension) {
+    const string dimension = _getBlockingDimension(command->blockingQueueRateLimitIdentifier, command->request.methodLine);
+    if (!dimension.empty()) {
         SINFO("Blocking queue rate limited command rejected", {
             {"dimension", dimension},
             {"identifier", command->blockingQueueRateLimitIdentifier},
             {"command", command->request.methodLine}
         });
-        STHROW("503 Blocking queue rate limited (" + string(dimension) + ")");
+        STHROW("503 Blocking queue rate limited (" + dimension + ")");
     }
 
     // Base class acquires its own `_queueMutex`.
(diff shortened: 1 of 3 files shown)


BedrockBlockingCommandQueue.cpp
```diff
@@ -23,8 +23,14 @@ uint64_t BedrockBlockingCommandQueue::_now() const
 
 void BedrockBlockingCommandQueue::push(unique_ptr<BedrockCommand>&& command)
 {
-    if (isBlocked(command->blockingQueueRateLimitIdentifier, command->request.methodLine)) {
-        STHROW("503 Blocking queue rate limited (time)");
+    const char* dimension = _getBlockingDimension(command->blockingQueueRateLimitIdentifier, command->request.methodLine);
```
danieldoglas: any reason why were using char* instead of string here?

danieldoglas: (and in the other instances)

danieldoglas (dismissed):
LGTM, I'd use string in those methods but I'm not sure we care so much

cead22:
Good catch, updated

BedrockBlockingCommandQueue.cpp
```diff
@@ -129,11 +141,22 @@ void BedrockBlockingCommandQueue::recordExecutionTime(const string& identifier,
 }
 
 bool BedrockBlockingCommandQueue::isBlocked(const string& identifier, const string& commandName)
+{
+    return !_getBlockingDimension(identifier, commandName).empty();
+}
```
danieldoglas: This is now only used in the automated tests. Should we make `_getBlockingDimension` public and use that instead in the tests too?

BedrockBlockingCommandQueue.h
```diff
@@ -39,10 +39,8 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
     // sample against the identifier (when `identifier` is non-empty) and against the command name.
     void recordExecutionTime(const string& identifier, const string& commandName, uint64_t elapsedUS);
 
-    // True if `identifier` or `commandName` is over its blocking-queue time limit within the window, or is
-    // still inside an active block. When a dimension newly trips, it is blocked for the block duration.
-    // Logs a "Blocking queue rate limit" line when a dimension blocks.
-    bool isBlocked(const string& identifier, const string& commandName);
+    // Return the active rate-limit dimension, or an empty string. Identifier blocks take precedence when both dimensions are active.
+    string _getBlockingDimension(const string& identifier, const string& commandName);
```
danieldoglas: usually we use `_` to say a function or variable is private

```suggestion
    string getBlockingDimension(const string& identifier, const string& commandName);
```

danieldoglas (approved):

PR #2739, opened 2026-08-21
Allow async DNS resolution for TCPManager::Socket

tylerkaraszewski:
### Details

This creates a new `SResolve` function that returns a shared pointer to an `SResolution` object that just create a new thread to do DNS resolution in. This object is `poll()`able because it has a pipe that we can use to check when it's done. The normal `poll()` loop we use everywhere else has been extended to support a `RESOLVING` state on sockets.

New sockets will block on DNS resolutions for up to 5ms, and then become async, so any addresses already in the machine's local DNS cache should still resolve and create a socket immediately, without waiting for the first `poll()` loop.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/673733

### Tests

New tests added.

Tested with Auth, FuzzyBot, and ExpensifyBackupManager with no issues:
Screenshot 2026-08-25 at 16 31 16

_________


libstuff/STCPManager.cpp
```diff
@@ -184,7 +205,19 @@ void STCPManager::postPoll(fd_map& fdm, Socket& socket)
...
+    // a state that prePoll would try to register a missing fd for.
+    if (state.load() == State::RESOLVING) {
+        state.store(State::CLOSED);
+        return;
+    }
+
+    // There may still be no fd: opening one can fail after the address is known.
+    if (s > 0) {
+        ::shutdown(s, SHUT_RDWR);
+    }
```
danieldoglas, line 220: I'm not sure I understand what's going on here. 

>There's no fd to shut down

Don't we start the FD as soon as we initialize `Socket` by doing `dnsResolution(_startResolution(host))`?

danieldoglas: ohhhh nvm, we only create the socket itself when we call `_connectAfterDNSResolution`. So the fd described above is not related to the Resolution pipe, but to the socked itself

libstuff/STCPManager.cpp
```diff
@@ -194,21 +227,69 @@ STCPManager::Socket::Socket(int sock, STCPManager::Socket::State state_, bool ht
...
+    s = S_socket(addr, true, false, false);
+    if (s < 0) {
+        state.store(State::CLOSED);
+        connectFailure = true;
+        STHROW("Couldn't open socket to " + SToStr(addr));
+    }
+
+    if (https) {
+        ssl = new SSSLState(hostname, s);
+    }
```
danieldoglas, line 277: Do we need to set the state as CONNECTING here?

libstuff/STCPManager.cpp
```diff
@@ -194,21 +227,69 @@ STCPManager::Socket::Socket(int sock, STCPManager::Socket::State state_, bool ht
...
     if (s < 0) {
-        STHROW("Couldn't open socket to " + host);
+        state.store(State::CLOSED);
+        connectFailure = true;
+        return;
     }
+
+    if (https) {
+        ssl = new SSSLState(hostToResolve, s);
+    }
```
danieldoglas: this is duplicated code from 260-268, let's DRY it

libstuff/STCPManager.h
```diff
@@ -58,6 +76,13 @@ struct STCPManager
         SFastBuffer sendBuffer;
 
         bool https;
+
+        // Will be null only if this socket was initialized from an address or an existing fd. A literal
+        // IP string still gets one, already resolved.
+        // Shared ownership allows the resolver thread to continue even if the socket is destroyed during
+        // a slow DNS lookup.
+        const shared_ptr<SResolution> dnsResolution;
```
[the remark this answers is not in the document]
danieldoglas: Not sure if this is needed. we could indeed release the reader fd, but we'll kill the socket after 10s if I remember correctly, so maybe that's not a problem?

libstuff/STCPManager.cpp
tylerkaraszewski: It's set on line 255.

libstuff/STCPManager.h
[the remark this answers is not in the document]
tylerkaraszewski: I'm not worried about an extra FD for the rest of the lifetime of the command.

danieldoglas (approved):

PR #2744, opened 2026-08-25
Add reusable compiled regexes to libstuff

rlinoz:
### Details

coming from https://github.com/Expensify/Auth/pull/23856#pullrequestreview-4997550522

Add an owning `SRECompiledRegex` type and `SRECompile` so callers can compile fixed regular expressions once and reuse them. Add overloads of `SREMatch`, `SREMatchAll`, and `SREReplace` that accept compiled expressions while preserving the existing string-based interfaces.

This allows Auth's comment HTML conversion to reuse static compiled expressions instead of either compiling them for every comment or adding heuristic string searches before each regex operation.

### Fixed Issues
For https://github.com/Expensify/Auth/pull/23856

### Tests

1. Build Auth against the modified Bedrock, verify reportactions are displayed correctly
2. Build https://github.com/Expensify/Auth/pull/23856 against the modified Bedrock, verify reportactions are displayed correctly

_________

### DIFF
--- a/libstuff/libstuff.cpp
+++ b/libstuff/libstuff.cpp
@@ -3216,25 +3216,18 @@ bool SIsValidSQLiteDateModifier(const string& modifier)
 
 bool SREMatch(const string& regExp, const string& input, bool caseSensitive, bool partialMatch, vector<string>* matches, size_t startOffset, size_t* matchOffset)
 {
-    int errornumber = 0;
-    PCRE2_SIZE erroroffset = 0;
-    uint32_t matchFlags = 0;
+    return SREMatch(SRECompile(regExp, caseSensitive), input, partialMatch, matches, startOffset, matchOffset);
+}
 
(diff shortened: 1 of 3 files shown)


tylerkaraszewski (commented):
Looks OK, but not approving just yet - do you have new benchmarks with this version compared to the original string scanning one? I want to see how close this gets to what we had before. I do think this is an improvement, though.

rlinoz:
It seems to be slightly slower than the guards in the common case of not having html, but is still reasonable and the code looks much better. Maybe we add just a first guard like `SContains('<')`?

| | main | guards | precompiled | vs main | vs guards |
|---|---|---|---|---|---|
| plain | 46.08 ms | 37.40 ms | 38.41 ms | -16.6% | +2.7% |
| mixed | 47.48 ms | 39.14 ms | 39.24 ms | -17.4% | +0.3% |
| markup | 49.79 ms | 41.75 ms | 40.81 ms | -18.0% | -2.2% |

tylerkaraszewski:
I think it's fine as it it, I don't really feel like one guard is worth it for the 2% performance benefit.

tylerkaraszewski (approved):

PR #2747, opened 2026-08-26
Capture test server shutdown state

MonilBhavsar:
### Details
When the Auth test client times out while both Bedrock listeners refuse connections, capture whether its child process is still running, has exited, or was terminated by a signal. Record the graceful-shutdown reason when that path is used.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/653978

### Tests
1. Run `make -j16` in Bedrock.
2. Run `make -j16` in Auth against the rebuilt Bedrock library.
3. Run `./authtest -threads 32 -logLevel INFO -enableHctree -only RevertSplitTransaction` from `Auth/test`.
4. Confirm the fixture completes without changing normal server lifecycle behavior.

_________

### DIFF
--- a/test/lib/BedrockTester.cpp
+++ b/test/lib/BedrockTester.cpp
@@ -72,6 +72,7 @@ BedrockTester::BedrockTester(const map<string, string>& args,
     }
 
     string dbFileName = getTempFileName();
+    string shutdownReasonFileName = getTempFileName("shutdown_reason_");
     map<string, string> defaultArgs = {
         {"-db", dbFileName},
         {"-serverHost", "127.0.0.1:" + to_string(_serverPort)},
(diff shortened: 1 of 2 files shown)


test/lib/BedrockTester.cpp
```diff
@@ -416,8 +420,23 @@ vector<SData> BedrockTester::executeWaitMultipleData(vector<SData> requests, int
                                 *errorCode = 2;
                             }
                             if (timeout) {
+                                SAUTOLOCK(listLock);
```
tylerkaraszewski: This is basically deprecated, can you switch to `lock_guard`

tylerkaraszewski (changes requested):

tylerkaraszewski (approved):

PR #2750, opened 2026-08-28
Acquire commitLock on writeLocalUnreplicated

danieldoglas:
### Details

Acquires the commit lock on `writeLocalUnreplicated` so we don't receive SQLITE_BUSY while commiting.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/671183

### Tests

N/A

_________

### DIFF
--- a/sqlitecluster/SQLite.cpp
+++ b/sqlitecluster/SQLite.cpp
@@ -831,21 +831,26 @@ bool SQLite::writeLocalUnreplicated(const string& query)
     // Held in shared mode for the same reason the journal trim in `prepare` holds it: BlockWrites needs to be able to
     // stop this.
     shared_lock<shared_mutex> lock(_sharedData.writeLock);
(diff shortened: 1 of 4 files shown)


iwiznia (approved):

sqlitecluster/SQLite.cpp
```diff
@@ -841,8 +840,17 @@ bool SQLite::writeLocalUnreplicated(const string& query)
         return false;
     }
 
-    // A commit that landed while we were writing shows up here as SQLITE_BUSY_SNAPSHOT.
-    if (SQuery(_db, "COMMIT")) {
+    bool commitFailed = false;
```
tylerkaraszewski, line 843: I would just leave this out and throw the `rollback` inside the lock scope for simplicity.

danieldoglas: My concern here is that rollback might be slow, and we'll hold the lock for the duraction of it. 

Even though we wrote this code for a specific use that shoudl be fast, there's no way to guarantee that everyone using bedrock will use it in that intended way, so I think I prefer leaving that one out.

PR #2754, opened 2026-08-31
Initial commit of Bedrock Wire Protocol (BWP) Specification Doc

fukawi2:
Initial documentation of the Bedrock Wire Protocol specification.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -1,0 +1,1811 @@
+---
+layout: default
+---
+
+# The Bedrock Wire Protocol (BWP)
+
+**Status:** Informational. Describes existing behaviour.
+
+**Verified against:** commit `b1e7fcf40050d29aea40f38409d6e559ef970f90` (2026-08-06).
+
+## Abstract
+
+Bedrock speaks a text protocol that resembles HTTP/1.1 but is not HTTP/1.1. This
+document specifies the message format, framing, header conventions, status
+codes, and connection semantics, for three uses of the same envelope:
+
+1. The client protocol;
+2. Command escalation between nodes; and
+3. Cluster replication.
+
+This document describes the protocol as deployed. Section 15 lists behaviour
+that appears to be defective; that behaviour is documented but is not normative,
+and implementations should not rely on it.
+
+---
+
+## The Protocol in Brief
+
+Everything needed to write a working client. The rest of the document is
+reference detail.
+
+A message is a method line, some `name: value` fields, a blank line, and an
+optional body. A request:
+
+```
+Status
+Content-Length: 0
+
+```
+
+A response has the same shape, with a status in place of the method line:
+
+```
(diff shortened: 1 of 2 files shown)


nathanmetcalf:
I've read this document and can understand it, but I do not attest to its accuracy. Please get a further review before merging.

nathanmetcalf (approved):

fukawi2:
@tylerkaraszewski Could I get your eyes/thoughts on this please? :pray:

docs/protocol.md
```diff
@@ -0,0 +1,1811 @@
...
+document specifies the message format, framing, header conventions, status
+codes, and connection semantics, for three uses of the same envelope:
+
+1. The client protocol;
+2. Command escalation between nodes; and
+3. Cluster replication.
+
+This document describes the protocol as deployed. Section 15 lists behaviour
+that appears to be defective; that behaviour is documented but is not normative,
+and implementations should not rely on it.
```
tylerkaraszewski, line 23: Weird statement, as if we're writing a spec for a standard that multiple implementers will build.

docs/protocol.md
```diff
@@ -0,0 +1,1811 @@
...
+```
+
+Six rules cover almost all of it:
+
+1. **End every line with CRLF.** A receiver also accepts a bare LF, which is
+   what typing into `netcat` produces.
+2. **A blank line ends the fields.** Everything after it is the body.
+3. **`Content-Length` gives the body length in octets.** Send it, and make it
+   exact, on any connection you will reuse. See section 4.4; this is the rule
+   most often got wrong.
```
tylerkaraszewski, line 59: List examples where this was often gotten wrong, please.

If I were making this an actual spec, I would just make Content-Length mandatory, even though we accept requests that lack it (which is useful for handwritten commands and such)

docs/protocol.md
```diff
@@ -0,0 +1,1811 @@
...
+4. **Field names are case-insensitive and must not contain a colon.** Field
+   values are backslash-escaped, so a value can carry a newline as `\n` and
+   therefore a whole JSON document on one line.
+5. **The status is the leading integer of the response's first line.** `2xx` is
+   success. The words after the integer are not decorative; see section 8.4.
+6. **Connections are persistent.** Send `Connection: close` when finished, or
+   the server waits for another request.
+
+Two things that surprise newcomers:
```
tylerkaraszewski, line 68: Who was surprised? Come on, did you actually read this? I question the value of using AI to generate a document that nobody will read and will quickly become out-of-date over just asking AI to freshly answer a specific question about the code based on the current state of the code when the question is asked.

And if we want this for some sort of compliance reason get the AI to write it as an RFC type document without the editorializing.

I am stopping with the review for now.

tylerkaraszewski (changes requested):

PR #2769, opened 2026-09-04
Auto-reject blocking queue commands when the thread saturates

danieldoglas:
### Details

This PR adds a new global blocking queue limit. If the blockgin queue was used for 55s from the past 60s, it will blockg everything in the next 60s.

Those numbers are all configurable by control commands.

It also refactors a bit of code so we can reuse all the code between all 3 dimentions (identifier, command and global).

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/673587

### Tests

New automated tests.

_________

Compiled `auth/Auth.cpp`, `auth/command/OpenApp.cpp` and `auth/lib/OnyxDeleter.cpp` (the Auth translation units that include `BedrockServer.h`, and therefore this header) against this branch. All clean.

### DIFF
--- a/BedrockBlockingCommandQueue.cpp
+++ b/BedrockBlockingCommandQueue.cpp
@@ -139,8 +139,18 @@ uint64_t BedrockBlockingCommandQueue::setBlockDuration(const uint64_t durationUS
 void BedrockBlockingCommandQueue::recordExecutionTime(const string& identifier, const string& commandName, uint64_t elapsedUS)
 {
     const uint64_t now = _now();
-    _recordAndCheck(_identifierStates, identifier, _identifierThresholdUS.load(), now, elapsedUS, "identifier");
-    _recordAndCheck(_commandStates, commandName, _commandThresholdUS.load(), now, elapsedUS, "command");
+    const uint64_t windowUS = _windowUS.load();
+    const uint64_t blockDurationUS = _blockDurationUS.load();
+
+    const Limits identifierLimits = {windowUS, _identifierThresholdUS.load(), blockDurationUS, LOG_THRESHOLD_US};
+    if (!identifier.empty() && identifierLimits.thresholdUS) {
+        _recordAndCheck(*_getOrCreateState(_identifierStates, identifier), "identifier", identifier, identifierLimits, now, elapsedUS);
+    }
+
+    const Limits commandLimits = {windowUS, _commandThresholdUS.load(), blockDurationUS, LOG_THRESHOLD_US};
+    if (!commandName.empty() && commandLimits.thresholdUS) {
+        _recordAndCheck(*_getOrCreateState(_commandStates, commandName), "command", commandName, commandLimits, now, elapsedUS);
+    }
 }
 
 string BedrockBlockingCommandQueue::getBlockingDimension(const string& identifier, const string& commandName)
@@ -148,11 +158,17 @@ string BedrockBlockingCommandQueue::getBlockingDimension(const string& identifie
     // Hot path: called by push() and by _dequeue() under the base `_queueMutex`. Keep it O(1) by reading only
     // the precomputed block deadline. The windowed time is summed in recordExecutionTime, off the blocking thread.
     const uint64_t now = _now();
-    if (_isBlocked(_identifierStates, identifier, now)) {
-        return "identifier";
+    if (!identifier.empty()) {
+        auto state = _getState(_identifierStates, identifier);
+        if (state && _isBlocked(*state, now)) {
+            return "identifier";
+        }
     }
-    if (_isBlocked(_commandStates, commandName, now)) {
-        return "command";
+    if (!commandName.empty()) {
+        auto state = _getState(_commandStates, commandName);
+        if (state && _isBlocked(*state, now)) {
+            return "command";
+        }
     }
     return "";
 }
@@ -174,70 +190,56 @@ shared_ptr<BedrockBlockingCommandQueue::DimensionState> BedrockBlockingCommandQu
     return it == map.states.end() ? nullptr : it->second;
 }
 
-void BedrockBlockingCommandQueue::_recordAndCheck(StateMap& map, const string& key, uint64_t thresholdUS, uint64_t now, uint64_t elapsedUS, const string& dimension)
+void BedrockBlockingCommandQueue::_recordAndCheck(DimensionState& state, const string& dimension, const string& key, const Limits& limits, uint64_t now, uint64_t elapsedUS)
 {
-    if (key.empty() || thresholdUS == 0) {
-        return;
-    }
-
-    auto state = _getOrCreateState(map, key);
-    lock_guard<decltype(state->m)> lock(state->m);
+    lock_guard<decltype(state.m)> lock(state.m);
 
     // Already inside an active block: a block lasts a fixed duration and is not extended by further hits, so
     // there's nothing to record or recompute until it expires.
-    if (state->blockedUntil > now) {
+    if (state.blockedUntil > now) {
         return;
     }
 
-    state->commands.push_back({now, elapsedUS});
-
-    const uint64_t windowUS = _windowUS.load();
+    state.commands.push_back({now, elapsedUS});
 
     // Drop samples that finished before the current window and sum the ones that are still valid.
     uint64_t total = 0;
-    for (auto it = state->commands.begin(); it != state->commands.end();) {
+    for (auto it = state.commands.begin(); it != state.commands.end();) {
         const uint64_t timeSinceFinishedUS = now > it->finishTime ? now - it->finishTime : 0;
-        if (timeSinceFinishedUS >= windowUS) {
-            it = state->commands.erase(it);
+        if (timeSinceFinishedUS >= limits.windowUS) {
+            it = state.commands.erase(it);
             continue;
         }
 
         // Doing the line below will allow us to count for partial timings spent in a command. Let's say the window is 100s,
         // and we had a command that took 50s and finished 80s ago. We should only count 20s of that command, since the other
         // 30s are outside the window. So the result would be min(50s, 100 - 80) = min(50, 20) = 20s.
-        total += min(it->elapsedTime, windowUS - timeSinceFinishedUS);
+        total += min(it->elapsedTime, limits.windowUS - timeSinceFinishedUS);
         ++it;
     }
 
-    if (total > thresholdUS) {
-        state->blockedUntil = now + _blockDurationUS.load();
+    if (total > limits.thresholdUS) {
+        state.blockedUntil = now + limits.blockDurationUS;
         SINFO("Blocking queue rate limit reached, blocking", {
             {"dimension", dimension},
             {"identifier", key},
             {"timeMS", to_string(total / 1000)},
-            {"thresholdMS", to_string(thresholdUS / 1000)},
-            {"blockDurationMS", to_string(_blockDurationUS.load() / 1000)}
+            {"thresholdMS", to_string(limits.thresholdUS / 1000)},
+            {"blockDurationMS", to_string(limits.blockDurationUS / 1000)}
         });
-    } else if (total > LOG_THRESHOLD_US) {
+    } else if (limits.logThresholdUS && total > limits.logThresholdUS) {
         // Log-only monitoring: the identifier is heavy but still under its block threshold.
         SINFO("Blocking queue rate limit above logging threshold", {
             {"dimension", dimension},
             {"identifier", key},
             {"timeMS", to_string(total / 1000)},
-            {"thresholdMS", to_string(LOG_THRESHOLD_US / 1000)}
+            {"thresholdMS", to_string(limits.logThresholdUS / 1000)}
         });
     }
 }
 
-bool BedrockBlockingCommandQueue::_isBlocked(StateMap& map, const string& key, uint64_t now)
+bool BedrockBlockingCommandQueue::_isBlocked(DimensionState& state, uint64_t now)
 {
-    if (key.empty()) {
-        return false;
-    }
-    auto state = _getState(map, key);
-    if (!state) {
-        return false;
-    }
-    lock_guard<decltype(state->m)> lock(state->m);
-    return state->blockedUntil > now;
+    lock_guard<decltype(state.m)> lock(state.m);
+    return state.blockedUntil > now;
 }
--- a/BedrockBlockingCommandQueue.cpp
+++ b/BedrockBlockingCommandQueue.cpp
@@ -157,17 +157,17 @@ string BedrockBlockingCommandQueue::getBlockingDimension(const string& identifie
     return "";
 }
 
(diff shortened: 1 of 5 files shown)


BedrockBlockingCommandQueue.cpp
```diff
@@ -105,100 +111,139 @@ STable BedrockBlockingCommandQueue::getState()
...
+        _recordAndCheck(*_getOrCreateState(_identifierStates, identifier), "identifier", identifier, _identifierLimits, now, elapsedUS);
+    }
+    if (!commandName.empty() && _commandLimits.thresholdUS.load()) {
+        _recordAndCheck(*_getOrCreateState(_commandStates, commandName), "command", commandName, _commandLimits, now, elapsedUS);
+    }
+
+    // Every command counts toward the global dimension, whoever sent it.
+    if (_globalLimits.thresholdUS.load()) {
+        _recordAndCheck(_globalState, "global", "", _globalLimits, now, elapsedUS);
+    }
```
danieldoglas, line 181: bot comment here: https://github.com/Expensify/Bedrock/pull/2769#discussion_r3936672562


P1 Badge Avoid rescanning the global window for every command

Because _globalState receives every blocking-queue command, this call makes _recordAndCheck traverse every sample retained in the 60-second window after each execution. recordExecutionTime runs synchronously on worker 0 (BedrockServer.cpp:455-456), so N fast commands incur Θ(N²) work on the thread this limiter is intended to protect; moreover, that bookkeeping occurs after the elapsed-time measurement and therefore cannot trigger the limit itself. A burst of short conflicting commands can consequently stall the blocking worker and grow its backlog without reaching the configured 55-second threshold. Maintain an incremental total or otherwise make global-window updates amortized O(1).

Useful? React with 👍 / 👎.

danieldoglas: I don't think we care about this. Sure, it will still run in the blocking queue, but the commit lock will be free already. If anything, this gives the cluster some more time to process other commands while we process it.

I also don't expect us to have thousands and thousands of commands here all the time

danieldoglas: cc @tylerkaraszewski, not sure if you disagree.

BedrockBlockingCommandQueue.cpp
```diff
@@ -78,6 +78,12 @@ size_t BedrockBlockingCommandQueue::clearRateLimits()
         size += _commandStates.states.size();
         _commandStates.states.clear();
     }
+    {
```
iwiznia, line 81: Do we need this?

iwiznia (dismissed):

BedrockBlockingCommandQueue.cpp
danieldoglas: I mean, if we're clearing the rate limits, we clear them all, right?

flodnv (dismissed):
🙏

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -39,9 +40,16 @@ struct BlockingQueueRateLimitTest : tpunit::TestFixture
...
+        // The global dimension is on by default. Disable it so it can't reject anything in the tests that
+        // aren't about it; testGlobalRateLimiting turns it back on.
+        SData disableGlobal("SetBlockingQueueTimeRateLimit");
+        disableGlobal["globalThresholdMS"] = "0";
+        leader.executeWaitVerifyContent(disableGlobal, "200", true);
+
         SData status("Status");
         STable json = SParseJSONObject(leader.executeWaitVerifyContent(status, "200", true));
         ASSERT_EQUAL(json["blockingBlockedIdentifiers"], "");
+        ASSERT_EQUAL(json["blockingGlobalBlocked"], "false");
```
cead22:
```suggestion
        ASSERT_EQUAL(json["globalRateLimiterTriggered"], "false");
```

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -54,6 +62,9 @@ struct BlockingQueueRateLimitTest : tpunit::TestFixture
         setLimits["identifierThresholdMS"] = "20000";
         setLimits["commandThresholdMS"] = "40000";
         setLimits["blockDurationMS"] = "60000";
+        setLimits["globalWindowMS"] = "60000";
+        setLimits["globalThresholdMS"] = "55000";
+        setLimits["globalBlockDurationMS"] = "60000";
```
cead22: Let's use consistent naming. I suggest global rate limiter instead of global / global block, since we're using rate limiter in other contexts already
```suggestion
        setLimits["globalRateLimiterWindowMS"] = "60000";
        setLimits["globalRateLimiterThresholdMS"] = "55000";
        setLimits["globalRateLimiterCooldownMS"] = "60000";
```

test/clustertest/tests/BlockingQueueRateLimitTest.cpp
```diff
@@ -39,9 +40,16 @@ struct BlockingQueueRateLimitTest : tpunit::TestFixture
         resetConflict["MaxConflictRetries"] = "3";
         leader.executeWaitVerifyContent(resetConflict, "200", true);
 
+        // The global dimension is on by default. Disable it so it can't reject anything in the tests that
+        // aren't about it; testGlobalRateLimiting turns it back on.
```
cead22: De-AI all comments
```suggestion
        // The global blocker is on by default, so we disable for tests unrelated to it.
```

test/tests/BlockingCommandQueueTest.cpp
```diff
@@ -310,4 +317,136 @@ struct BlockingCommandQueueTest : tpunit::TestFixture
         queue.clearRateLimits();
         ASSERT_EQUAL(queue.getBlockingDimension("acct1", "cmd"), "");
     }
+
+    void testGlobalOverThresholdBlocksEveryone()
```
cead22:
```suggestion
    void testGlobalRateLimiterOverThresholdBlocksEveryone()
```

test/tests/BlockingCommandQueueTest.cpp
```diff
@@ -310,4 +317,136 @@ struct BlockingCommandQueueTest : tpunit::TestFixture
...
+    {
+        // The global dimension counts every command, so a burst spread over several identifiers trips it even
+        // though no single identifier or command is over a limit of its own.
```
cead22:
```suggestion
        // The global rate limiter counts every command, so a burst of commands with different identifiers trips it even
        // though no single identifier or command is over a limit of its own.
```

test/tests/BlockingCommandQueueTest.cpp
```diff
@@ -310,4 +317,136 @@ struct BlockingCommandQueueTest : tpunit::TestFixture
...
+        queue.recordExecutionTime("acct1", "cmd1", 30);
+        ASSERT_EQUAL(queue.getBlockingDimension("acct1", "cmd1"), "");
+
+        queue.recordExecutionTime("acct2", "cmd2", 30);
+        ASSERT_EQUAL(queue.getBlockingDimension("acct2", "cmd2"), "global");
+        ASSERT_EQUAL(queue.getBlockingDimension("acct3", "cmd3"), "global");
+        ASSERT_EQUAL(queue.getBlockingDimension("", ""), "global");
+    }
+
+    void testGlobalRejectsOnPushAndDequeue()
```
cead22:
```suggestion
    void testGlobalRateLimiterRejectsOnPushAndDequeue()
```

test/tests/BlockingCommandQueueTest.cpp
```diff
@@ -310,4 +317,136 @@ struct BlockingCommandQueueTest : tpunit::TestFixture
...
+        bool pushRejected = false;
+        try {
+            pushQueue.push(pushQueue.makeCommand("acct2", "otherCmd"));
+        } catch (const SException& e) {
+            pushRejected = true;
+            ASSERT_EQUAL(string(e.what()), "503 Blocking queue rate limited (global)");
+        }
+        ASSERT_TRUE(pushRejected);
+
+        // Commands that were already queued are rejected on the way out, so a backlog drains instead of running.
```
cead22: NAB since we want to be very conservative here to recover fast, but since the blocking queue won't be accepting new commands, should we consider processing those already queued?

If we do it this way, I think we can make the default cool down period shorter than 60s

test/tests/BlockingCommandQueueTest.cpp
```diff
@@ -289,6 +295,7 @@ struct BlockingCommandQueueTest : tpunit::TestFixture
         queue.setWindow(100);
         queue.setIdentifierThreshold(0);
```
cead22: I think these should match the params, eg setBlockingIdentifierThresholdMS for blockingIdentifierThresholdMS

test/tests/BlockingCommandQueueTest.cpp
```diff
@@ -310,4 +317,136 @@ struct BlockingCommandQueueTest : tpunit::TestFixture
...
+        queue.recordExecutionTime("acct1", "cmd", 60);
+        ASSERT_EQUAL(queue.getBlockingDimension("acct1", "cmd"), "global");
+    }
+
+    void testGlobalWindowIsIndependent()
+    {
+        // The global dimension has its own window, so a sample that has aged out of the shared window can still
+        // count toward the global threshold.
+        TestBlockingCommandQueue queue;
+        queue.setWindow(100);
```
cead22: This needs a more specific name, like setSharedRateLimiterWindow

test/tests/BlockingCommandQueueTest.cpp
```diff
@@ -310,4 +317,136 @@ struct BlockingCommandQueueTest : tpunit::TestFixture
...
+        queue.setCommandThreshold(0);
+        queue.setGlobalWindow(1000);
+        queue.setGlobalThreshold(50);
+        queue.setGlobalBlockDuration(1000);
+        queue.setNow(1000);
+
+        queue.recordExecutionTime("acct1", "cmd", 30);
+        ASSERT_EQUAL(queue.getBlockingDimension("acct1", "cmd"), "");
+
+        // 500us later the first sample is far outside the 100us shared window but inside the 1000us global
```
cead22: 500us or 1500us? Also I think using ms instead of us here is less confusing

BedrockBlockingCommandQueue.h
```diff
@@ -35,15 +36,24 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
     uint64_t setCommandThreshold(const uint64_t thresholdUS);
     uint64_t setBlockDuration(const uint64_t durationUS);
 
+    // The global dimension counts every command that runs on the blocking thread, so that a burst spread over
+    // many identifiers still trips it. It gets its own window, threshold and block duration because it
+    // measures total saturation of the thread rather than one identifier's share of it.
```
cead22:
```suggestion
    // The global rate limiter counts every command that runs on the blocking thread, so that a burst of commands with different
    // identifiers still trips it. It gets its own window, threshold and block duration because it
    // measures total saturation of the thread rather than one identifier's share of it.
```

BedrockBlockingCommandQueue.h
```diff
@@ -78,34 +88,57 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
...
+        unordered_map<string, shared_ptr<DimensionState>> states;
+    };
+
+    // The tunables for one dimension, in microseconds. The blocking thread reads these while the
+    // SetBlockingQueueTimeRateLimit control command writes them from a worker, so each one is atomic. A
+    // `thresholdUS` of 0 disables the dimension. A `logThresholdUS` of 0 disables the log-only line for
+    // identifiers that are heavy but under the threshold.
+    struct Limits
+    {
+        atomic<uint64_t> windowUS;
```
cead22, line 97: If we're only going to be setting these using MS, what's the use of dividing by 1000 and storing them in US?

cead22 (dismissed):
LGTM. All my comments are about using non-AI language to make this easier to understand

BedrockBlockingCommandQueue.h
danieldoglas: MS is easier for humans to understand. US is faster for us to process in the machine.

test/tests/BlockingCommandQueueTest.cpp
danieldoglas: I think we should block everything going forward. Mainly because we might have a 5k commands blocking queue here, which could still hold the lock for a long time.

I prefer rejecting everything and coming back as fast as possible.

BedrockBlockingCommandQueue.cpp
tylerkaraszewski: I think @iwiznia means the scoping, which doesn't really add anything in this particular case.

BedrockBlockingCommandQueue.cpp
tylerkaraszewski: I kinda care but don't think it's critical unless it starts to become an issue. I think this just adds numbers which is pretty fast, even in N^2.

BedrockBlockingCommandQueue.h
```diff
@@ -61,10 +70,9 @@ class BedrockBlockingCommandQueue : public BedrockCommandQueue {
     // An identifier's recently finished blocking-queue commands, oldest first.
     typedef deque<RecentlyFinishedCommand> RecentlyFinishedCommandList;
 
-    // Rate-limit state for one identifier (an identifier or a command name). Each entry has its own mutex, so
-    // different identifiers never contend on one lock. `blockedUntil` is the time (microseconds) an active
-    // block ends; 0 means not blocked.
-    struct IdentifierState
+    // Rate-limit state for one dimension. Each entry has its own mutex, so different identifiers never contend
+    // on one lock. `blockedUntil` is when an active block ends, in microseconds; 0 means not blocked.
+    struct DimensionState
```
tylerkaraszewski: This is such an LLM-y name. What's a DimensionState?

tylerkaraszewski (dismissed):

BedrockBlockingCommandQueue.h
danieldoglas: The name we had before was identifier, because this started with the identifier (which internally is accountID). Then we added `command` which was also an identifier. But now we have a global one, so it was renamed from Identifier to Dimension - we can block over an identifier, command or globally.

Do you have any other suggestion for naming here?

iwiznia (dismissed):

BedrockBlockingCommandQueue.h
tylerkaraszewski: `BlockingCategoryState`? I'm not sure, but `Dimension` means nothing to me here.

BedrockBlockingCommandQueue.h
cead22: Agreed. But why not store the vars in MS?

tylerkaraszewski (approved):

cead22 (approved):

PR #2770, opened 2026-09-04
Trim journal tables from a background deleter thread

danieldoglas:
### Details

Journal rows were deleted inside whatever command's transaction happened to trigger the cleanup. 

Because `id` is an INTEGER PRIMARY KEY, CONCURRENT mode records the read range of `DELETE FROM <journal> WHERE id < X` as everything below `X`, so two commands that pick the same journal table conflict with each other. A command that did a lot of real work then gets retried purely because it could not delete some old journal rows, and journal conflicts hit every customer rather than only the people touching one report.

This moves the deletion into a single background thread. 

`-journalDeleterBatchSize` (default 1000) sets how many rows a pass removes, and the `SetJournalDeleter` control command takes `enable` and `batchSize` so the thread can be turned off without a deploy.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/678886

### Tests

New automated tests.

_________

### DIFF
--- a/BedrockJournalDeleter.cpp
+++ b/BedrockJournalDeleter.cpp
@@ -1,0 +1,121 @@
+#include "BedrockJournalDeleter.h"
+
+#include <BedrockServer.h>
+#include <sqlitecluster/SQLitePool.h>
+
+BedrockJournalDeleter::BedrockJournalDeleter(BedrockServer& server) : _server(server)
+{
+}
+
+BedrockJournalDeleter::~BedrockJournalDeleter()
+{
+    stop();
+}
+
+void BedrockJournalDeleter::wake()
+{
+    {
+        lock_guard<decltype(_wakeMutex)> lock(_wakeMutex);
+        _wakeRequested = true;
+    }
+    _cv.notify_one();
+}
+
+bool BedrockJournalDeleter::isTrimmableState(SQLiteNodeState state)
+{
+    // We also commit while SYNCHRONIZING, but a node catching up drains what it accumulated once it is FOLLOWING.
+    return state == SQLiteNodeState::LEADING || state == SQLiteNodeState::FOLLOWING || state == SQLiteNodeState::STANDINGDOWN;
+}
+
+void BedrockJournalDeleter::start(SQLiteNodeState state)
+{
+    if (!enableDeleterThread || !isTrimmableState(state)) {
+        return;
+    }
+
+    lock_guard<decltype(_lifecycleMutex)> lifecycleLock(_lifecycleMutex);
+    if (_thread) {
+        return;
+    }
+
+    // Set before the thread exists, and any previous one was joined under _lifecycleMutex, so nothing reads this
+    // while we write it. A wake left over from while we were stopped just costs one trim.
+    _stop = false;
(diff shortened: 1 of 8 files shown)


main.cpp
```diff
@@ -330,6 +332,7 @@ int main(int argc, char* argv[])
     SETDEFAULT("-plugins", "db,jobs,cache,mysql,compression");
     SETDEFAULT("-priority", "100");
     SETDEFAULT("-maxJournalSize", "1000000");
+    SETDEFAULT("-journalDeleterBatchSize", "1000");
```
danieldoglas: We currently delete 10, but we were usually inserting one. With the new deleter thread, we will go through the journal tables one at a time. So if we have 500 journals, we probably have ~500 to delete every time we reach a specific table.

Considering we also might skip deleting if there's a conflict, we're setting this to 1000 to be sure we'll delete everything we needed to delete once we reach that table.

BedrockJournalDeleter.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+void BedrockJournalDeleter::start(SQLiteNodeState state)
+{
+    if (!enableDeleterThread || !isTrimmableState(state)) {
+        return;
+    }
+
+    lock_guard<decltype(_lifecycleMutex)> lifecycleLock(_lifecycleMutex);
+    if (_thread) {
+        return;
```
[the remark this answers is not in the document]
iwiznia: I don't think this is a concern, right? We will always receive more commits

iwiznia (dismissed):

BedrockJournalDeleter.cpp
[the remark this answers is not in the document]
danieldoglas: I added `wake()` in this if statement to cover that. I don't think it's a huge problem, but with journal we should always aim to delete as muh as possible becaue we have too many tables.

BedrockServer.cpp
```diff
@@ -2461,6 +2501,7 @@ void BedrockServer::handleSocket(Socket&& socket, bool fromControlPort, bool fro
 
 void BedrockServer::notifyStateChangeToPlugins(SQLite& db, SQLiteNodeState newState)
 {
+    _journalDeleter.start(newState);
```
tylerkaraszewski: This doesn't seem like the right place for this? Why start it here?

main.cpp
tylerkaraszewski: I don`t think this is true, right? We wake up once per commit. If we have 500 tables, then each time we hit a particular table, it's been 500 commits since the last time we cleaned it up, but the chances are that almost all of those other commits landed in other tables. So the average number of commits we'll delete is still 1.

tylerkaraszewski (changes requested):

main.cpp
danieldoglas: You are indeed correct. For some reason in my head we were inserting 500 before each delete pass, which is not true. We insert 1, then on the next pass the max we'd have is 2, usually. Changed this to 10.

iwiznia (approved):

tylerkaraszewski (approved):

PR #2775, opened 2026-09-08
Roll back failed synchronization transactions

tylerkaraszewski:
### Details

We changed to EXCLUSIVE here: https://github.com/Expensify/Bedrock/pull/2758 for replication, then we allowed the deleter thread to run during SYNCHRONIZATION, so we need that for synchronization as well.

`_recvSynchronize()` ignores the result of `commit()`, and because it always assuemd it would never conflict, it doesn't handle that case. This change also fixes exception handling here.

A failed commit leaves the transaction, uncommitted hash, and commit lock open, so the node keeps warning while synchronizing and the next replay fails with "already inside transaction". Failed writes can also leave the handle in a transaction.

This checks the commit result and rolls back before reconnecting, for both `SYNCHRONIZE_RESPONSE` and `SUBSCRIPTION_APPROVED`. The handlers also catch SQLite's standard exceptions, including constraint errors. Replay takes the exclusive commit lock before beginning its snapshot, rather than waiting until prepare, so local unreplicated writes can't invalidate it. That holds the lock for the whole replayed transaction, but releases it on failure as well as success.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/679126
$ https://github.com/Expensify/Expensify/issues/679127
$ https://github.com/Expensify/Expensify/issues/679295
$ https://github.com/Expensify/Expensify/issues/679296

### Tests

Added regression coverage for failed commits, failed writes, constraint exceptions, and hash mismatches through both response paths. It checks that another thread can acquire the commit lock and that a valid retry succeeds on the same handle without test-side cleanup. The commit, write, and constraint cases failed before their corresponding fixes. Commit failure is injected with `COMMIT_DISABLED`, not a timing-dependent snapshot conflict.

AI tests run in Vagrant:

- Bedrock full build and Auth rebuild, including `authtest`: passed.
- `./test -only SQLiteNode,AfterCommitCallback,WriteLocalUnreplicated -threads 1 -q`: 13 passed.
- `./test -only SQLiteNode -threads 1 -q -enableHctree`: 6 passed.
- `./clustertest -only MultipleLeaderSync,WriteLocalUnreplicatedCluster -threads 1 -q`: 2 passed, and another 2 passed with `-enableHctree`.
- `uncrustify --check` on both changed files and `git diff --check`: passed.

Full unit, cluster, and Auth test suites were not run.

_________

### DIFF
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -1324,14 +1324,14 @@ void SQLiteNode::_onMESSAGE(SQLitePeer* peer, const SData& message)
                     // Also, extend our timeout so long as we're still alive
                     _stateTimeout = STimeNow() + RECV_TIMEOUT + SRandom::rand64() % STIME_US_PER_S * 5;
                 }
-            } catch (const SException& e) {
+            } catch (const exception& e) {
                 // Transaction failed
                 SWARN("Synchronization failed '" << e.what() << "', reconnecting and re-SEARCHING.");
                 _db.rollback();
                 _reconnectPeer(_syncPeer);
                 _syncPeer = nullptr;
                 _changeState(SQLiteNodeState::SEARCHING);
-                throw e;
+                STHROW(e.what());
             }
         } else if (SIEquals(message.methodLine, "SUBSCRIBE")) {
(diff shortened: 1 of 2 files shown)


danieldoglas (approved):
LGTM

PR #2777, opened 2026-09-08
Accept GUID journal hashes with generation disabled

tylerkaraszewski:
### Details
This is the reader-first part of replacing chained hashes for parallel commits. The journal can now contain either the existing chained SHA1 or `GUID:SHA1`, where the GUID is a hex-encoded 128-bit random number and the digest is `SHA1(GUID + uncompressed SQL)`. The new format checks the current transaction only, not its history. Both live replication and synchronization accept it. `prepare()` takes just the optional GUID; callers extract it from messages and still verify the complete hash before committing. Local generation is implemented but hard-disabled until these readers are deployed everywhere.

The formats can be intermixed. When we switch back to chained hashes, the next hash includes the entire previous `GUID:SHA1`, so an older leader can continue from GUID entries already in its database. It still cannot catch up through GUID entries it has not received; rollback versions that need to synchronize must support reading them. We also reject live transactions from peers other than the selected leader, since the new hashes no longer detect a mismatched previous history.

GUID generation uses two `SRandom::rand64()` draws. To make those safe without a shared lock, `SRandom` now owns its engine through a thread-local `unique_ptr`, allocated and seeded only on that thread's first use and destroyed at thread exit. Normal draws use no shared mutable state. Initialization uses eight `random_device` draws through `seed_seq`; Valgrind builds use a seed counter so new threads do not all repeat the same sequence. The public API is unchanged and the generator remains non-cryptographic.

### Fixed Issues
Part of https://github.com/Expensify/Expensify/issues/337537

### Tests
Added coverage for mixed-format histories, identical SQL with different GUIDs, compressed and empty transactions, corruption and malformed GUID rejection, rollback/retry, reopening a GUID-tailed journal and resuming legacy hashes, and rejecting non-leader transactions. The RNG tests cover bounds, strings, boolean endpoints, concurrent calls to all four APIs, and distinct streams across two waves of new threads.

_________

### DIFF
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -1836,7 +1836,7 @@ void SQLiteNode::_recvSynchronize(SQLitePeer* peer, const SData& message)
             STHROW("failed to write transaction");
         }
         string newHash;
-        if (!_db.prepare(nullptr, &newHash)) {
+        if (!_db.prepare(nullptr, &newHash, chrono::hours(24), nullptr, commit["Hash"])) {
             STHROW("failed to prepare transaction");
         }
         if (newHash != commit["Hash"]) {
@@ -2037,7 +2037,7 @@ bool SQLiteNode::_handlePrepareTransaction(SQLite& db, SQLitePeer* peer, const S
     }
 
     bool success = true;
-    if (!db.prepare()) {
+    if (!db.prepare(nullptr, nullptr, chrono::hours(24), nullptr, message["NewHash"])) {
         SALERT("failed to prepare transaction");
         success = false;
         db.rollback();
--- a/sqlitecluster/SQLiteNode.cpp
+++ b/sqlitecluster/SQLiteNode.cpp
@@ -1389,9 +1389,11 @@ void SQLiteNode::_onMESSAGE(SQLitePeer* peer, const SData& message)
                 return;
             }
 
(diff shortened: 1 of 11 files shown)


sqlitecluster/SQLiteNode.cpp
```diff
@@ -1389,6 +1389,13 @@ void SQLiteNode::_onMESSAGE(SQLitePeer* peer, const SData& message)
                 return;
             }
 
+            // Transactions from a previous leader may still arrive after we switch leaders.
+            // Only our selected leader may extend our transaction history.
+            if (peer != _leadPeer) {
+                PINFO("Ignoring transaction from non-leader");
+                return;
+            }
+
```
danieldoglas, line 1398: What's this for? I don't understand why we'd need to add this check specifically

danieldoglas (approved):
LGTM

sqlitecluster/SQLiteNode.cpp
tylerkaraszewski: So, I'm not sure how absolutely necessary this is, but I think it's a pretty easy check that might add some safety in a weird case. If something happens where a leader connects/disconnects rapidly and so who's leader swaps back and forth, we might get a message from a node we no longer think is our leader. I don't know if there's an actual case where we could trigger this, but it's an easy check to ignore replication messages that don't come from leader.

danieldoglas: Ok, sounds good

PR #2780, opened 2026-09-10
Always generate GUID journal hashes

tylerkaraszewski:
~HOLD for:
https://github.com/Expensify/Bedrock/pull/2777~

Needs to be fully deployed.

### Details
This is the writer follow-up to [the mixed-format readers](https://github.com/Expensify/Bedrock/pull/2777). Local transactions always generate `GUID:SHA1` in `SQLite::prepare()`, and the disabled generation gate in `SQLiteCore` is removed. Replication still supplies the received GUID, or an explicit empty string for a chained hash, so live replication and synchronization continue to recompute and validate both formats.

This must wait until the mixed-format readers are deployed everywhere. Any rollback version that may need to synchronize must also support them.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/337537

### Tests
Added coverage for automatic GUID generation through both `prepare()` and leader commits, distinct GUIDs when retrying identical SQL, empty transactions, persisted hashes, and continuing with GUID hashes after reopening the journal. Existing mixed-history and corruption-rejection tests still run for live replication, synchronization, and subscription.

### AI Tests
- Reproduced the new tests failing before the implementation: both default preparation and post-restart commits produced 40-character hashes instead of 73.
- Rebuilt Bedrock, both test binaries, and the test plugin with Clang 18 in the Vagrant VM. JSON symbol checks passed.
- Rebuilt Auth against this Bedrock worktree successfully.
- `./test -only SQLiteNode,SRandom,LibStuff,AfterCommitCallback,WriteLocalUnreplicated,JournalDeleter -threads 4`: 64 passed, 0 failed, in both WAL and HCTree modes.
- `./clustertest -only ForkCheck,Compression,MultipleLeaderSync,BoundParameters,GracefulFailover -threads 4`: 8 passed, 0 failed, in WAL mode. The first invocation exceeded the 120-second command timeout; the rerun with a longer timeout completed successfully.
- `git diff --check` passed. Full suites and HCTree cluster tests were not run.

_________

### DIFF
--- a/docs/blockchain.md
+++ b/docs/blockchain.md
@@ -9,7 +9,7 @@ Though the blockchain hype has largely come and gone, it's worth pointing out th
 To be clear, Bedrock is not a "public" blockchain, like Bitcoin -- it is not designed to synchronize a series of records between thousands or millions of anonymous peers over the open internet.  Rather, Bedrock uses a "private" blockchain, meaning that a small cluster of servers (3-6) operating in a controlled environment connect to each other on an equal footing to synchronize a historical record of commits, and apply them in the same order.
 
 ## The Journal
-With chained hashes (the default), it works like this:
+New local commits use the GUID transaction hashes described below. Existing journals can also contain legacy chained hashes, which work like this:
 
 * Journal entries have three columns:
     * `id` - Simple monotonic index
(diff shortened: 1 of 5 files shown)


danieldoglas (approved):
LGTM

tylerkaraszewski:
I think I'm going to wait for one more deploy before merging this to make rollback cases just slightly clearer.

PR #2784, opened 2026-09-14
Cache SQResult column lookups for stable C-string keys

tylerkaraszewski:
Before this change, reading a database result by column name, such as `row["amount"]`, rebuilt that name and searched through the column headers every time, even when the same column was read across thousands of rows. This change remembers the matching column number after the first lookup, so later reads can jump directly to the value. Across five representative test cases containing about 7,500 to 56,000 transactions, converting the transactions to JSON was **2–8% faster** in every case, saving roughly 3–37 ms each time a complete test dataset was converted. When timed together with the other in-memory transaction processing in the benchmark, four datasets finished **3–5% faster** and the noisiest dataset showed no reliable difference. The output remained identical, all 89 relevant Bedrock tests passed, and Auth still builds against the change.

| Transactions processed | Time to build JSON before | Time to build JSON after | Time saved building JSON | JSON building speedup | Total processing speedup |
|---:|---:|---:|---:|---:|---:|
| 7,477 | 57.37 ms | 54.17 ms | 3.20 ms | 5.9% faster | 3.6% faster |
| 26,514 | 215.66 ms | 198.38 ms | 17.28 ms | 7.7% faster | 5.1% faster |
| 56,404 | 382.09 ms | 344.92 ms | 37.17 ms | 2.0% faster | No reliable difference |
| 34,915 | 229.27 ms | 216.26 ms | 13.01 ms | 6.0% faster | 3.2% faster |
| 27,278 | 176.29 ms | 167.01 ms | 9.28 ms | 5.6% faster | 3.0% faster |

Each row represents converting one complete transaction dataset from an in-memory database result into JSON. “Total processing” includes the other in-memory transaction-processing steps measured by this benchmark, but does not include reading from the database, network time, or the rest of an OpenApp request. The before/after times are the middle result from repeated runs. The speedup percentages compare matching runs on the same CPU to reduce normal timing noise, which is why they do not always equal the simple difference between the two displayed times; the 56,404-transaction case varied the most.

$ https://github.com/Expensify/Expensify/issues/681551

_________


libstuff/SQResult.cpp
```diff
@@ -104,20 +106,85 @@ string SQResultRow::operator[](const string& key)
...
+    headerIndexesByAddress.clear();
+    headers = move(newHeaders);
+}
+
+optional<size_t> SQResult::findHeaderIndex(const char* key) const
+{
+    lock_guard<mutex> lock(headerIndexMutex);
+    const auto cached = headerIndexesByAddress.find(key);
+    if (cached != headerIndexesByAddress.end()) {
+        return cached->second;
```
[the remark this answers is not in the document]
tylerkaraszewski: This needs an auth change before we can fix it in a follow-up

libstuff/SQResult.cpp
```diff
@@ -104,20 +106,85 @@ string SQResultRow::operator[](const string& key)
...
+SQResult::SQResult(vector<SQResultRow>&& rows, vector<string>&& headers)
+    : headers(move(headers)), rows(move(rows))
+{
+    rebindRows();
+}
+
+void SQResult::rebindRows()
+{
+    for (auto& row : rows) {
+        row.result = this;
```
danieldoglas, line 154: just curious, why do we bind the row to the result object?

libstuff/SQResult.cpp
```diff
@@ -104,20 +106,85 @@ string SQResultRow::operator[](const string& key)
...
+}
+
+const vector<string>& SQResult::getHeaders() const
+{
+    return headers;
+}
+
+void SQResult::setHeaders(vector<string> newHeaders)
+{
+    lock_guard<mutex> lock(headerIndexMutex);
```
danieldoglas, line 165: I understand why we need this mutex, but do we have any cases where we have SQResults being used by multiple threads?

danieldoglas (approved):
LGTM, just a few questions

libstuff/SQResult.cpp
tylerkaraszewski: I expect we don't, and we could probably omit this here for that reason, but this is safer.

libstuff/SQResult.cpp
tylerkaraszewski: The actual column names are a property of the `SQResult` rather than the row, so that we need only one copy of them. Each row, when accessed by name, looks up that name on its parent.

PR #2786, opened 2026-09-15
Reduce JSON path lookup and object construction overhead

tylerkaraszewski:
### Details

This PR changes the following `JSON::Value` functions:

- `Value::Value(initializer_list<KeyValue>)` now uses end-hinted insertion. Objects whose keys are supplied in ascending order take less work to build, while unsorted and duplicate keys keep their existing behavior.
- `Value::extractTo(iterator, target, key, hint)` is a new overload that transfers and renames an existing map node. Callers can move a member into another object without allocating and rebuilding that member, and can use a target insertion hint when they know the destination order.
- `Value::getValueAtPath(const list<string>&, const Value&) const&` now searches with `find()` instead of using `operator[]` and catching a missing-key exception. Missing optional fields return the default directly rather than constructing an exception and capturing a backtrace.
- `Value::findValueAtPath(initializer_list<string>)` is a new shared lookup helper. Literal paths such as `{"transaction", "amount"}` can be searched without first allocating a linked list.
- `Value::getValueAtPath(initializer_list<string>, const Value&) const&` exposes that cheaper lookup for borrowed values and defaults.
- `Value::getValueAtPath(initializer_list<string>, const Value&) &&` does the same lookup on temporary values and moves the result out rather than copying it.
- `Value::getValueAtPath(initializer_list<string>) const&` and `Value::getValueAtPath(initializer_list<string>) &&` add the corresponding literal-path overloads when no explicit default is needed.

Hold for https://github.com/Expensify/Bedrock/pull/2784. This PR is stacked on that branch. The [Auth consumer change](https://github.com/Expensify/Auth/pull/24507) uses these APIs to borrow parsed comments, construct transaction fields in key order and transfer existing transaction map nodes into the Onyx collection.

The focused benchmarks used optimized builds and 20 alternating runs against the baseline:

- Repeated optional-path lookup over 56,404 records fell from about 952 ms to 100 ms. That phase used about **90% less time**, or finished in roughly one tenth of the baseline time.
- Across five result sets containing 7,477 to 56,404 transactions, the complete helper phase used **86.1% to 90.1% less time**.
- Building and destroying 10,000 objects with 32 ascending keys fell from 23.73 ms to 16.78 ms, about **29.5% less time**.
- The same object test still improved when keys were not sorted: **6.6% less time** for reversed keys and **4.6% less time** for a fixed mixed order.

The path result measures the complete helper phase, not an isolated call to `getValueAtPath()`, and also includes comment borrowing from the Auth change. Node transfer was measured where Auth uses it rather than as a synthetic Bedrock-only operation. The combined Bedrock and Auth changes are projected to reduce the measured Auth command time by **1.1% to 16.2%**, or the whole endpoint time by **0.9% to 11.4%**, depending on how much of each logged phase the benchmark represents. Those whole-command numbers are estimates rather than direct end-to-end timings.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/681551

### Tests

The focused Bedrock test run completed with **85 passed and 0 failed**:

```text
./test -only JSONValue,JSON,JSONParser,JSONUtils,SQResult
[ TEST RESULTS ] Passed: 85, Failed: 0
```

The new tests verify literal paths and defaults, sorted and unsorted object construction, duplicate keys, moving and renaming nodes, overwriting target keys and rejecting invalid aliases. All five benchmark outputs remained byte-for-byte identical to the baseline and successfully parsed back as JSON, so the measured speedup did not change their output.

Auth, its test plugin and `authtest` also compiled and linked in an optimized build against this Bedrock branch. The Auth Transaction/OpenApp tests could not start because untouched HubSpot static-initialization code crashes before the test runner reaches `main`; that startup problem was not changed or fixed here.

_________


tylerkaraszewski:
@MelvinBot create a followup issue to address the TODO in this change with an Auth PR to use getHeaders and a finla follow-up after that to change `headers` visibility. Assign to me.

danieldoglas (approved):

PR #2787, opened 2026-09-15
Close port if WAL file grows too large

iwiznia:
### Details
Context https://expensify.slack.com/archives/C05CBC62HGW/p1789472071027169?thread_ts=1788881696.435499&cid=C05CBC62HGW
When WAL file grows (usually due to long queries running back to back), it generates slows on both followers and leader. This allows us to set a max amount of wal frames to tolerate before the server closes the command port to stop processing commands and allow it to checkpoint the wal frames while it is not serving customer requests, thus reducing customer impact.
I set the default to 0 (never close the port due to wal frames being large) so we can deploy this safely, we can change the number via `SetMaxOutstandingWALFrames` command after it is deployed to test it, after successful tests I'll change the default. We are expecting 200k to be the number to choose.

### Tests
Tested the `SetMaxOutstandingWALFrames` command, but I have no idea how to actually test the port closing in dev.

_________


sqlitecluster/SQLiteNode.h
```diff
@@ -239,6 +239,8 @@ class SQLiteNode : public STCPManager {
     // commitCount that we do, this will return null.
     void _updateSyncPeer();
 
+    void _updateCommandPortForWALSize(uint64_t outstandingFramesToCheckpoint);
```
flodnv, line 245: Add a short doc please

BedrockServer.cpp
```diff
@@ -1937,6 +1948,19 @@ void BedrockServer::_control(unique_ptr<BedrockCommand>& command)
...
+        if (!command->request.isSet("maxOutstandingWALFrames")) {
+            response.methodLine = "400 Missing maxOutstandingWALFrames";
+        } else if (newMax < 0) {
+            response.methodLine = "400 maxOutstandingWALFrames can't be negative";
+        } else {
+            const uint64_t oldMax = _maxOutstandingWALFrames.exchange(static_cast<uint64_t>(newMax));
+            SINFO("Setting _maxOutstandingWALFrames to " << newMax << " from " << oldMax);
+            response["oldMaxOutstandingWALFrames"] = to_string(oldMax);
+            response["newMaxOutstandingWALFrames"] = to_string(newMax);
+            _notifyDoneSync.push(true);
```
flodnv: Why is this here?

iwiznia: heh AI added it and I asked the same 😄 
Because we updated the threshold and this makes the change apply immediately by waking the sync thread, instead of having to wait till next time it wakes up... maybe it's not needed since I think that will happen at most 1s later. Should I remove it?

sqlitecluster/SQLiteNode.cpp
```diff
@@ -479,6 +481,8 @@ bool SQLiteNode::update()
         SWARN("Killed, skipping update");
         return false;
     }
+
+    _updateCommandPortForWALSize(getOutstandingFramesToCheckpoint());
```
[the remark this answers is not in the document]
iwiznia: While this is technically true, I don't think we care much because we always have writes, so waiting till the next write is such a small time that it basically does not matter.

BedrockServer.cpp
tylerkaraszewski: I would remove it for consistency's sake, if the sync thread never wakes up who cares what the WAL limit is.

tylerkaraszewski (changes requested):

BedrockServer.cpp
iwiznia: Done

tylerkaraszewski (dismissed):

tylerkaraszewski (approved):

flodnv (approved):
🙏

iwiznia:
Tests for this are here https://expensify.slack.com/archives/C094TGUTZ/p1789649692515419 it worked correctly

PR #2798, opened 2026-09-18
Allow blank journal entries and remove chained hashes

tylerkaraszewski:
### Details
HC-Tree can allocate a commit ID before a transaction fails, and that ID still needs to be replicated as a blank journal entry. This prepares followers to accept those entries during replication and synchronization, committing the journal row without executing SQL. It also removes chained-hash replay now that local commits use GUID hashes. Local commits continue to generate GUID hashes; enabling blank writers comes later, once every node can read them.

`CommitCount` includes blank entries, while `HashCommitID` and `Hash` identify the latest nonblank entry. Fork checks compare the latest nonblank entry in the shared prefix, so nodes can agree while having different numbers of trailing blanks. Synchronization still transfers every row, and pruning preserves the latest nonblank entry and its suffix so we can recover that identity after a restart. Missing `HashCommitID` defaults to `CommitCount` during rollout. Nodes must have a GUID-only synchronization tail before upgrading because chained entries can no longer be replayed.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/337537

#### AI Tests
Bedrock and Auth compiled successfully in Vagrant.

From `test/`:
```text
./test -threads 8
[ TEST RESULTS ] Passed: 362, Failed: 0

./test -threads 8 -enableHctree
[ TEST RESULTS ] Passed: 362, Failed: 0
```

From `test/clustertest/`:
```text
./clustertest -threads 2 -except ClusterUpgrade
[ TEST RESULTS ] Passed: 83, Failed: 0

./clustertest -threads 2 -except ClusterUpgrade -enableHctree
[ TEST RESULTS ] Passed: 83, Failed: 0
```

The focused `SQLiteNode,JournalDeleter` suite also passed in both modes after the final changes: 21 passed, 0 failed each. It covers blank replication/synchronization, fork agreement, restart, retention, failure cleanup, and rejection of chained hashes.

`ClusterUpgrade` was attempted but could not initialize because Git inside the VM cannot resolve the worktree's host-side Git path. It is excluded from the successful cluster runs above.

_________


danieldoglas (approved):
LGTM. Quite a lot of changes, maybe do a CP to deploy this and confirm it won't cause issues?

PR #2803, opened 2026-09-21
Return job snapshots as JSON strings

inimaga:
### Details

PHP currently has to decode and re-encode job data to send the original snapshot back to Bedrock, which can change JSON types. This adds an `expectedData` string to `GetJob` and `GetJobs` so workers can return the original data unchanged when finishing, retrying, or failing a job. The existing `data` object is preserved.

### Fixed Issues

https://github.com/Expensify/Expensify/issues/670760

### Tests

Covered by added automated test

### Query Timings/Plans

N/A


plugins/Jobs.cpp
```diff
@@ -1125,7 +1126,10 @@ void BedrockJobsCommand::process(SQLite& db)
                 // We will fail this job, don't return it.
                 continue;
             }
-            jobList.push_back(SComposeJSONObject(job));
+            string jobResponse = SComposeJSONObject(job);
+            // Force the snapshot to a JSON string without re-encoding the existing data object.
+            jobResponse.insert(1, "\"expectedData\":" + SToJSON(job["data"], true) + ",");
```
iwiznia, line 1131: Instead of manipulating the string like this, can't we do `job["expectedData"] = job["data"];`?

test/tests/jobs/GetJobTest.cpp
```diff
@@ -96,7 +97,7 @@ struct GetJobTest : tpunit::TestFixture
         response = tester->executeWaitVerifyContentTable(command);
         uint64_t end = STimeNow();
 
-        ASSERT_EQUAL(response.size(), 9);
+        ASSERT_EQUAL(response.size(), 10);
```
iwiznia, line 100: Why is this changing? Weren't we already returning the data in some way?

plugins/Jobs.cpp
inimaga: I considered this, but [SComposeJSONObject()](https://github.com/Expensify/Bedrock/blob/46112647921129174557c35451cde2344fdeb577/libstuff/libstuff.cpp#L1557-L1570) would return `expectedData` as an object. It has to be a string so PHP passes it back unchanged. Using `forceString` for the whole response would also turn data and the child-job arrays into strings. So I went with this approach to make only `expectedData` a string.

test/tests/jobs/GetJobTest.cpp
inimaga: Yes, we already return data as an object, and that stays the same. This is for `expectedData` which is added as a separate string. That adds one field to the response, so the count increases from 9 to 10.

iwiznia: > we already return data as an object

Why do we need it both as an object and as a string?

inimaga: Existing workers use data to do their work, so I'd prefer to keep it as an object to avoid breaking them. `expectedData` is the original copy they send back so Bedrock can check whether the data changed while they were running.

plugins/Jobs.cpp
pecanoro: What I don't fully understand is why we can't just pass it here and then pass the object back directly again in PHP?

inimaga: @pecanoro The issue is that PHP doesn’t pass the object straight back. The client first uses `json_decode($body, true)` [here](https://github.com/Expensify/Bedrock-PHP/blob/bfabf18fcbe3ab18678718574f8df2777750a25f/src/Client.php#L827), and this turns it into a PHP array. 
So if we want to pass and just send it back, something like `{"options":{}}` can become `{"options":[]}`. Bedrock would see that difference and run the job again, even though nothing changed.

inimaga:
@pecanoro bump for review as @iwiznia is out of office

plugins/Jobs.cpp
pecanoro: Can we just use `job["data"].serialize()`?

